PR #28, opened 2024-05-23
Add initial github actions workflow for testing java converter

springmeyer:
We don't need sophisticated CI tooling yet, but something simple to help confirm that local fixes run the same on linux would be extremely useful. This gets us started, testing only the java converter and running its tests.

/c @nyurik @mactrem


nyurik:
I can't enable actions on this yet (I don't have admin rights, just merge rights I think - will need that in order to transition the repo to ML). Regardless, seems ok to merge.

PR #33, opened 2024-05-24
Add windows to CI

springmeyer:
This enables CI for windows. This might be useful to merge if we see any differences in test behavior on windows vs unix systems (to be determined).


nyurik (approved):
makes sense to merge it now, thx

PR #34, opened 2024-05-24
Add FSST decoder (typescript)

etnav:
In this PR we're adding a typescript project which provides a decoding function for FSST


encodings/fsst/js/src/fsstDecoder.ts
```diff
@@ -0,0 +1,23 @@
+export class FsstDecoder {
```
mactrem: Let's simplify and get rid of that class and just export the function like `decodeFsst`

mactrem (commented):
LGTM! just one minor change and we can merge it

mactrem:
Thanks! In a separate step we should then get rid of the overhead due the slice operation

PR #47, opened 2024-05-28
Get JS parser building

ebrelsford:
This PR
 * Explicitly sets the node version to `20.10` (to match [maplibre-gl-js](https://github.com/maplibre/maplibre-gl-js/blob/main/.nvmrc), but open to a different version if desired)
 * Re-builds the `package-lock.json` file since the URLs in the old one were not resolving

I know there will be bigger changes to make to the parser down the road but figured I'd start here.


nyurik (approved):
lgtm

springmeyer (approved):

PR #48, opened 2024-05-28
[WIP] Another batch of java encoder fixes

springmeyer:
Starting another PR to work on the Java encoder code.

So far this PR:

 - Gets the build working cross-platform: now all tests are able to run on macOS, ubuntu, and windows 🎉 
   - Note: some tests are still out of sync with the converter code after #25 (and therefore fail). Fixing them will happen in a separate PR
 - Consolidates the fsst JNI encoding code into java/converter
 - Removes windows specific `FsstWrapper.dll` and `fsst.dll` files that were checked into the repo
 - Github actions workflow improvements:
    - Reduces code duplication between windows and ubuntu
    - Adds macOS
    - Consolidates to a single java distribution across all platforms
   

TODO:
  - [x] Get windows builds working on CI
  - [ ] Use BFG to purge dll files? https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository#using-the-bfg


converter/java/build.gradle
```diff
@@ -42,4 +40,16 @@ java {
...
+    doFirst {
+        if (System.properties['os.name'].toLowerCase().contains('windows')) {
+            executable "./resources/compile.bat"
+        } else {
+            executable "./resources/compile.sh"
+        }
+    }
+}
+
+compileJava.dependsOn compileWrapper
```
nyurik: please configure your IDE to auto-add newlines - these get in a way when reviewing. Thx :)

.github/workflows/test.yml
```diff
@@ -87,4 +52,22 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
       - uses: gradle/actions/setup-gradle@v3
-      - run: .\gradlew.bat test
+      - name: Set gradle command
+        shell: bash
+        id: vars
+        run: |
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+              echo gradle_cmd=".\gradlew.bat" >> $GITHUB_OUTPUT
```
nyurik: I could be wrong, but I think you can simply set this at the top level, and `./gradlew` will work in all cases. This way you can simply change all of this to:

```
defaults:
  run:
    shell: bash

//...

  - name: Install JDK (Linux)
    if: runner.os == 'Linux'
    run: sudo apt-get install openjdk-17-jdk
  - name: Install JDK (macOS)
    if: runner.os == 'macOS'
    run: brew install microsoft-openjdk@17
  
  - name: ... tests
    run: ./gradlew test --tests "com.mlt.converter.encodings.*"

nyurik (commented):
minor nits

converter/java/build.gradle
springmeyer: 👍🏼 done

.github/workflows/test.yml
springmeyer: Thanks! Stepping out for the day now, but will take a look at integrating this suggestion tomorrow.

springmeyer: Excellent, it works. Thanks! https://github.com/maplibre/maplibre-tile-spec/pull/48/commits/92b16d781861c61abcc874bc109a773d75e2d6bc

springmeyer:
This is ready for review. Overall my motivation here was to:

 - Get the fsst encoding working in the converter code/tests: ✅  (this was pretty easy)
 - See if I could get windows working on CI: ✅ (this was not as easy since I don't use windows or have access to a machine, but I figured it out with some exploratory commits while making dinner and breakfast)

Now, on all 3 platforms we see roughly the same test behavior for the converter code:

> 69 tests completed, 24 failed

In followup PRs we can deal with the failing tests, and ensure they all start passing on all platforms in the same way.

The windows approach I took was to use [mingw](https://code.visualstudio.com/docs/cpp/config-mingw) which I realize may not be the preferred way to develop on windows for others. But it offered a fairly seamless way to get the same code working on all platforms, and test it on CI. I'm hoping this works for now to land in `main` and then I'm open to ideas for better ways to support windows in the future (more matched to how the team wants to use/test on windows locally).

nyurik (approved):

PR #49, opened 2024-05-29
Run JS parser unit tests

ebrelsford:
This PR sets up automated unit tests for the JS parser. It includes:
 * Splitting unit tests and benchmarking tests up
 * Adding to the existing test workflow
 * Skipping a tile that was causing timeouts because the test was too slow with it

I know the parser is going to change significantly, my hope is to get these tests running as a foundation to build on as we get the parser in sync with the encoder.


parser/js/test/unit/jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+    displayName: 'Unit',
+    transform: {
+        "^.+\\.ts": "ts-jest",
+    },
+    testRegex: "(.*|(\\.|/)(test|spec))\\.(js|ts)$",
+    moduleFileExtensions: ["ts", "js", "json", "node"],
+};
```
springmeyer: @ebrelsford Surprisingly, I noticed that this introduces another failure when running:

```
npm test
```

```
 FAIL   Unit  test/unit/jest.config.js
  ● Test suite failed to run

    Your test suite must contain at least one test.

      at onResult (../../node_modules/@jest/core/build/TestScheduler.js:133:18)
      at ../../node_modules/@jest/core/build/TestScheduler.js:254:19
      at ../../node_modules/emittery/index.js:363:13
          at Array.map (<anonymous>)
      at Emittery.emit (../../node_modules/emittery/index.js:361:23)
```

parser/js/test/benchmark/jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+    displayName: 'Benchmarking',
+    transform: {
+        "^.+\\.ts": "ts-jest",
+    },
+    testRegex: "(/test/benchmarking/.*|(\\.|/)(test|spec))\\.(js|ts)$",
+    moduleFileExtensions: ["ts", "js", "json", "node"],
+};
```
springmeyer: If I'm understanding right, I think a file like this is not needed as the benchmark scripts are not jest tests.

ebrelsford: Yes, good catch!

parser/js/test/unit/jest.config.js
ebrelsford: Ah yeah I overcomplicated the jest config here, simplifying it should fix--thanks for pointing that out.

springmeyer: [6732f66](https://github.com/maplibre/maplibre-tile-spec/pull/49/commits/6732f660b627b731716ce3d7f316f2951d8079b9) looks good 👍🏼

ebrelsford:
@springmeyer I believe I addressed your comments above. I couldn't get `jest` to hang on the other tiles you saw that happening with.

springmeyer (approved):
Change looks good. Seems like a value add to get this landed. However now that CI is green (#56) I'm hesitant to merge this as I'd love to keep CI green.

So, perhaps we keep plugging away at this until there is a way visible to get the test actually skipped or passing?

ebrelsford:
That's fair @springmeyer, I added `.skip` to the tests that were failing:

```bash
 PASS  test/unit/evaluation/decodingUtils.spec.ts
 PASS  test/unit/decoder/decodingUtils.spec.ts
 PASS  test/unit/decoder/covtDecoder.spec.ts (22.304 s)

Test Suites: 2 skipped, 3 passed, 3 of 5 total
Tests:       6 skipped, 17 passed, 23 total
Snapshots:   0 total
Time:        22.535 s
Ran all test suites.
```

I know the parser needs to be updated a bit to match the encoder, so I tried not to get too bogged down in fixing the tile comparison tests. I did notice that part of the issue was that MVT geometries had member functions that were making the comparison difficult, though, and added a step to account for that in the future.

springmeyer (approved):
👍🏼 Great work. Looks good to merge.

Followup thoughts while I'm here for future PRs of js parser:

 - Renaming all COVT references in code to MLT
 - Regenerating all .covt tiles used in the tests (as those will be very out of date with current encoder)

Issue #38, opened 2024-05-24
Ensure unit tests run cleanly for encodings/fastpfor/js/


springmeyer:
Currently the unit tests appear to pass but still result in a failure due to a minor issue with nyc & Ava usage:

>   ✘ No tests found in src/tests/benchmarks/fastpfor.ts, make sure to import "ava" at the top of your test file

springmeyer:
Looks like there is another problem coming up on CI:

```
> fastpfor-ts@1.0.0 build:module
> tsc -p tsconfig.module.json

src/index.ts:12:28 - error TS2307: Cannot find module './BitPacking' or its corresponding type declarations.

12 import { fastunpack } from './BitPacking';
                              ~~~~~~~~~~~~~~


Found 1 error in src/index.ts:12

src/index.ts:12:28 - error TS[23](https://github.com/mactrem/cov-tiles/actions/runs/9228505345/job/25392651682#step:5:24)07: Cannot find module './BitPacking' or its corresponding type declarations.

12 import { fastunpack } from './BitPacking';
                              ~~~~~~~~~~~~~~
```

https://github.com/mactrem/cov-tiles/actions/runs/9228505345/job/25392651682#step:5:14

springmeyer:
Fixed by #50

PR #50, opened 2024-05-29
Fix fastpfor JS tests

ebrelsford:
Building and testing the fastpfor library are currently failing, this PR attempts to fix this by:
 * Fixing the `bitpacking` reference (capitalization)
 * Only running unit tests (not benchmarks), we'll probably want to do that separately if we want those to be automated

### DIFF
--- a/encodings/fastpfor/js/package.json
+++ b/encodings/fastpfor/js/package.json
@@ -19,7 +19,7 @@
     "test:lint": "eslint src --ext .ts",
     "test:prettier": "prettier \"src/**/*.ts\" --list-different",
     "test:spelling": "cspell \"{README.md,.github/*.md,src/**/*.ts}\"",
-    "test:unit": "nyc --silent ava",
+    "test:unit": "nyc --silent ava src/tests/unit/**/*.ts",
     "check-cli": "run-s test diff-integration-tests check-integration-tests",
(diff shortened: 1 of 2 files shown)


encodings/fastpfor/js/package.json
```diff
@@ -19,7 +19,7 @@
     "test:lint": "eslint src --ext .ts",
     "test:prettier": "prettier \"src/**/*.ts\" --list-different",
     "test:spelling": "cspell \"{README.md,.github/*.md,src/**/*.ts}\"",
-    "test:unit": "nyc --silent ava",
+    "test:unit": "nyc --silent ava src/tests/unit/**/*.ts",
```
nyurik, line 22: if we are not running everything here, we should probably add `test:bench` and maybe some other `test:...` to run the other types. Otherwise they will be in the code, but we won't have an easy way to run them.

springmeyer (approved):
Looks good, and fixes #38. Handling benchmark running on CI seems reasonable to tackle separately from this PR.

encodings/fastpfor/js/package.json
ebrelsford: Agreed that this should be added, but previous to this PR these benchmarks were not running and were breaking the `test:unit` target since they don't use `ava`. So I was considering this as a first step to get the unit tests running.

PR #55, opened 2024-05-30
Add CLI tool to convert an MVT into an MLT

springmeyer:
This PR:

 - Adds a java command line program to convert an MVT into an MLT (mvt2mlt)
 - Adds a CI job that tests the program can run and outputs a file of the expected length

This is intended as an exploratory tool to ensure that all MVT can be decoded and re-encoded without unexpected errors. If we discover tiles that cannot be converted, then those tiles would be candidates to build into the formal unit tests.

Future PRs to this code could add command line options to control encoding settings. For now all optimizations are disabled by default in order to avoid issues like #59


TimSylvester:
Resolves #144

PR #54, opened 2024-05-30
Get CI green

springmeyer:
This limits to running just the java converter tests which are expected to pass currently. It excludes:

 - DecodingUtilsTest
 - VectorizedDecodingUtilsTest
 - StringDecoderTest
 - IntegerDecoderTest
 - MltConverterTest

The goal here is to help with upcoming development by getting CI to the place where we expect it to pass for future PRs.

PR #56, opened 2024-05-30
Temporarily introduce @Disabled decorator in Java tests

springmeyer:
Some tests are not intended to be passing yet after recent refactoring. While those tests get fixed it would be ideal to keep all existing passing tests working, and getting CI 🍏 is one way to accomplish this. As an alternative to #54, this PR starts working on selectively marking failing tests as skipped. The idea would be to:

 - Skip all currently failing tests until CI is 🍏 
 - Then in followup PRs incrementally fix and re-enable tests once passing

As an example, this command:

> ./gradlew test --tests "com.mlt.decoder.IntegerDecoderTest"

Now gives:

Screenshot 2024-05-30 at 12 46 07 PM

Whereas previously it gave:

Screenshot 2024-05-30 at 12 51 05 PM

### DIFF
--- a/converter/java/build.gradle
+++ b/converter/java/build.gradle
@@ -33,6 +33,8 @@ test {
     testLogging {
         outputs.upToDateWhen {false}
         showStandardStreams = true
+        events "passed", "skipped", "failed"
+        exceptionFormat "full"
     }
 }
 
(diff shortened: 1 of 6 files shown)


ebrelsford (approved):
Confirmed locally, this works for me

springmeyer:
All tests are now passing or skipped with this latest PR 🎉 Overall it took skipping 23 tests:

```
 grep -roh '@Disabled' src/test | wc -w
      23
```

PR #60, opened 2024-05-30
Minor java testing improvements

springmeyer:
This PR fixes a few ergonomic issues with the java code.

It fixes this distracting warning:

```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
```

It also ensures that test failures will report why/the exception.

And finally, there are numerous `-Xlint:unchecked` warnings being reported but not in a way that you can see what they are. This makes them visible so that we can more easily notice when we've introduced them in code we are changing.

### DIFF
--- a/converter/java/build.gradle
+++ b/converter/java/build.gradle
@@ -33,6 +33,7 @@ test {
     testLogging {
         outputs.upToDateWhen {false}
         showStandardStreams = true
+        exceptionFormat "full"
     }
 }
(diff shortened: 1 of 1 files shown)


nyurik (approved):
seems ok, but i'm no java expert

PR #62, opened 2024-05-31
Use cmake to build FSST JNI module

springmeyer:
Moves to using `cmake` over `make` for easier cross-platform support for building libfsst and the FSST JNI bindings.

@mactrem does this work on your machine?


converter/java/resources/compile-windows.bat
```diff
@@ -1,7 +1,16 @@
...
-echo %cd%
-bash resources/compile
+
+IF EXIST build\Release\FsstWrapper.so (
+    echo "FsstWrapper.so exists, skipping compilation"
+    echo "  Remove ./build/Release/FsstWrapper.so to recompile"
+    echo "  Remove ./build to reconfigure & compiled"
+) ELSE (
+    echo "FsstWrapper.so does not exist, building now"
+    mkdir -p build
```
mactrem: the `-p` option does not exist on windows. So it creates one `-p` and one `build` directory. On windows the mkdir command executed in the Command Prompt will complete without returning an error if the directory already exists. It only reports that the subdirectory already exists. So we can remove the `-p` parameter. It only reports an error when the command it is executed in the PowerShell instead of the Command Prompt, which is currently probably not an issue?

mactrem (commented):
@springmeyer Great work! Now the build works on my windows machine too. Thanks for that! Just one minor improvement and then we can merge it

converter/java/resources/compile-windows.bat
springmeyer: Great, thanks for catching this. Fixed in a5cff1a.

> It only reports an error when the command it is executed in the PowerShell instead of the Command Prompt, which is currently probably not an issue?

My windows knowledge is very limited so I have no idea! But if you works for you and works on CI then I'm delighted.

PR #63, opened 2024-06-01
Add varint decoding support for vertexBuffer

springmeyer:
I think it is important that we allow for tiles to be encoded and decoded without using FAST_PFOR and FSST (for now).

One blocker with the Java code: tiles can be encoded without FAST_PFOR, but then they cannot be decoded. This PR attempts to fix the decoding support.

/cc @nyurik @mactrem @ebrelsford


converter/java/src/main/java/com/mlt/decoder/GeometryDecoder.java
```diff
@@ -50,12 +50,14 @@ public static GeometryColumn decodeGeometryColumn(byte[] tile, int numStreams, I
                     break;
                 case DATA:
                     if(DictionaryType.VERTEX.equals(geometryStreamMetadata.logicalStreamType().dictionaryType())){
-                        //TODO: add Varint decoding
-                        if(geometryStreamMetadata.physicalLevelTechnique() != PhysicalLevelTechnique.FAST_PFOR){
-                            throw new IllegalArgumentException("Currently only FastPfor encoding supported for the VertexBuffer.");
-                        }
-                        vertexBuffer = DecodingUtils.decodeFastPfor128DeltaCoordinates(tile, geometryStreamMetadata.numValues(),
+                        if(geometryStreamMetadata.physicalLevelTechnique() == PhysicalLevelTechnique.VARINT){
+                            vertexBuffer = DecodingUtils.decodeVarint(tile, offset, geometryStreamMetadata.numValues());
```
mactrem, line 54: In that case the coordinates are also ZigZag and delta encoded in addition to the Varint encoding. I have already implemented it for the vectorized use case see https://github.com/maplibre/maplibre-tile-spec/blob/cfc66fc88595341c62c2d3b3c708029dacaf27b0/converter/java/src/main/java/com/mlt/decoder/vectorized/VectorizedGeometryDecoder.java#L55 So there are currently two ways for decoding a tile. One without the usage of SIMD (where some cases are not fully implemented) and then the new one which will use vectorization. In the future we can get rid of the "old" decoding path.

springmeyer: Thanks for this context! Makes sense there is also zigzag + delta encoding applied. My next step here was to ensure we have tests for the non-vectorized & non-fast-pfor path. And then to confirm they fail before this change and pass after. I figure I would have run into the missing zigzag + delta at that time.

> So there are currently two ways for decoding a tile. One without the usage of SIMD (where some cases are not fully implemented) and then the new one which will use vectorization. In the future we can get rid of the "old" decoding path.

Ah, got it. +1 to cleaning things up so there is only one way. But at the same time: are there any challenges to supporting the vectorization pathway with the Javascript decoder? Does the vectorization pathway simply structure the data in a way to make using SIMD possible? Or does it somehow require it?

mactrem: There is currently no support for SIMD instructions in JS only in WebAssembly. So the goal in the second step is to write a Rust decoder which is using vectorized instructions and compile it to WebAssembly. Sure, for the first step a Js POC decoder is enough. But it should not make that much difference only the implementations of some encodings has to be exchanged for example scalar FastPFOR has to be used instead of a vectorized FastPfor version or a scalar Rle version instead of a vectorized one

springmeyer:
@ebrelsford another idea: considering having all test assertions be on JSON objects. We talked before about writing the tests to assert on each part of the decoded tile (which would be valuable for incrementally ensuring parts start working and for showing specifically what is failing to work). Still like this idea. But also, here is a quick alternative idea: to have the decoder create an entire GeoJSON in memory and then use a library like https://www.npmjs.com/package/json-diff to diff against an entire expected GeoJSON. This could accomplish the former goals with an additional benefit: as the tile fixtures grow increasingly complicated in terms of geometries we could then easily visualize by dumping the GeoJSON output to disk and tossing in a mapping program of choice. And of course the two approaches could be blended together.

springmeyer:
This issue was solved by #129

PR #68, opened 2024-06-03
Fix Java converter CI runs

springmeyer:
#64 cased a 4 byte increase in the output MLT used in the CLI test. This fixes the CI failures by updating the test expectation. @mactrem I presume this slight increase in bytes is expected?


mactrem:
@springmeyer Just rearranged some data, to fix the test, but added no additional data. Interesting, that this has resulted in a negligible increase in file size

PR #71, opened 2024-06-03
Enhance the java CLI tool

springmeyer:
I'd like to use the java CLI tool to:

 - Find tiles we don't yet support
 - Debug why conversion is not correct if something fails
 - Write both to a directory and to an individual file
 - Test decoding the MLT wire format to ensure that works without errors
 - Compare the decoded MLT with the original MVT it is based on
 - Be able to toggle advanced encoding options on and off

This PR accomplishes all these features with new options.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,5 +76,5 @@ jobs:
       - name: Test running CLI
         run: |
           ./gradlew cli
-          java -jar ./build/libs/mvt2mlt-1.0-SNAPSHOT.jar -mvt ../../test/fixtures/omt/mvt/10_530_682.mvt -output output
+          java -jar ./build/libs/mvt2mlt-1.0-SNAPSHOT.jar -mvt ../../test/fixtures/omt/mvt/10_530_682.mvt -dir output
           python -c 'import os; expected=66984; ts=os.path.getsize("output/10_530_682.mlt"); assert ts == expected, f"tile size changed from expected ({expected}), got: {ts}"'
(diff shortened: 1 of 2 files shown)


ebrelsford (commented):
Is it expected that anything that uses the decoder will fail for now? Specifically `-printmlt` and `-compare` fail for me with the omt tiles:

```bash
> java -jar ./build/libs/mvt2mlt-1.0-SNAPSHOT.jar -mvt ../../test/fixtures/omt/mvt/10_530_682.mvt -mlt output/adv.mvt -advanced -printmlt

Time elapsed for encoding: 1021 milliseconds
Writing converted tile to output/adv.mvt
java.lang.IllegalArgumentException: Invalid number of points in LineString (found 1 - must be 0 or >= 2)
	at org.locationtech.jts.geom.LineString.init(LineString.java:89)
	at org.locationtech.jts.geom.LineString.<init>(LineString.java:79)
	at org.locationtech.jts.geom.GeometryFactory.createLineString(GeometryFactory.java:621)
	at org.locationtech.jts.geom.GeometryFactory.createLineString(GeometryFactory.java:612)
	at com.mlt.decoder.GeometryDecoder.decodeGeometry(GeometryDecoder.java:114)
	at com.mlt.decoder.MltDecoder.decodeMlTile(MltDecoder.java:68)
	at com.mlt.converter.MltCliAdapter.main(MltCliAdapter.java:230)
```

Otherwise this looks good to me.

springmeyer:
> Is it expected that anything that uses the decoder will fail for now?

Yes, expected. There are bugs in the java decoder, so tiles cannot be round tripped currently. The `@Disabled` tests in the java code capture several issues, but this CLI is very good at catching more (hence my recent issues created today).

Issue #45, opened 2024-05-25
Rework dir structure


nyurik:
The repo has become a bit disorganized, and could use a clearer structure. Here are some ideas, but since I don't know all the components of the system yet, its a rough draft that will keep improving until we can start moving dirs around and close this ticket.

I propose we organize by implementation languages plus additional dirs for the specification, shared test files, and shared utilities. Some of the dir names could be improved too.

* `/spec` - "official" documentation describing MLT specification.
  * `/v0` - since we do not have the first version yet, all work will be in `v0` until ready.  Once stable, we will rename this dir to `v1` and won't change it significantly thereafter.
* `/docs` - this is different than `/spec` - it will contain explanations, suggested implementations, performance benchmark results, and anything else useful to the reader.  Some README content will be moved here too.  I propose we use [mdbook format](https://rust-lang.github.io/mdBook/) (already used in Martin and other projects)
* `/test` - anything test-related that is not specific to an individual implementation language
  * `/fixtures/(omt|bing|...) - sample tiles from various sources, with their own licenses. This will be used for validation and performance benchmarks.
  * `/expected/(omt|bing|...)` - if converting MVT to MLT will always produce the same result, we should store that result here - so that every implementation can be validated against it.
* `/java` - JVM-focused implementation (can use java/kotlin/...). Published to Maven?
* `/js` - code for [NPM](https://npmjs.com), probably under `@maplibre/...` name, probably written in TypeScript
* `/rust` - Rust implementation, published [Crates](https://crates.io/), most likely will have bindings for C, C++, Python, and others
* `<more ideas here>`

mactrem:
I like the proposed structure. It's similar to that what i had in mind of how we can structure the repo and also similiar to the structure of [Apache Arrow](https://github.com/apache/arrow). I would go with `js` for the JavaScript decoder

springmeyer:
I also like the overall proposed structure 👍🏼 How about we aim to get CI green with the existing structure (ideally this week) and then with that foundation in place we could re-organize the repo soon after while ensuring all tests keep passing (of course the gh actions workflow will need updated too)?

nyurik:
sounds perfect, thx!  Let me know if you need any help

nyurik:
One related proposal:  I think it would be best to use [just](https://github.com/casey/just) for all orchestration - a user friendly variant of `make` with similar syntax. We have used in in Martin and numerous other projects.

[...]

PR #72, opened 2024-06-03
Introduce new repo structure

springmeyer:
This starts implementing #45.


ebrelsford:
I'm sure you're still working on this but there are some dotfiles in `parser/js/` that probably should have been moved to `js/`

springmeyer:
Yes very much a work in progress. But thanks I'll make sure I get the .dot files.

nyurik:
just to double check on this PR -- shouldn't most of the changes be done before moving the files? This way git will retain the history - i.e. it will be able to track that the file was moved rather than deleted and created somewher else?

As this stands now, seems like lots of deletions and recreations, which is bad because it looses history.

nyurik:
P.S. I do not mean any adjustments to the paths in the CI jobs, etc, to keep it in a testable state.

springmeyer:
@nyurik agree that we should aim for moves rather than deletes + recreations. However some deletions are intentional as I'm taking the liberty to delete code that is old or not fitting in the new layout, like `./evaluation` and `js/src/evaluation`. So those intentional deletions might be what you are seeing. If you see recreations can you give an example and I'll take a closer look?

springmeyer:
> P.S. I do not mean any adjustments to the paths in the CI jobs, etc, to keep it in a testable state.

@nyurik can you share more about what you mean?

springmeyer:
I think this is ready for first review. What I recommend doing is:

 - Ignore looking at the diff at first
 - Rather check the branch out locally and have a look around locally
 - Make sure you can find all the code you'd expect to exist, in the places you'd expect
 - Go into the java or js folders and build + run the tests and make sure things run and work right
 - Comment back here about any problems or breakages and I'll make adjustments
 
Once we confirm that the general layout is okay and the builds + tests are still working for everyone that wants to test, then as a separate followup step let's revisit additional issues before merging:

 - Is it okay to delete the file I deleted (like `evaluation` folders)? If not, where should they be put in the new layout?
 - Are there any files that were moved which git is actually reporting as deletions and recreations?
 - Any changes I made during this re-org that you'd like to see landed as a separate PR (if only to review them easier?)

To solve these final issues I imagine I'll probably re-create a fresh new PR that re-runs the moves more carefully to make sure history is preserved as we'd expect.

mactrem:
I have checked out the branch and tested locally on my machine. The build and tests of the Java project are working. Also the draft of the spec with the referenced documents (images, proto files, ...) is displayed correctly. From my side we can merge it

springmeyer:
Great you both think this is ready to merge @mactrem @ebrelsford - Any concerns around deleted files now if you scan the diff?

@nyurik can you review as well?

ebrelsford:
I don't see any issues with deletions, mostly noticing the redundant js module files getting consolidated, looks fine to me.

nyurik (approved):
awesome work, lets merge!

PR #78, opened 2024-06-05
[WIP] Parse metadata in JS parser tests

ebrelsford:
This is a work in progress but submitting it here in case you have any feedback so far @springmeyer.

So far this:
 * Uses a pbf parser generated by [bufbuild/protobuf-es](https://github.com/bufbuild/protobuf-es)
 * Adds expected metadata for one of the Bing tiles
 * Tests loading that metadata and passing it to the decoder (the decoder just ignores it for now)


js/package-lock.json
```diff
@@ -8,6 +8,9 @@
       "name": "cov-tiles",
       "version": "0.0.1",
       "dependencies": {
+        "@bufbuild/buf": "^1.32.2",
+        "@bufbuild/protobuf": "^1.10.0",
```
springmeyer, line 11: From reading the docs briefly at https://github.com/bufbuild/protobuf-es it looks like `"@bufbuild/protobuf"` is the runtime library such that I think the other deps can be in `devDependencies` to keep the overall library size down.

Also, overall I think using a fully featured library like `bufbuild` for now seems ideal, and then as a future step (when library size matters) we could consider moving using the tiny `pbf` library.

js/package.json
```diff
@@ -25,6 +25,9 @@
     "typescript": "^5.1.3"
   },
   "dependencies": {
+    "@bufbuild/buf": "^1.32.2",
+    "@bufbuild/protobuf": "^1.10.0",
```
springmeyer, line 30: Note: I made a comment above against `package-lock` about looking into switching a few of these to `devDeps`. Meant to make the comment here!

js/test/unit/decoder/covtDecoder.spec.ts
```diff
@@ -85,19 +103,25 @@ function mapMvtProperties(mvtProperties: Map<string, unknown>): Map<string, unkn
...
-    return mappedTiles;
+        .map((mvtFilename) : { mvt: string; mlt: string; mltMetadata: string } => {
+            const mvtPath = Path.join(dir, mvtFilename);
+
+            const mltDir = dir.replace('fixtures', 'expected');
+            const mltFilename = mvtFilename.replace(/\.(pbf|mvt)$/,'.mlt');
+            const mltPath = Path.join(mltDir, mltFilename);
+
+            const mltMetadataFilename = mltFilename + '.meta.pbf';
+            const mltMetadataPath = Path.join(mltDir, mltMetadataFilename);
```
springmeyer: This works, but a high level thought is that the meta.pbf will have, if I'm understanding the metadata format correctly, a 1:N relationship with a set of tiles. So I think the ideal path structure to honestly represent that, and avoid having to store duplicate `.meta.pbf` would be:

> test/fixtures/{tilesource}/meta.pbf

Where `tilesource` is `bing` or `stamen`, etc.

I think, if I'm reading this code right, it is doing `test/fixtures/<tilesource>/z-x-y.meta.pbf`. But for every `z-x-y` the meta data should be the same, right? If not, then of course I'll need to dig into/understand why and this comment will need revisited.

springmeyer (commented):
👍🏼 I like this direction. Made a few very minor inline comments.

js/test/unit/decoder/covtDecoder.spec.ts
ebrelsford: Yeah that's a good point, I was originally thinking of smaller fixtures that were likely to all have different metadata, but in this case it's definitely redundant. I'll make that change.

js/package.json
ebrelsford: Done!

ebrelsford:
@springmeyer From here the path seems (to me) to be really digging into the parser. I have a start on this but I think that's best a separate branch on top of this. You okay with merging this as-is?

springmeyer (approved):
Sounds good to merge + plan core parser work in followup

PR #79, opened 2024-06-05
Add JMH benchmarks for decoding MVT and MLT into in-memory representation

mactrem:
Add benchmarks for the decoding performance of OpenMapTiles schema based tiles into the MVT and MLT in-memory representations. First Java benchmarks are really promising, with an order of magnitude improvement in the decoding performance in the current benchmarks (but has to be double checked with different libraries in particiular with the original Js lib for example):

Benchmark                         Mode  Cnt   Score   Error  Units
MLT decoding Z10  avgt    5   0,346 ± 0,042  ms/op
MVT decoding Z10   avgt    5  11,075 ± 2,051  ms/op

MLT decoding Z12 avgt    5   0,480 ± 0,052  ms/op
MVT decoding Z12  avgt    5   4,905 ± 0,324  ms/op

MLT decoding Z14 avgt    5   1,811 ± 0,270  ms/op
MVT decoding Z14  avgt    5  39,027 ± 1,670  ms/op

MLT decoding Z2   avgt    5   0,910 ± 0,114  ms/op
MVT decoding Z2   avgt    5  11,990 ± 0,793  ms/op

MLT decoding Z4  avgt    5   1,284 ± 0,152  ms/op
MVT decoding Z4   avgt    5  24,963 ± 2,668  ms/op

MLT decoding Z5  avgt    5   1,380 ± 0,178  ms/op
MVT decoding Z5   avgt    5  18,798 ± 1,704  ms/op

MLT decoding Z7   avgt    5   1,082 ± 0,141  ms/op
MVT decoding Z7   avgt    5  22,721 ± 2,873  ms/op

MLT decoding Z9   avgt    5   0,822 ± 0,095  ms/op
MVT decodingZ9   avgt    5  12,141 ± 1,266  ms/op

### DIFF
--- a/java/src/jmh/java/com/mlt/OmtDecoderBenchmark.java
+++ b/java/src/jmh/java/com/mlt/OmtDecoderBenchmark.java
@@ -1,0 +1,365 @@
+package com.mlt;
+
+import com.mlt.converter.ConversionConfig;
+import com.mlt.converter.FeatureTableOptimizations;
+import com.mlt.converter.MltConverter;
+import com.mlt.converter.mvt.ColumnMapping;
+import com.mlt.converter.mvt.MapboxVectorTile;
+import com.mlt.converter.mvt.MvtUtils;
+import com.mlt.decoder.MltDecoder;
+import com.mlt.metadata.tileset.MltTilesetMetadata;
+import com.mlt.vector.FeatureTable;
+import io.github.sebasbaumh.mapbox.vectortile.adapt.jts.model.JtsMvt;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import no.ecc.vectortile.VectorTileDecoder;
+import org.apache.commons.lang3.tuple.Pair;
+import org.openjdk.jmh.annotations.*;
+
+/**
+ * Benchmarks for the decoding performance of OpenMapTiles schema based tiles into the MVT and MLT
+ * in-memory representations.
+ */
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Mode.AverageTime)
+@Threads(value = 1)
+@Warmup(iterations = 5)
+@Measurement(iterations = 5)
+@Fork(value = 1)
+public class OmtDecoderBenchmark {
+  /* java-vector-tile library */
+  private static final Map<Integer, byte[]> encodedMvtTiles = new HashMap<>();
+  /* mapbox-vector-tile-java library */
+  private static final Map<Integer, ByteArrayInputStream> encodedMvtTiles2 = new HashMap<>();
+  private static final Map<Integer, byte[]> encodedMltTiles = new HashMap<>();
+  private static final Map<Integer, MltTilesetMetadata.TileSetMetadata> tileMetadata =
+      new HashMap<>();
+
+  @Setup
+  public void setup() throws IOException {
+    encodeTile(2, 2, 2);
+    encodeTile(3, 4, 5);
+    encodeTile(4, 8, 10);
+    encodeTile(5, 16, 21);
+    encodeTile(6, 32, 41);
+    encodeTile(7, 66, 84);
+    encodeTile(8, 134, 171);
+    encodeTile(9, 265, 341);
+    encodeTile(10, 532, 682);
+    encodeTile(11, 1064, 1367);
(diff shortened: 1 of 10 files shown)


java/src/jmh/java/com/mlt/OmtDecoderBenchmark.java
```diff
@@ -0,0 +1,190 @@
...
+        var mvtFilePath = Paths.get(com.mlt.test.constants.TestConstants.OMT_MVT_PATH, tileId + ".mvt" );
+        var encodedTile =  Files.readAllBytes(mvtFilePath);
+        var decodedTile = MvtUtils.decodeMvt(mvtFilePath);
+        return Pair.of(new ByteArrayInputStream(encodedTile), decodedTile);
+    }
+
+    @Benchmark
+    public void decodeMvtZ2() throws IOException {
+        var mvTile = encodedMvtTiles.get(2);
+        var decodedMvTile = MvtUtils.decodeMvt2Fast(mvTile);
```
springmeyer: @mactrem I don't know much about JVM optimizations, so this may not apply: but I wonder if it might be important to do something with this variable in each benchmark (like asserting on the final size or something like that) to ensure that the JVM optimizer does not do anything to throw away work (since I presume it can detect an unused variable).

java/src/jmh/java/com/mlt/OmtDecoderBenchmark.java
```diff
@@ -0,0 +1,190 @@
...
+        var columnMappings = Optional.of(List.of(columnMapping));
+        var metadata = MltConverter.createTilesetMetadata(encodedMvtTile.getRight(), columnMappings, true);
+        tileMetadata.put(z, metadata);
+
+        var allowIdRegeneration = true;
+        var allowSorting = false;
+        var optimization = new FeatureTableOptimizations(allowSorting, allowIdRegeneration, columnMappings);
+        var optimizations = Map.of("place", optimization, "water_name", optimization, "transportation",
+                optimization, "transportation_name", optimization, "park", optimization, "mountain_peak",
+                optimization, "poi", optimization, "waterway", optimization, "aerodrome_label", optimization);
```
springmeyer: I'm curious: what happens if any of these field names are misspelled or don't exist in the tileset? Will we get an error? I would assume throwing would be ideal to warn the developer that the optimization is not going to work as expected.

springmeyer (approved):
Great stuff! One small non-blocking request: can you explain what some of the benchmark results terms are:

 - `avgt` - average time?
 - `Cnt` - total times the benchmark was run?
 -  etc

And converting the output to a markdown table would make it easier to read: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables

java/src/jmh/java/com/mlt/OmtDecoderBenchmark.java
mactrem: yes, we should get an error. But this is in general a hack which has to be refactored. We should think about a real stable solution of how to map MVT fields to nested MLT fields

java/src/jmh/java/com/mlt/OmtDecoderBenchmark.java
mactrem: good point. I think some time ago i checked it by doing some additional instructions on the returned data and didn't see any significant changes. But i will check this again. One way would be to return a value which is done for example by Lemire (leading in this field) see https://github.com/lemire/microbenchmarks/blob/master/src/main/java/me/lemire/microbenchmarks/bitset/Bitset.java but in Arrow for example the benchmarks are similar see https://github.com/apache/arrow/blob/9ee6ea701e20d1b47934f977d87811624061d597/java/performance/src/main/java/org/apache/arrow/vector/VectorUnloaderBenchmark.java#L31

mactrem:
Yes avg is the average time. There are different metrics to benchmark in JMH for example also the througput see https://jenkov.com/tutorials/java-performance/jmh.html#jmh-benchmark-modes. Cnt is the number of benchmark iterations

springmeyer:
Thanks! Good to merge when you are ready.

nyurik:
@mactrem or @springmeyer I guess its ok to merge - whichever one has more time to shepherd it forward?

mactrem:
@nyurik i will add some additional benchmarks for the missing zom levels and do some minor improvements of the benchmarking then we can merge it

mactrem:
@springmeyer in the macOS tests there is some tiny tile size increase (25 bytes). The regeneration of the fixture tiles for the test is not working on my machine, so maybe we can have a look at. I will now merge it anyway to move forwared and try to fix the broken Ci later with your help

springmeyer:
@mactrem sounds good, PR to fix the CI is up at https://github.com/maplibre/maplibre-tile-spec/pull/138

PR #80, opened 2024-06-05
Minor workflow adjustments

springmeyer:
- Remove macOS runner to save resources (the macOS runner has a [10x minute multiplier](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers), so it is more likely to lead to hitting the 2000 minute limit)
   - Testing on ubuntu is reliable enough predictor of macOS support
 - Remove `fail-fast` option so that we default back to failing fast if any platform fails. This is also useful to save resources.
 - Start allowing workflow runs on any branch or PR. This will allow for better developer flow since @ebrelsford and I are soon planning to target non-main branches in PRs soon as we work to improve the JS decoder.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
     defaults:
       run:
         working-directory: 'java'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,7 @@ name: test
 
 on:
   push:
+    branches: [ "main" ]
   pull_request:
   workflow_dispatch:
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,7 @@ name: test
 
 on:
   push:
-    branches: [ "main" ]
   pull_request:
-    branches: [ "main" ]
   workflow_dispatch:
 
 defaults:
@@ -29,9 +27,8 @@ jobs:
   test-java:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest]
     defaults:
       run:
         working-directory: 'java'


ebrelsford (approved):

.github/workflows/test.yml
```diff
@@ -2,9 +2,7 @@ name: test
 
 on:
   push:
-    branches: [ "main" ]
```
nyurik, line 5: You probably want to keep it here - otherwise you effectively double builds:
* it builds when you push to a fork branch (e.g. most forks will automatically build it when you initially push your code)
* it also builds on pull request

Instead, I would recommend keeping the list of `push` branches to the officially maintained ones (e.g. `main`, `v1.x`, `complex-feature`, etc), but pull requests could in theory be built for everything (i.e. it will build if I create a PR into one of your branch)...     But, tbh, i have not seen a case when these settings didn't work for a project, but there is always a new case of course.

.github/workflows/test.yml
```diff
@@ -29,9 +27,8 @@ jobs:
   test-java:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
```
nyurik, line 34: one more thing - do you foresee slow builds for this repo?  Martin repo for example uses about 1.5hrs for each build, and we have 10-20 a day.

springmeyer: I do foresee potentially running benchmarks in the future, which could take more time. But not 1.5hrs, no! Here I'm trying to be conservative with resources since I'm assuming that the limit of 2000 minutes is across all maplibre org repos and I don't have a sense of the total minutes others take.

.github/workflows/test.yml
springmeyer: Thanks for your thoughts. What I'm after is:

 - builds on pushes to any branch, even forks (now that we are attempting to work in forks)
 - Builds of the merge commit for a PR to branches that are targeting ([feature/update-js-parser](https://github.com/maplibre/maplibre-tile-spec/tree/feature/update-js-parser) and also potentially further PRs "stacked" on top targeting the PR targeting [feature/update-js-parser])

.github/workflows/test.yml
nyurik: In that case I would rather keep it -- `macOS` is a major pain when it comes to Linux compatibility - i.e. lots of shell scripts break because some flags have different meaning.  On the other hand, having macOS target ensures all of our contributors can be certain their code works on every platform - and macOS people won't have surprises if they start contributing.

As for benchmarks - we can easily run them conditionally on linux only when we have them.

.github/workflows/test.yml
springmeyer: Restricted back to main in 7c11123. I think I'll try using manual builds with `workflow_dispatch` for any situations that come up where a CI run is useful and a PR does not exist.

.github/workflows/test.yml
springmeyer: Okay, sounds good 👍🏼

springmeyer: Added back in f6ed6ae. BTW, I normally develop using macOS so I'm also a big fan of keeping this working :) Good to hear you are not worried about the resource use. Easy to revisit if it becomes an issue.

.github/workflows/test.yml
nyurik: Just to understand - you want github action to run on every one of these steps?
* you push your code to your own fork, e.g. `springmeyer:new-feature`
* you create a pull request, attempting to merge `springmeyer:new-feature` into `maplibre:main` or into `maplibre:feature/update-js-parser`
* you merge changes into either of the two branches

The first two are usually identical - so you just run the same CI twice -- once in the context of your own fork, and once in the context of the primary one (but without access to the primary's secrets). That's why in most repos CI runs only on the specific primary branches like `main` and `feature/update-js-parser`, but nothing else.  PRs on the other hand can run on everything.

Note that pushing your changes to your own fork is no different than pushing it to the primary repo -- CI would still not have ran with the current settings. It only ran after you created a PR.

.github/workflows/test.yml
nyurik: OMG, you use Mac? jk :) 
 
Developers and their time and on-boarding experience are far more valuable than a few extra minutes of a `macOS` runner ;)

.github/workflows/test.yml
nyurik: @springmeyer just out of curiosity - do you have an example of such case?  I personally never seen a branch on a fork that has no PR into one of the primary repo's branches, so really curious. Thx!

PR #81, opened 2024-06-05
Update JS parser

ebrelsford:
This is a feature branch for updating the JS parser to account for changes made in the spec and Java encoder over the past ~6 months. The intention is for this branch to be broken and in-progress while smaller PRs are applied to it to get the JS parser functioning again.

So far this branch can read the header of a single layer and the structure is in place for reading each column for a layer. Where possible this branch attempts to keep a structure that aligns with the Java decoder.


ebrelsford:
@springmeyer Here's a start at what we discussed. As we talked about, it seems reasonable for us both to open smaller PRs against this branch from our forks.

springmeyer:
@ebrelsford heads up that last week I added further fixes to this branch. But then I backed out of them because I found that having the old decoder working and unchanged was valuable to me as a reference. So instead I created a fresh decoder alongside it in https://github.com/maplibre/maplibre-tile-spec/pull/120, which is where my last week and this weeks work is landing.

PR #89, opened 2024-06-06
[js][decoder] add intWrapper class + tests

springmeyer:
Started looking into missing functionality for the JS decoder. Java uses this small `intWrapper` class everywhere, so it seems wise to mimic the class in TS. This adds it + tests. Done against `main` instead of #81 since this code is new, working, tested, and simple. Will pull this back into #81 after it lands and I need it there.


springmeyer:
Yes I had the same thought about performance. When we get there we can revisit to ensure no overhead from this.

PR #120, opened 2024-06-11
New JS decoder

springmeyer:
This adds a new JS decoder implementation. This is mostly working, but currently still lacks:

 - FSST and FastPfor decoding support (and therefore geometry and string decoding are not fully working)
 - Support for the vectorized path

Despite not being a 100% working implementation, I've designed this PR to be mergable in its current, partially implemented state. I've done this by writing tests that assert on the expected failure states for where functionality is missing. I will plan followup PRs to address the missing functionality, incrementally.

Design notes:
 - Currently the new code is put into a `decoder2` directory in order to keep the old decoder unchanged as a reference (useful for me at least while developing locally).
 - The directory structure and file naming mirrors closely, for now, the java decoder.


springmeyer:
Despite saying above I think this is mergable in its partial state, definitely still welcome reviews!

- @mactrem - see anything I've mirrored from java that is incorrect?
- @ebrelsford - does this approach of writing tests for expected failures work for you? Seems to me like a nice way to communicate to each other and ourselves about the expected status of things as we work. This way we can also more likely be able to continuously land work into `main` when useful.

ebrelsford (approved):
This is looking great! The tests make sense to me considering where we're at.

Do you think it's still useful to keep the original decoder around?

springmeyer:
> This is looking great! The tests make sense to me considering where we're at.

Great.

> Do you think it's still useful to keep the original decoder around?

Yes. For a moment more. I've been finding that it is a time saver to grep into `src` and see what lights up for keywords. Having to look in two places would slow me down slightly and I'm trying to work fast. I think once I get string decoding working in the new decoder the value of keeping the original around will fade. At that time I'll plan to drop the old one.

mactrem (approved):

Issue #70, opened 2024-06-03
Encoding error in MltConverter.convertMvt:  Max vertex value has to >= 0.


springmeyer:
I found an input tile which triggers a problem during encoding at https://github.com/maplibre/maplibre-tile-spec/blob/9bffeb56e03b7644edf02acb750d20382f8900f5/converter/java/src/main/java/com/mlt/converter/geometry/SpaceFillingCurve.java#L11

```
java.lang.IllegalArgumentException: Max vertex value has to >= 0.
	at com.mlt.converter.geometry.SpaceFillingCurve.<init>(SpaceFillingCurve.java:12)
	at com.mlt.converter.geometry.HilbertCurve.<init>(HilbertCurve.java:9)
	at com.mlt.converter.encodings.GeometryEncoder.encodeGeometryColumn(GeometryEncoder.java:98)
	at com.mlt.converter.MltConverter.convertMvt(MltConverter.java:171)
	at com.mlt.converter.MltCliAdapter.main(MltCliAdapter.java:146)
```

In this case `minVertexValue` is `-1219`.

Test tile that triggered the problem: https://github.com/mapbox/mvt-fixtures/blob/main/real-world/nepal/13-6040-3428.mvt

PR #122, opened 2024-06-11
[java][encoder] curve implementation more tolerant of negative vertex values

springmeyer:
A tentative fix for #70


springmeyer:
This got hit with a faulty merge and I'm not confident this will handle edge cases correctly. Closing and we can discuss the right solution at #70

PR #123, opened 2024-06-11
[WIP] CLI improvements

springmeyer:
As part of working on #120, I've benefited from some enhancements to the CLI tools in order to help test & compare against the java decoder. This PR is a work in progress that I'll continue to use/update as I work on #120 and related followup PRs. 

Currently it includes:

 - Removing the version string of `1.0-snapshot` from the jar to avoid having to type those characters
 - Renaming the `mvt2mlt` to just `encode` which is easier on the eyes
 - Adds an option to request decoding a tile using the `vectorized` pathway
 - Adds a new `decode` tool which will behave more like the JS decoder (which needs to read the TilesetMetadata from disk)
 - Adds a new `meta` tool to generate the tileset metadata


nyurik (approved):
seems reasonable and low risk, merge at will

PR #124, opened 2024-06-11
Format Java and JS code consistently

springmeyer:
This applies google java format to the java code and eslint --fix to the js code. It also starts asserting properly formatted code on the java. It does not yet do this for the JS code (the code has too many errors still).


nyurik (approved):
yeppi!

vinayakkulkarni:
@nyurik: Maybe we can add super linter support in CI checks as well? 

https://github.com/super-linter/super-linter#get-started

nyurik:
@vinayakkulkarni oh wow, i wasn't aware of this. I am all for it - perhaps as a separate PR?

springmeyer:
@vinayakkulkarni I like superlinter and have used it in the past. The challenge I see is that running it locally & manually is not the easiest. So I consider it an additional tool to consider but not a replacement for tooling like this specific to a build system like gradle.

nyurik:
oh, good point, thx. It would be important to have an easy to run command too, so probably not worth setting up a superlinetr.

I usually have the following setup:  in the `justfile`, I have a `fmt` recipe that calls `fmt-java`, `fmt-rust`, etc recipes. This way anyone can run `just fmt` or `just fmt-java` -- and that would re-format that code. Additionally, there would be a `lint`, `lint-java`, ... recipes that can be ran both by CI and by developer, and used to validate that fmt, together with any other linting system (like eslint, clippy, etc) runs.

PR #125, opened 2024-06-12
Update & extend "expected" fixtures

springmeyer:
This addresses several problems with the expected fixtures:

 - We should be testing, in all decoding implementations, decoding both with and without "advanced" encodings. This starts generating a `expected.mlt` (which does not used `advanced` encodings) and an `expected.advanced.mlt` (which does). In a followup PR I'll work to make sure both implementations work with both of these correctly. Currently both implementations have lacking functionality hidden behind disabled tests.
 - We previously started adding a root `mltmetadata.pbf` per tile source (.e.g bing) directory. But the metadata is currently different per tile. So this addresses that problem by generating new metadata per tile and renaming the several `mltmetadata.pbf` the JS tests had started depending on
 - In order to make it easier to view the metadata (and know if/how it may change if the tiles are updated) I've also checked in the yaml-ish representation of it (made with `protoc --decode`)
 - Some tiles fail to encode currently, so this dumps a `.failed` record in the `expected` directory as a signal of the problem


springmeyer:
Note: if folks like this proposed structure, I'll followup by adding a script to this PR to regenerate things. Currently my draft script depends on #123 and looks like:

```bash
trap "exit" INT

set -euo pipefail

tiles=$(find ./test/fixtures -type f \( -name "*.mvt" -o -name "*.pbf" \))

for fixture in ${tiles}; do
    expected_path=$(echo $fixture | sed 's/fixtures/expected/g' | sed 's/\.[^.]*$//')
    mlt_path=${expected_path}.mlt
    meta_path_pbf=${expected_path}.mlt.meta.pbf
    meta_path_dump=${expected_path}.mlt.meta.yaml
    geojson_path=${expected_path}.geojson
    if [ ! -f "${expected_path}.advanced.mlt" ]; then
        echo "Processing: $fixture"
        # temporarily move into the java directory so that the fsst so is found
        cd ./java
        set +e
        RESULT_ERROR=$(java -jar ./build/libs/encode.jar -mvt ../${fixture} -mlt ../${expected_path}.advanced.mlt -advanced 2>&1 >/dev/null)
        RETURN_CODE=$?
        set -e
        cd ../
        if [[ ${RETURN_CODE} != 0 ]]; then
            echo "Failed to encode MLT file: ${RETURN_CODE} $RESULT_ERROR"
            mkdir -p $(dirname ${expected_path}.advanced.mlt.failed)
            echo $RESULT_ERROR > ${expected_path}.advanced.mlt.failed
            continue
        fi
        java -jar ./java/build/libs/meta.jar   -mvt ${fixture} -meta ${meta_path_pbf}
        java -jar ./java/build/libs/encode.jar -mvt ${fixture} -mlt ${expected_path}.mlt
        protoc --decode=mlt.TileSetMetadata spec/schema/mlt_tileset_metadata.proto < ${meta_path_pbf} > ${meta_path_dump}
        # vt2geojson -z 0 -x 0 -y 0 ${fixture} > ${geojson_path}
        # GZIP=-9 tar -czvf ${geojson_path}.tar.gz ${geojson_path}
        # rm ${geojson_path}
    fi
done

failed=$(find ./test/expected -type f \( -name "*.failed" \))
if [ -z "$failed" ]; then
    echo "All fixtures encoded successfully!"
    exit 0
else
    for fixture in ${failed}; do
        echo "Failed to encode MLT file: "
        cat $fixture
    done
fi
```

nyurik:
@springmeyer are you checking in both yaml and pbf?  It seems redundant for each test file to have two representations checked in.  Perhaps just the yaml one should be enough? BTW, those yaml files look very strange - are they actually yaml or something else?

springmeyer:
Only yaml-ish. The 'protoc' tool from Google encodes that format. Unsure what it is exactly. But it is human readable. I'm inclined to not check it in. The pbf is all we need. I can add a script or docs for how to dump the yaml if anyone needs it.

nyurik:
@springmeyer i would actually prefer that yaml-ish output to be checked in -- this way we can see the difference between versions, and it would be useful when reviewing. Checking in binary blobs is almost always the worst option because we cannot easily inspect how they changed with versions, and if the change is valid or not.  But lets rename the files to some other extension so as not to assume they are yaml (or are they?).

springmeyer:
@nyurik renamed from `yaml` to to `txt` in [974ca75](https://github.com/maplibre/maplibre-tile-spec/pull/125/commits/974ca7584da88847b762ec762c57075b6dadd70e). I think `.txt` is reasonable: I cannot find documentation in the protobuf library about the format and the tool that generates it just calls it "text": https://github.com/protocolbuffers/protobuf/blob/bbf52275dbad317bad3421f3c69b5f3759eb2b23/src/google/protobuf/compiler/command_line_interface.h#L385

springmeyer:
As far as whether to check in the pbf or not, here are the considerations I see:

 - The `pbf` format is required throughout the codebase. I think that changing to a different format, while good to consider overall, is out of scope for this PR
 - The `pbf` format can be dynamically generated by the `protoc` command using the txt (formally yaml) as input. So we could consider having the test frameworks in java and js shell out to `protoc` dynamically. But I'm hesitant to do that, as the `protoc` tool is a binary C program and adding tooling to download versions of it seems a little too complicated compared to just checking in stable versions.

nyurik:
These are valid concerns, and at the same time having tons of unreadable and hard to parse opaque blobs in git is pretty bad. Also, I do agree that shelling out is sub-optimal.

Since this is really an integration test (i.e. back-to-back type of test, rather than a small internal test for one of the compressions, etc), how about we simply dump all generated pbfs to the `test/output/java/*` dir, add the `test/output` dir to `.gitignore`, and simply add an extra bash step after running the tests to dump all pbf files as .txt, and use `diff` command to compare txt to the expected values?  I am more than happy to help with these automation steps.

This way we:
* actually see why test result differs in CI logs or in console, instead of simply getting "binary blobs do not match" - and have to run the protoc dumping tool by hand to see why results differ
* are always certain what our tests generate (.txt becomes the source of truth)
* can inspect any changes between versions by looking at git diff
* do not have to worry of yet another set of "unknown binary blobs" (that was the recent XZ attack path - so the fewer bin blobs, the safer we are)
* we can download protoc util on the fly in the CI - possibly as part of the same script that runs it

springmeyer:
> I am more than happy to help with these automation steps.

Go for it! I'll happily hand over this PR to you.

springmeyer:
@nyurik you were going to take a look at this PR. What is the status in your mind? I see:

> [nyurik](https://github.com/nyurik) [force-pushed](https://github.com/maplibre/maplibre-tile-spec/compare/7744178135627701b1a9e6fcd437415e85191d0f..5d5835a5cc16d9f7bf9916096da2d79e83a4ff8d) the dane/fixture-expected-update  branch from [7744178 ](https://github.com/maplibre/maplibre-tile-spec/commit/7744178135627701b1a9e6fcd437415e85191d0f)to [5d5835a ](https://github.com/maplibre/maplibre-tile-spec/commit/5d5835a5cc16d9f7bf9916096da2d79e83a4ff8d)[5 days ago](https://github.com/maplibre/maplibre-tile-spec/pull/125#event-13135116271)

But I'm unsure what you thought the next steps were? Maybe we should voice on this briefly today?

nyurik (approved):
@springmeyer I thought we discussed this in the last video chat?  My eventual goal is to have a simple to use tool to do the conversions - this way we can just store the text versions in git.  In the mean time, I do not want to block these efforts, so feel free to merge.   (i didn't your comment yesterday, my apologies).

springmeyer:
@nyurik thanks for the clarification. I agree too that "this way we can just store the text versions in git." is a desirable goal. I'd like that too and so I'm going to close this and take a look at doing that in a separate PR. The most important point of attention is the input MVT tiles not the metadata, so that is what I'm going to explore storing as only text. Will revisit when I have the attention to give it.

PR #133, opened 2024-06-12
Fix FastPFor JS decoder

ebrelsford:
This PR:
 * Fixes an issue with the FastPFor JS decoder, there was an assumption that an object was passed by reference but a new object was created so the output data was empty
 * Consolidates test case logic
 * Adds more compressed output from the Java encoder (there is no JS encoder to test against), inspired by the [C++ tests](https://github.com/lemire/FastPFor/blob/master/unittest/test_fastpfor.cpp)


ebrelsford:
@Gh0stD3x In case you're interested to look at this, here is a start!

springmeyer (approved):
Awesome work getting this better tested. If issues remain, let's discover them through integrating and testing with real world data next. Ready to merge!

PR #135, opened 2024-06-12
Add justfile with java and js tests

nyurik:
This will allow CI to run exactly the same tests as developers locally.  This also sets up the framework to compare the output of a directory to the expected values. Currently, it gets faked by copying the expected dir to output dir.

Note the `just test-int` -- a good way to compare the results against the expected values. When called, it:

* runs `clean-int-test` to cleanup the `test/output` dir
* runs `test-run-int` to do any kind of output generation (TODO)
* runs `diff-dirs` to compare the content of `test/output` and `test/expected`

Another counterpart to it is to run `just bless` -- which does almost the same, except that it updates the `test/expected` with the new results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ In the following, the size of MLT and MVT files are compared based on a selected
 
 Join the #maplibre-tile-format slack channel at OSMUS: get an invite at https://slack.openstreetmap.us/
 
+### Development
+
+* This project is easier to develop with [just](https://just.systems/man/en/), a modern alternative to `make`.
+* To get a list of available commands, run `just`.
+* To run tests, use `just test`.
+
 ## License
 
 * All project documentation and specification content is licensed under [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/) - Public Domain Dedication.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,58 +13,39 @@ defaults:
 jobs:
   test-js:
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: 'js'
     steps:
       - uses: actions/checkout@v4
+      - uses: taiki-e/install-action@v2
+        with: { tool: just }
       - uses: actions/setup-node@v4
         with:
           node-version-file: 'js/.nvmrc'
           cache: 'npm'
           cache-dependency-path: 'js/package-lock.json'
-      - run: npm ci
-      - run: npm test
+      - run: just test-js
+
   test-java:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-    defaults:
-      run:
-        working-directory: 'java'
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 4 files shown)


HarelM:
Why not simply use make?

nyurik:
Make has far too many "gotchas" and was never designed for task execution. It was designed for large project compilation - i.e. each "recipe" is actually a file target.  Task execution was more of an afterthought - that's why you have to add `.PHONY` (see [reasons](https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile)).

That's why there were so many attempts to get rid of that old clunky system, despite its ubiquity.

Features like self-documentation with comments -- in Just, you just add a `# foo` right above the recipe, and it automagically work -- running `just --list` would show a list of recipes and their documentation.  I won't even go into `:=` vs `=` vs ... ([this](https://stackoverflow.com/questions/448910/what-is-the-difference-between-the-gnu-makefile-variable-assignments-a)).  Just targets many of these ([here](https://just.systems/man/en/chapter_81.html))

P.S. Some of the below comments by @HarelM and I were deleted as irrelevant, per our separate discussion.

HarelM:
I'm not a make expert (more over, I generally hate make and its syntax, but just has the same problem of not knowing what depends on what and its unmaintainable hierarchy).
The reason I asked it, is the development setup, which should be easier if you use make and not a tool you'd need to install.
In any case, I prefer composition over inheritance, and just and make usually force you to use inheritance...

nyurik:
inheritance?  Makefile has gained so much popularity (despite a horrid syntax) specifically because it is so convenient to define **functions**.  Just simply improves on that idea, giving you a clear way to create a function with any optional parameters, and a way to call those functions either before your current task executes, or after. This leads to a fairly robust system of managing large number of small scripts in a single file.

nyurik:
P.S. Just IDE plugins allow you to jump between receips, same as you would in any other code file - so the experience is actually very similar to writing any other procedural code

nyurik:
"before" does not mean you go backwards.  Imagine a function `foo` which calls `funcA`, `funcB`, then does some foo-specific stuff, and then does some "cleanup" - `funcC`.  This is the typical pattern here.  `foo` would very often be at the top, whereas "sub-functions" would be below it.  Almost identical to the regular code in that sense.

The result is the ability to say:  I need to test. In order to do that, first i preinstall things, then run build, then I generate some test data, and after all that, the actual "meat" of the task is to test.

README.md
```diff
@@ -43,6 +43,12 @@ In the following, the size of MLT and MVT files are compared based on a selected
 
 Join the #maplibre-tile-format slack channel at OSMUS: get an invite at https://slack.openstreetmap.us/
 
+### Development
+
+* This project is easier to develop with [just](https://just.systems/man/en/), a modern alternative to `make`.
```
mactrem, line 48: In the future we should add a separate CONTRIBUTING.md to address the development. I have setup an initial one for the java project

nyurik: agree, a guide like that would be helpful, but I don't think we are there yet - because most likely it will be out of date right away

PR #138, opened 2024-06-13
Update expected tile size in cli tests

springmeyer:
/cc @mactrem 

To make this easier to update in the future, I'll plan to port these CLI tests to proper unit tests.

PR #140, opened 2024-06-13
Fix update exp tilesize

nyurik:


springmeyer (approved):

springmeyer:
@nyurik is your thinking to merge this into #138 and then approve #138 and then merge that into `main`? Seems like you could also just go right into `main` and then we can close #138.

PR #141, opened 2024-06-13
Validate js & java code as part of the CI

nyurik:


.github/workflows/test.yml
```diff
@@ -22,6 +22,8 @@ jobs:
           node-version-file: 'js/.nvmrc'
           cache: 'npm'
           cache-dependency-path: 'js/package-lock.json'
+      - name: Lint JS code. If fails, run `just fmt-js` to reformat, and `just lint-js` to see issues locally
+        run: just lint-js
```
springmeyer, line 26: Can linting be moved to a separate job? It is best that we see whether tests are passing or failing and not have them hidden behind a linting error.

springmeyer (changes requested):

.github/workflows/test.yml
```diff
@@ -22,8 +22,23 @@ jobs:
           node-version-file: 'js/.nvmrc'
           cache: 'npm'
           cache-dependency-path: 'js/package-lock.json'
+      - name: Lint JS code. If fails, run `just fmt-js` to reformat, and `just lint-js` to see issues locally
```
springmeyer, line 25: ❤️  So friendly to other humans. Nice work

springmeyer:
@nyurik thanks for moving the java linting to separate job. How about moving js linting to separate job too? Or are you worried about too many jobs?

Also, is it intended that the js linting is still a NO-OP? I see it is just `TODO: Add js lint command (e.g. eslint)`. Currently this likely makes sense given we've not fixed all the linting errors yet in the old js code (and probably should instead invest in removing the code rather than linting it).

nyurik:
@springmeyer yes, i would prefer fewer jobs for this type of thing -- one is usually enough to do the basics like linting, plus another to do the tests.  The number of jobs will grow soon with all other languages, so keeping it a bit more tidy.

Yes, JS is meant as a noop but it will be easy to get it to run by just fixing the TODO and cleaning up JS code

PR #151, opened 2024-06-14
Add mvt-fixtures submodule

nyurik:
This adds [mvt-fixtures repo](https://github.com/mapbox/mvt-fixtures.git) as a submodule `/test/mvt-fixtures`

It might be a good idea to add a few symlinks to the needed directories directly into the `/test/fixtures` -- TBD.

TODO:
* [ ] use new test fixtures in our integration tests

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -130,7 +130,6 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with: { persist-credentials: false }
-      - run: git submodule update --init --depth 1 test/mvt-fixtures
       - uses: BRAINSia/free-disk-space@7048ffbf50819342ac964ef3998a51c2564a8a75 # v2.1.3
         with:
           tool-cache: false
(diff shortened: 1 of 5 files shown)


springmeyer:
@nyurik thanks for putting this up for discussion. My take is that we should hold off on integrating the mvt-fixtures repo (and any other new fixtures like #159) until one of us has the time/interest/ability to integrate the new fixtures into the tests.  So, I'd recommend closing this for now and pulling in new fixture data can happen in a separate PR that updates both Java and JS to use the fixtures.

nyurik:
Sure, lets hold off on this until ready

nyurik:
We are clearly now at the point where the canonical set of MVT tiles would be useful to support

CommanderStorm (approved):

HarelM:
Does this PR test anything besides adding this submidule?
Approved otherwise.

nyurik:
@HarelM adding the submodule doesn't do much, but the moment i add a symlink to `/test/fixtures/real-data -> real-data`, testing becomes much slower (30MB turns out to take a long time with current tests).  I am still researching why this is the case, but once fixed, it should be there. The result is that all `**/*.mvt` in the fixtures dir will be tested, increasing the coverage (in theory)

PR #153, opened 2024-06-14
Improve hilbert/morton lookup performance

msbarry:
Testing out performance converting a ~700kb mvt to mlt, the converter was spinning forever. Visual VM showed it was from treeset/treemap submap `size()` calls:

image

This change creates a sorted primitive array first, then does binary search for each element to get the index rather than the more expensive `treeMap.headMap(value).size()` calls.  With this change it takes ~500ms


springmeyer:
Awesome insight & fix! This did not change overall tile sizes (based on the testing there) so no concern about merging this.

springmeyer (approved):

PR #154, opened 2024-06-14
Make JS able to be installed as a module

ebrelsford:
In order to use the JS decoder as a module, we need to:
 * Add a `prepare` step to compile the typescript files locally on install
 * Avoid explicitly referencing `src/` since we want the built code to use other built files
 * Make the package a module
 * Add a `main` entrypoint that exports something to be used by other code

This is a WIP while I build a library that depends on it.


ebrelsford:
@springmeyer Feeling provisionally pretty good about this as I've been able to use it locally, anything you want me to change?

springmeyer (approved):
Looks great, ready to merge.

PR #156, opened 2024-06-14
Swap js decoders

springmeyer:
The new JS decoder is coming along (https://github.com/maplibre/maplibre-tile-spec/pull/155). I'm now ready to swap it out for the old. This PR therefore removes the old decoder (which has been broken for 6+ months) with the newly working JS decoder. Previously it was useful to have them live side by side (for last week) to easily reference between them. Now that is no longer the case, so removing the old is a necessary clean up.


nyurik (approved):
seems benign :)

ebrelsford (approved):
Tests are failing because `src/decoder/DecodingUtils.ts` and `src/decoder/decodingUtils.ts` both exist, otherwise looks good to me.

PR #167, opened 2024-06-20
Add compression ratio benchmarks

mactrem:
- Fix sorting of the features and the regeneration of the ids to improve the compression ratio in some cases
- Simplfiy tests in the MltDecoderTest class by using parameterized tests


java/src/main/java/com/mlt/converter/encodings/GeometryEncoder.java
```diff
@@ -234,18 +242,21 @@ public static Triple<Integer, byte[], Integer> encodeGeometryColumn(
     var mortonDictionaryEncodedSize =
         encodedMortonEncodedDictionaryOffsets.encodedValues.length
             + encodedMortonVertexDictionary.encodedValues.length;
+
     if (plainVertexBufferSize <= dictionaryEncodedSize
         && plainVertexBufferSize <= mortonDictionaryEncodedSize) {
       // TODO: get rid of extra conversion
+      // TODO: refactor to use sorted points vertex buffer
```
springmeyer, line 249: Curious if you can say more about what would be involved here in "refactor to use sorted points vertex buffer"?

springmeyer (approved):
Looks great. Just one non-blocking question above I'm curious about.

Great refactoring of the tests too. I have a similar PR in flight that refactors MltDecoderTest. But lets get this landed first and I will adapt my work on top.

java/src/main/java/com/mlt/converter/encodings/GeometryEncoder.java
mactrem: The comment is not valid anymore, i forgot to remove, Thanks for the hint!

springmeyer: Okay, I can remove the TODO in a followup PR I'm working on now.

PR #168, opened 2024-06-20
Improve the java decoder tests

springmeyer:
Currently we support both a "Non-vectorized (aka sequential)" decoding path and a  "Vectorized" decoding path. In the future we can and should consolidate to just one (vectorized), but for the moment we need to keep two because the JS decoder is targeting the non-vectorized path.

With that in mind we need to ensure we are testing the non-vectorized java decoder.

So, this PR makes improvements to the decoder tests to:

 - Test both decoding paths as much as possible
 - The non-vectorized path has more issues, so temporary accommodations are made to allow tests to pass by asserting on expected errors (this gets the code running at least until we can fix its behavior)
 - It is not going to be obvious in the diff, but this also starts testing some tiles which used to fail in the vectorized path, but no longer do (after #158 was fixed)
 - Finally, by asserting on the number of expected errors returned for known failures, this starts allowing the tests to iterate through more features during the comparison which ensures the tests are not hiding additional errors behind other asserts which fail


java/src/test/java/com/mlt/TestUtils.java
```diff
@@ -80,24 +85,49 @@ public static void compareTilesSequential(MapLibreTile mlTile, MapboxVectorTile
       for (com.mlt.data.Feature mvtFeature : mvtFeatures) {
         var mltFeature =
             mltFeatures.stream().filter(f -> f.id() == mvtFeature.id()).findFirst().get();
+
         assertEquals(mvtFeature.id(), mltFeature.id());
 
         var mltGeometry = mltFeature.geometry();
         var mvtGeometry = mvtFeature.geometry();
-        assertEquals(mvtGeometry, mltGeometry);
+        if (!mvtGeometry.toString().equals(mltGeometry.toString())) {
```
mactrem: Do you get any additonal error with this change?
The Geometry classes used for the equlity tests are from the JTS Topology Suite library, which does proper geometry equality comparison based on the specific geometry types and  is battle tested :) Since we are provding no tolerance values for the current equality comparision we are also not safe for floating point rounding errors like in the String variant but we could introduce a value for this to be more robust.
By using `toString` JTS converts it to WKT. 
What are your thoughts on this?

java/src/test/java/com/mlt/decoder/MltDecoderTest.java
```diff
@@ -1,130 +1,136 @@
...
+            + " but there were "
+            + numErrors
+            + " errors");
   }
 
-  private static Stream<Triple<Integer, Integer, Integer>> omtTileIdProvider() {
-    return Stream.of(
-        Triple.of(2, 2, 2),
-        Triple.of(3, 4, 5),
-        Triple.of(4, 8, 10),
```
mactrem, line 49: My thoughts on this were -> by using int values ​​we can dynamically change the file name, e.g. change the separator of xyz in an easy way as i have tiles lying around that have different separators.   
This change is ok for me, but maybe we can agree on a common file naming structure for all tiles so using "_" or "-" for example

java/src/test/java/com/mlt/TestUtils.java
springmeyer: Yes, I also tried with `if (!mvtGeometry.equals(mltGeometry)) {` and I get 7 exceptions like:

```
MltDecoderTest > Decode OMT Tiles > com.mlt.decoder.MltDecoderTest.decodeOMTTiles(String)[13] FAILED
    org.locationtech.jts.geom.TopologyException: side location conflict [ (2639.311475409836, 3217.967213114754, NaN) ]
        at app//org.locationtech.jts.geomgraph.EdgeEndStar.propagateSideLabels(EdgeEndStar.java:289)
        at app//org.locationtech.jts.geomgraph.EdgeEndStar.computeLabelling(EdgeEndStar.java:125)
        at app//org.locationtech.jts.operation.relate.RelateComputer.labelNodeEdges(RelateComputer.java:325)
        at app//org.locationtech.jts.operation.relate.RelateComputer.computeIM(RelateComputer.java:125)
        at app//org.locationtech.jts.operation.relate.RelateOp.getIntersectionMatrix(RelateOp.java:109)
        at app//org.locationtech.jts.operation.relate.RelateOp.relate(RelateOp.java:54)
        at app//org.locationtech.jts.geom.Geometry.relate(Geometry.java:1035)
        at app//org.locationtech.jts.geom.Geometry.equalsTopo(Geometry.java:1089)
        at app//org.locationtech.jts.geom.Geometry.equals(Geometry.java:1057)
        at app//com.mlt.TestUtils.compareTilesSequential(TestUtils.java:93)
        at app//com.mlt.decoder.MltDecoderTest.testTile(MltDecoderTest.java:158)
        at app//com.mlt.decoder.MltDecoderTest.decodeOMTTiles(MltDecoderTest.java:122)
```

Do you also see that locally?

Try changing from:

```
assertEquals(mvtGeometry, mltGeometry);
```

To:

```
import static org.junit.jupiter.api.Assertions.assertTrue;
assertTrue(mvtGeometry.equals(mltGeometry), "MVTGeom != MLTGeom");
```

java/src/test/java/com/mlt/decoder/MltDecoderTest.java
springmeyer: Ah, okay. Yes, how about we standardize on `-` as the separator? `z-x-y.mvt`? I can make the change in this PR to rename any non-conforming tiles. Does that sound good?

springmeyer: > I can make the change in this PR to rename any non-conforming tiles. Does that sound good?

Actually, it's a rather large change to do the renaming. I'd like to do in a followup PR instead of this one as it will require updating JS side as well.

java/src/test/java/com/mlt/TestUtils.java
springmeyer: For now in this PR I've reverted this custom handling: `[cb75bdc](https://github.com/maplibre/maplibre-tile-spec/pull/168/commits/cb75bdc075ceec0d37dcfdf227177c657d41c42e)`

Let's discuss larger issue at https://github.com/maplibre/maplibre-tile-spec/issues/178

java/src/test/java/com/mlt/decoder/MltDecoderTest.java
springmeyer: ticketed at https://github.com/maplibre/maplibre-tile-spec/issues/179 to ensure we agree and then I can get a PR going after.

mactrem: Sounds good

mactrem (approved):

PR #172, opened 2024-06-20
Improve geometry handling in JS decoder

springmeyer:
This PR overhauls the geometry storage in the JS decoder in order to be able to correctly handle complex geometries like polygons with rings and multi-geometries.

Specifically:
 - GeoJSON support is finished in order to use GeoJSON as the testing format.
 - "@mapbox/point-geometry" and a `feature.loadGeometry` API is added in order to make the geometries compatible with Maplibre GL JS
 - The `js/src/mvtUtils.ts` was only being used inside tests, so the code was moved
 - Test coverage was improved for geometry decoding


js/src/data/Projection.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class Projection {
+    size: number;
+    x0: number;
+    y0: number;
+    s1: number;
+    s2: number;
+
+    constructor(x: number, y: number, z: number) {
+        const extent = 4096;
```
nyurik: shouldn't this be per layer rather than hardcoded?

nyurik (approved):
one uncertain thing, but seems ok

js/src/data/Projection.ts
springmeyer: Nice catch! I was meaning to fix this, but had forgotten. Solved in 100644

springmeyer: /cc @ebrelsford `feature.extent` is now available.

PR #174, opened 2024-06-21
Run JS Bench on CI

springmeyer:
This gets the JS benchmark script running on CI. It is set to run fast (limited to 2 seconds per test) as the purpose of running on CI currently is just to ensure the script stays working and does not regress.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,24 @@ jobs:
       - name: Lint JS code. If fails, run `just fmt-js` to reformat, and `just lint-js` to see issues locally
         run: just lint-js
       - run: just test-js
+
+  bench-js:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: taiki-e/install-action@v2
+        with: { tool: just }
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '17'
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -26,6 +26,25 @@ jobs:
...
+          distribution: 'temurin'
+          java-version: '17'
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: 'js/.nvmrc'
+          cache: 'npm'
+          cache-dependency-path: 'js/package-lock.json'
+      # currently requires java encoder to be run
+      - name: Bench JS code
+        run: just bench-js
```
springmeyer, line 46: Because the JS requires Java to be set up and the encoder to be compiled it seemed worth its own job, so as not to slow down the JS test job with java install+compile.

js/bench/decode.ts
```diff
@@ -31,6 +31,12 @@ tiles.forEach(tile => {
   }
 });
 
+let maxTime = 10;
+if (process.env.GITHUB_RUN_ID) {
+  maxTime = 2;
+  console.log(`Running in CI, using smaller maxTime: ${maxTime} seconds`);
```
springmeyer, line 37: This keeps the script runtime nice and swift, faster than the Java setup required for the encode.jar to be available.

nyurik (approved):

PR #177, opened 2024-06-21
Move JS package configuration to root

ebrelsford:
This PR
 * Moves JS package configuration to the root of the repo
 * Updates anything that depends on the cwd to recognize the new location of package configuration

This is helpful because installing this library as a dependency in another JS library (using `npm install maplibre/maplibre-tile-spec`) assumes there is a `package.json` in the root. Without this change, the library using this one would need to install, then do extra work to build it.


springmeyer:
@ebrelsford this looks good if this is our best way forward. On thought: would the alternative be to keep things in `js` and start publishing a version to npmjs? Would that suffice? Thinking out loud: I guess the deeper question is whether being able to install from GitHub URL is important for this project. If not, then just moving to starting to publish an alpha release (maybe `0.0.1-alpha1` or something?) would be worth considering.

ebrelsford:
> is being able to install from GitHub URL is important for this project?

Nope!

>  If not, then just moving to starting to publish an alpha release (maybe 0.0.1-alpha1 or something?) would be worth considering.

I hadn't considered that an option at this point in the project but I'd love to explore that if we think it's a reasonable path. @nyurik who should we talk to at Maplibre about publishing this on npm if that sounds reasonable? Ideally it would be published in a way that we'd be able to make a few new releases as we develop over the next ~week at least.

nyurik:
@HarelM probably knows the most about npm publishing. My preference would be to publish it under maplibre/ org name

HarelM:
In maplibre-gl-js and maplibre style spec there is an automated way to publish a package when you change the package.json version.
This CI logic can be easily copied here if needed. The npm token is saved in the org secrets and I can add it to this repo if/when needed. 
I guess the most tricky part is choosing the name 😃
Ping me if you need anything specific, github tagging will usually get my attention.

HarelM:
P.S. nit-picking: not all ts files have the same casing when it comes to file name.

ebrelsford:
> This CI logic can be easily copied here if needed. The npm token is saved in the org secrets and I can add it to this repo if/when needed.

@HarelM That makes sense, it would be really helpful to have this in place if it's not too much trouble.

>  I guess the most tricky part is choosing the name

I think for now sticking with `maplibre-tile-spec` seems reasonable to me, with the expectation that some parts of the repo/package may eventually spin out into their own namespaces. But I'm not too opinionated.

HarelM:
Most new packages names are `@maplibre/...`.
Please go over the process in maplibre-gl-js or maplibre-style-spec of creating a bump PR version action and the main CI logic of testing for a version change.
If this flow suites for this repo as well, I can copy it over assuming publishing using dry-run is working as expected.
Bottom line, I'd love to help, please let me know when everything is ready by opening the right issue with all the information, so I'd know that I did it correctly.

Issue #178, opened 2024-06-21
Geometry equality in Java tests


springmeyer:
## Context

JTS geometries are used for the Java geometry representation. JTS geometries have sophisticated equality methods as is needed given the complexity of geometry storage: 

See:

 - https://docs.geotools.org/latest/userguide/library/jts/equals.html.
 - https://github.com/locationtech/jts/blob/master/modules/core/src/test/java/org/locationtech/jts/geom/GeometryImplTest.java#L126-L139

Currently in our tests we use Junit `assertEqual`.

## Problem

The results of `assertEqual(mltgeometry, mvtgeometry)` do not match always match `assertEquals(true, mltgeometry.equals(mvtgeometry))`

Which raises the question of what `assertEqual` is doing under the hood? Is it converting to strings to compare objects? Is it using `geom.equals`? Something else?

## Next steps

Overall goal here is to ensure that our Java tests are correctly asserting that MVT geometries can be perfectly round-tripped between MVT and MLT. In other words, that their representations are identical after both are decoded into a given in-memory representation.

I think this is already the case, but we should double-check/ensure it is by ensuring our tests are correctly asserting on equality.

So, how about?

- [ ] Move to `assertEquals(true, mltgeometry.equals(mvtgeometry))` or `assertTrue( mltgeometry.equals(mvtgeometry))`?
   - [ ] Fix failures that arise
- [ ] Or consider some other approach?

 

/cc @mactrem @ebrelsford

springmeyer:
Perhaps also @msbarry has some insight here? Testing issues that lead to this ticket came from https://github.com/maplibre/maplibre-tile-spec/pull/168#discussion_r1649090387

springmeyer:
Started to look into this a bit. Found that `mvtGeometry.equals(mltGeometry)` throws while `mvtGeometry.equalsExact(mltGeometry)` does not throw (and passes). The difference is that the former checks for `topological equality` while the latter checks for `structural equality`. I think what may be happening is that the geometries are equal/identical (and therefore the `structural equality` test passes as it should) but they are invalid in some way so that `topological equality` test cannot cope and fails to finish.

springmeyer:
> they are invalid in some way

Visualizing the WKT https://wktmap.com in shows that most of the failing geometries are polygons which have collapsed to zero area lines. So this problem appears to be coming from either the input geometries (likely) or the handling of them in https://github.com/ElectronicChartCentre/java-vector-tile.

[...]

PR #180, opened 2024-06-21
Test additional code paths in java decoder

springmeyer:
Currently we do not fully test all these combinations:

 - Vectorized path + advanced encodings
 - Vectorized path + without advanced encodings
 - Non-Vectorized path + advanced encodings
 - Non-Vectorized path + without advanced encodings
 
This PR starts testing all those paths.

The code coverage changes are:

 - **overall:** `36%`to `45%`
 - **mlt.converter:** `64%` to `85%`
 - **mlt.decoder:** `55%` to `71%`

In addition this PR includes:

 - Some refactoring in TestUtils.java to share more code between vectorized and non-vectorized
 - Improvements to TestUtils.java around geometry comparisons (refs #178)
 - Improvements to TestUtils.java around property comparisons

Finally, this PR does not attempt to solve any bugs. Rather it finds, isolates, and writes tests harnesses to capture them in code. Once bugs are solved in the future it should be easy to update the tests accordingly.

TODO:
 - [x] ~This PR uncovered a number of bugs that do not have open issues and did not have tests that hit them until now. So I need to create following issues to capture fixing these.~
     - Added:
        - #181
        - #182 
        - #183 
        - #184
        - #185
        - #186
 - [ ] Property nesting in MLT makes comparison to MVT difficult so I disabled it in this test refactor because comparison to MVT is what the decoder tests rely on. My sense is that standalone unit tests of nesting might be more appropriate than including nesting in the decoder tests. Do others agree?


java/src/test/java/com/mlt/TestUtils.java
```diff
@@ -3,123 +3,95 @@
...
+        // TODO: remove the below special case once this bug is fixed:
+        // https://github.com/maplibre/maplibre-tile-spec/issues/181
+        if (mvtPropertyKey.equals("id")) {
+          continue;
+        }
+        var mltProperty = mltProperties.get(mvtPropertyKey);
+        if (mltProperty == null) {
+          // System.out.println(mvtFeature.id() + " mlt prop missing for " + mvtProperty.getKey());
+          // System.out.println("  mvtProperties: " + mvtProperties);
+          // System.out.println("  mltProperties: " + mltProperties);
```
springmeyer, line 52: Left in intentionally since these debug prints can be handy when trying to understand a failure. But commented so the current tests will not be extremely noisy (rather the number of errors is asserted on to keep things calm and readable)

java/src/test/java/com/mlt/TestUtils.java
```diff
@@ -3,123 +3,95 @@
...
+          numErrors++;
+        } else if (!mltProperty.equals(mvtProperty.getValue())) {
+          // System.out.println(
+          //     "Failure comparing property "
+          //         + mvtProperty.getKey()
+          //         + " for feature: "
+          //         + mvtFeature.id());
+          // System.out.println("    mvtProperty: " + mvtProperty.getValue());
+          // System.out.println("    mltProperty: " + mltProperty);
```
springmeyer, line 61: Left in intentionally since these debug prints can be handy when trying to understand a failure. But commented so the current tests will not be extremely noisy (rather the number of errors is asserted on to keep things calm and readable)

java/src/test/java/com/mlt/TestUtils.java
msbarry: You could also consider returning a list of strings with these error messages then at the end `assertTrue(list.isEmpty(), String.join("\n", list))` or something like that so the detailed errors are printed only on failure.

java/src/test/java/com/mlt/TestUtils.java
```diff
@@ -3,123 +3,95 @@
...
+  private static int compareFeatures(
+      List<Feature> mltFeatures, List<Feature> mvtFeatures, boolean isFeatureTableSorted) {
+    int numErrors = 0;
+    for (var j = 0; j < mltFeatures.size(); j++) {
+      var mltFeature = mltFeatures.get(j);
+      var mvtFeature =
+          isFeatureTableSorted
+              ? mvtFeatures.stream().filter(f -> f.id() == mltFeature.id()).findFirst().get()
+              : mvtFeatures.get(j);
+      assertEquals(mvtFeature.id(), mltFeature.id());
```
msbarry: You probably want to include a message here indicating which item number had a different ID?

java/src/test/java/com/mlt/TestUtils.java
```diff
@@ -3,123 +3,95 @@
...
 import com.mlt.vector.FeatureTable;
-import java.util.Map;
+import java.util.ArrayList;
+import java.util.List;
 
 public class TestUtils {
+
+  private static int compareFeatures(
+      List<Feature> mltFeatures, List<Feature> mvtFeatures, boolean isFeatureTableSorted) {
+    int numErrors = 0;
```
msbarry, line 16: start by asserting sizes are the same?

java/src/test/java/com/mlt/TestUtils.java
msbarry: Also is ID equivalent to mapbox vector tile feature ID? That's optional and often omitted or reused across several features in a tile.

java/src/test/java/com/mlt/decoder/MltDecoderTest.java
```diff
@@ -27,71 +28,100 @@ enum DecoderType {
   VECTORIZED
 }
 
+enum EncodingType {
+  BOTH,
+  NONADVANCED,
+  ADVANCED
+}
+
+final class DecodingResult {
```
msbarry: Minor: could also do `record DecodingResult(int numErrors, int numErrorsAdvanced) {}` to skip some boilerplate

java/src/test/java/com/mlt/decoder/MltDecoderTest.java
```diff
@@ -27,71 +28,100 @@ enum DecoderType {
...
+    var tileId = "4-8-5";
+    var result =
+        testTile(tileId, TestSettings.BING_MVT_PATH, DecoderType.BOTH, EncodingType.ADVANCED, true);
+    assertEquals(
+        1148,
+        result.numErrorsAdvanced,
+        "Error for " + tileId + "/advanced: " + result.numErrorsAdvanced);
+  }
+
   private static Stream<String> bingProvider() {
```
msbarry, line 56: Do you want to list a subset of tiles in the bing directory? It's also possible to list all of those files so whenever you drop a new one in it picks that up automatically.

java/src/test/java/com/mlt/TestUtils.java
springmeyer: @msbarry nice suggestion. I'll consider adding that in a followup PR if this need persists. Right now I'm hoping we'll be able to solve a lot of these issues soon and revert back to asserting on the values without any debugging assistance.

java/src/test/java/com/mlt/TestUtils.java
springmeyer: Nice catch: added in 3bdf608

java/src/test/java/com/mlt/TestUtils.java
springmeyer: > You probably want to include a message here indicating which item number had a different ID?

Done in 3bdf608

> Also is ID equivalent to mapbox vector tile feature ID? That's optional and often omitted or reused across several features in a tile.

Yes, this is the mapbox vector tile spec ID. In our case it defaults to zero if not set, so I think this is a valid comparison (even if its only asserting that 0 == 0).

java/src/test/java/com/mlt/decoder/MltDecoderTest.java
springmeyer: Slick, thanks for teaching me about `record`. Done in 1f69099

java/src/test/java/com/mlt/decoder/MltDecoderTest.java
springmeyer: Yes! Eventually that is where I'd like to get. Right now I'm just trying to get a better sense of the errors with the current tiles. But in the future I'd like both the JS tests and the Java tests (and any other language that comes along) to do the same loop over the same test of tiles.

If you are curious and able, I'd love to see a PR that adds the looping and checks if any new failures pop up. That's not something I currently have bandwidth for, but would be super helpful to start on.

java/src/test/java/com/mlt/TestUtils.java
nyurik: Just to double check - @msbarry are you certain `String.join(...)` would not run on every pass?  In most languages, a regular function call like `assertTrue` would get all its params computed before calling it (some languages like Rust use `macro` constructs - i.e. they do pre-compilation step to wrap the String.join in the if statement)

msbarry: Got it! Seems like when the tests are all passing this can all be simplified down a bit and just have assertions instead of tracking the number of failures.

@nyurik String.join should run first then the result message with the diffs would get passed to assertTrue so that it can fail with a more detailed error message.

java/src/test/java/com/mlt/TestUtils.java
```diff
@@ -3,123 +3,111 @@
...
-
-              assertEquals(mvtPropertyValue, mltValue);
-            }
-          } else {
-            assertEquals(mvtFeature.properties().get(mltPropertyKey), mltPropertyValue);
-          }
-        }
+      var mltFeatures = new ArrayList<Feature>();
+      while (featureIterator.hasNext()) {
+        mltFeatures.add(featureIterator.next());
```
msbarry, line 89: minor: could simplify to `var mltFeatures = new ArrayList<>(featureTable);` but also why not just do `numErrors += compareFeatures(featureTables[i], mvtFeatures, isFeatureTableSorted);` and skip making the copy?

java/src/test/java/com/mlt/TestUtils.java
msbarry: What if multiple features have the same ID? Seems like the `!isFeatureTableSorted` would handle that, but not when feature table is sorted. Might be a bigger change though...

msbarry (approved):
Looks good! A few minor comments TIOLI

java/src/test/java/com/mlt/TestUtils.java
nyurik: @msbarry thx, that's what i suspected - thus its a really bad pattern (in general) -- because a lot of slow computation (string manipulation/mem alloc) gets done every time even if assertion would not be triggered. That said, in this case it is obviously not a huge deal because if the list is empty, there is only one memory allocation (empty string).  In Rust this would not even be a mem alloc (rust does not allocate empty vectors), but in other langs there might be an alloc of some small memory -- I don't know what JVM does in this case.  Thx for checking!

java/src/test/java/com/mlt/TestUtils.java
springmeyer: > What if multiple features have the same ID? 

Right now that is okay & tolerated.

> Seems like the !isFeatureTableSorted would handle that, but not when feature table is sorted. Might be a bigger change though...

Right, there is a known limitation right now - ticketed at https://github.com/maplibre/maplibre-tile-spec/issues/186.

java/src/test/java/com/mlt/TestUtils.java
springmeyer: If I'm understanding right, it's not just a copy, it's a decoding into `List<Feature>` (the `featureIterator.next()` has to be called in order to trigger feature decoding). In general, there are two different decoding paths and they are both turned into `List<Feature>` here in order to have the same code run asserts on the output of the two paths.

springmeyer:
> Looks good! A few minor comments TIOLI

Thanks @msbarry! Replied to your comments. I'll merge this shortly.

springmeyer:
As intended & expected, this failed after pulling in #184:

```
MltDecoderTest > Decode Bing Tiles > com.mlt.decoder.MltDecoderTest.decodeBingTiles(String)[1] FAILED
    org.opentest4j.AssertionFailedError: Error for 4-8-5/non-advanced: 0 ==> expected: <2> but was: <0>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:563)
        at app//com.mlt.decoder.MltDecoderTest.decodeBingTiles(MltDecoderTest.java:70)
```

Which is fixed by updating the test expectation in 7123203

java/src/test/java/com/mlt/TestUtils.java
msbarry: Oh right I see what you're getting at. There's a version of assertEquals that takes a supplier so you can do `assertEquals(expected, actual, () -> String.join(...))` to defer generating the message if it is too expensive. https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertEquals-byte-byte-java.util.function.Supplier-

PR #195, opened 2024-06-25
Add JS release workflows based on maplibre-gl-js

ebrelsford:
As suggested in #177, we could start publishing prerelease versions of the JS module in order to use it downstream.

This PR:
 * Attempts to duplicate the relevant GitHub Actions workflows from `maplibre-gl-js`
 * Adds `@maplibre/` to the name


ebrelsford:
@HarelM I attempted to copy over the relevant GitHub Actions work here as you mentioned in #177. Would this work for you? If so I think I'll mostly need the secrets to be configured so we can publish.

nyurik:
I just added `NPM_ORG_TOKEN` to this repo, you may need to restart CI

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+        working-directory: js
+        id: create_regular_release
+        if: ${{ steps.prepare_release.outputs.release_type == 'regular' }}
+        uses: ncipollo/release-action@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag: ${{ steps.tag_version.outputs.new_tag }}
+          name: ${{ steps.tag_version.outputs.new_tag }}
+          bodyFile: ${{ steps.release_notes.outputs.release_notes }}
```
HarelM: Is this defined? The creation of release notes...?

ebrelsford: It is not currently, I removed other steps that referred to it and missed this one.

HarelM (approved):
I guess the dry-run is intended, right?

ebrelsford:
> I guess the dry-run is intended, right?

For now yes, didn't want to inadvertently publish something before we're ready.

HarelM:
Looks good. You should be able to test bump pr action and after it is merged a new version should be created in main (dry run)

ebrelsford:
@HarelM Great! Should I leave `--dry-run` on for now as-is for testing?

HarelM:
I would also recommend creating a RELEASE.md file describing the steps to carve out a new release as this might be done in large time intervals, and one might forget how to do it.

HarelM:
Note that while dry-run is on tags will still be created. 
I would recommend removing the dry run and simply create a pre-release as the first version.

.github/workflows/release.yml
```diff
@@ -0,0 +1,144 @@
...
+
+      - name: Use Node.js from nvmrc
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: 'js/.nvmrc'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Get version
+        id: package-version
+        working-directory: js
```
HarelM, line 57: Are you sure about this?

HarelM: Nevermind, I see that other steps are using it.

ebrelsford: I could be wrong but since all the JS including the `package.json` is in `js/` I thought this would be necessary.

PR #198, opened 2024-06-25
Fix release workflow

ebrelsford:
This PR:
 * Uses specific parameters where possible for `package.json`
 * Remove the `working-directory` parameter where possible

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -32,6 +32,7 @@
     "jest-matcher-deep-close-to": "^3.0.2",
     "nyc": "^17.0.0",
     "prettier": "^3.3.2",
+    "semver": "^7.6.2",
     "ts-jest": "^29.1.5",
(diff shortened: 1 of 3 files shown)


js/package.json
```diff
@@ -32,6 +32,7 @@
     "jest-matcher-deep-close-to": "^3.0.2",
     "nyc": "^17.0.0",
     "prettier": "^3.3.2",
+    "semver": "^7.6.2",
```
springmeyer, line 35: Was this an intended add?

springmeyer (approved):

js/package.json
ebrelsford: Yes it's used in the workflow.

HarelM:
Please note that a version was not released yet. 
I would recommend trying to create another bump version pr to see if the actions are behaving as expected.

PR #212, opened 2024-06-27
Add mlt-vector-tile-js

ebrelsford:
This PR adds a way to wrap MLT tiles in a similar API to [mapbox/vector-tile-js](https://github.com/mapbox/vector-tile-js/), which makes it possible to use MLTs in MapLibre GL JS. Eventually we would expect this to be a separate package but since this is experimental and the MLT API is changing we think it's reasonable to incubate this here for now.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -44,6 +44,7 @@
     "@types/varint": "^6.0.3",
     "bitset": "^5.1.1",
     "bytebuffer": "^5.0.1",
+    "geojson": "^0.5.0",
     "ieee754": "^1.2.1",
     "pbf": "^3.2.1",
     "varint": "^6.0.0"
--- a/js/src/index.ts
+++ b/js/src/index.ts
@@ -1,2 +1,4 @@
+import * as MltVectorTileJs from './mlt-vector-tile-js/index';
+export { MltVectorTileJs };
 export { MltDecoder } from './decoder/MltDecoder';
 export { TileSetMetadata } from './metadata/mlt_tileset_metadata_pb';
--- a/js/src/mlt-vector-tile-js/VectorTileFeature.ts
+++ b/js/src/mlt-vector-tile-js/VectorTileFeature.ts
@@ -1,0 +1,42 @@
+import { Feature } from 'geojson';
+import Point = require("@mapbox/point-geometry");
+
+class VectorTileFeature {
+  properties: { [key: string]: any } = {};
+  extent: number;
+  type: 0|1|2|3 = 0;
+  id: number;
+
+  _raw: any;
+  _geometry: any;
+  _keys: string[];
+  _values: any[];
+
+  constructor(feature) {
+    this.properties = feature.properties;
+    this.extent = feature.extent;
+    this._geometry = feature.geometry;
+    this._raw = feature;
+    if (feature.id !== null) {
+      this.id = Number(feature.id);
+    }
+  }
+
+  loadGeometry(): Point[][] {
+    const newGeometry = [];
+    const oldGeometry = this._raw.loadGeometry();
+    for (let i = 0; i < oldGeometry.length; i++) {
+      newGeometry[i] = [];
+      for (let j = 0; j < oldGeometry[i].length; j++) {
+        newGeometry[i][j] = new Point(oldGeometry[i][j].x, oldGeometry[i][j].y);
+      }
+    }
+    return newGeometry;
+  }
+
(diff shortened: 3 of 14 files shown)


js/package.json
```diff
@@ -45,6 +45,7 @@
     "@bufbuild/protobuf": "^1.10.0",
     "@types/bytebuffer": "^5.0.49",
     "bitset": "^5.1.1",
-    "bytebuffer": "^5.0.1"
+    "bytebuffer": "^5.0.1",
+    "geojson": "^0.5.0"
```
springmeyer: Can we do without this package? It looks like it is only being used for types but does not actually provide types and is old: https://www.npmjs.com/package/geojson

js/src/index.ts
```diff
@@ -1,2 +1,4 @@
+import * as MltVectorTileJs from './mlt-vector-tile-js/index';
+export { MltVectorTileJs };
```
springmeyer: If possible it would be nice to have this package separate so we could test them standalone. I think that would be feasible simply by not integrating into the root index.ts.

js/src/mlt-vector-tile-js/VectorTileLayer.ts
```diff
@@ -0,0 +1,25 @@
...
+  name: string | null;
+  extent: number;
+  length: number = 0;
+
+  _raw: any;
+  _keys: string[];
+  _values: any[];
+  _features: VectorTileFeature[] = [];
+
+  constructor(layer) {
```
springmeyer: Would be nice to have `layer` typed here.

js/src/mlt-vector-tile-js/VectorTileLayer.ts
```diff
@@ -0,0 +1,25 @@
...
+    this.name = layer.name;
+    this._features = layer.features.map((feature) => new VectorTileFeature(feature));
```
springmeyer: I recommend avoiding this `map` and rather lazily creating a `new VectorTileFeature` below in the `feature(i)` call like vector-tile-js: https://github.com/mapbox/vector-tile-js/blob/77851380b63b07fd0af3d5a3f144cc86fb39fdd1/lib/vectortilelayer.js#L60

js/src/index.ts
ebrelsford: Yeah I went back and forth about this, I can rearrange it.

js/src/mlt-vector-tile-js/VectorTileLayer.ts
ebrelsford: Done!

ebrelsford:
@springmeyer I think I got to everything here, appreciate you taking a look at it!

springmeyer:
Okay thanks. Will have a chance to take a look in about 1.5 hrs.

js/src/mlt-vector-tile-js/VectorTileFeature.ts
```diff
@@ -0,0 +1,41 @@
...
+class VectorTileFeature {
+  properties: { [key: string]: any } = {};
+  extent: number;
+  type: 0|1|2|3 = 0;
+  id: number;
+
+  _raw: any;
+  _geometry: any;
+  _keys: string[];
+  _values: any[];
```
springmeyer: It looks like _geometry, _keys, and _values are unused and can be removed?

js/src/mlt-vector-tile-js/VectorTileFeature.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  constructor(feature) {
+    this.properties = feature.properties;
+    this.extent = feature.extent;
+    this._geometry = feature.geometry;
```
springmeyer: _geometry can be removed?

js/src/mlt-vector-tile-js/VectorTileFeature.ts
```diff
@@ -0,0 +1,41 @@
...
+    this._raw = feature;
+    if (feature.id !== null) {
+      this.id = Number(feature.id);
+    }
+  }
+
+  loadGeometry(): Point[][] {
```
springmeyer, line 20:
```suggestion
  loadGeometry(): Point[][] {
  // TODO: optimize to avoid needing this deep copy 
```

springmeyer (approved):
Looking good. Just caught a few other things that look unneeded.

Issue #190, opened 2024-06-24
Layer order is different in MLT output than input MVT


springmeyer:
The layer order is unmatched between MLT output and the input MVT.

Unless there is a good reason for them to be different we should probably ensure they match.

For example in `bing/4-8-5`:

MVT Order:
```
  vector_background
  autorail
  country_region
  island
  land_cover_forest
  land_cover_grass
  populated_place
  reserve
  road
  water_feature
  water_pattern_area
```

MLT Order:

```
  water_feature
  road
  island
  land_cover_grass
  reserve
  water_pattern_area
  country_region
  land_cover_forest
  vector_background
  populated_place
  autorail
```

/cc @mactrem @ebrelsford

PR #220, opened 2024-06-28
[WIP] Improve performance and add additional benchmarks for the Java decoder

mactrem:


springmeyer:
@mactrem great to see this coming together! Huge change! I won't have a change to fully review before the weekend but will look forward to taking a closer look next week. ~~One thing now however: I recall asking you about the output format of `jmh` before. I do apologize, I seem to have forgot as I'm having trouble understanding the table above. What is faster? What is smaller? What is "score"?~~ EDIT, never mind, found docs at https://medium.com/@AlexanderObregon/introduction-to-java-microbenchmarking-with-jmh-java-microbenchmark-harness-55af74b2fd38.

~~Maybe we should work to consolidate on a single reporting format between Java and JS benchmarks: I went with `ops/sec`  which is documented at https://github.com/maplibre/maplibre-tile-spec/blob/main/js/bench/readme.md#benchmark-design, but I'm flexible. I also went with console output in JS but it would be easy to create an option to dump a table. Let's figure out how to be consistent across implementations.~~ EDIT: looks like it will be easy to match formats, will just need a table view for the JS side of things...

java/src/jmh/java/com/mlt/BingMapsDecoderBenchmark.java
```diff
@@ -57,13 +59,67 @@ private void encodeTile(int z, int x, int y) throws IOException {
...
         tileMetadata,
         TestSettings.BING_MVT_PATH,
         SEPARATOR);
   }
 
   @Benchmark
+  public FeatureTable[] decodeMltZ4() {
+    var mlTile = encodedMltTiles.get(4);
+    var mltMetadata = tileMetadata.get(4);
+    return MltDecoder.decodeMlTileVectorized(mlTile, mltMetadata);
```
springmeyer, line 73: @mactrem if I understand correctly `decodeMlTileVectorized` returns a `FeatureTable[]` but does not actually return features or decode geometries until the iterator.next() is iterated through on every FeatureTable. If just getting the FeatureTable is what you intend to benchmark (rather than feature + geometry decoding), then the closest equivalent to that for MVT would be just doing:

> VectorTile vectorTile = new VectorTile(pbf, pbf.length);

And nothing else. But I see that below the `org.springmeyer.VectorTile` library is being used to iterate through and decode all features.

java/src/main/java/com/mlt/converter/mvt/MvtUtils.java
```diff
@@ -50,6 +61,23 @@ public static JtsMvt decodeMvt2Fast(ByteArrayInputStream mvtTile) throws IOExcep
...
+   * */
+  public static Map<String, VectorTileLayer> decodeMvtMapbox(byte[] mvtTile) throws IOException {
+    Pbf pbf = new Pbf(mvtTile);
+    VectorTile vectorTile = new VectorTile(pbf, pbf.length);
+    for (var layer : vectorTile.layers.values()) {
+      for (int i = 0; i < layer.length; i++) {
+        var feature = layer.feature(i);
+        var geometry = feature.loadGeometry();
+      }
+    }
```
springmeyer, line 68: Two thoughts on this:

1. Parity with MLT decoding in Java

See my comment above that I think this is doing significantly more decoding work than the MLT benchmark which currently appears to be calling `MltDecoder.decodeMlTileVectorized` without actually iterating the features/geometries.

3. Parity with JS

This is roughly equivalent to what I'm benchmarking against in JS. See here: https://github.com/maplibre/maplibre-tile-spec/blob/96852ba4577e849dfb29a3f14f3da7cf83d3f760/js/bench/utils.ts#L68-L88. The differences are:

A. I'm collecting the length of the top level geometry array as a cheap way to ensure that dead code elimination does not strike. I gather than jmh uses black hole techniques to avoid the JVM eliminating this code so that should not be needed? But maybe the java benchmarks should collect a count of something and return it just to be safe?
In JS I've also profiled to ensure that all these functions are being called as I would expect.

B. I'm sorting the layers first to workaround https://github.com/maplibre/maplibre-tile-spec/issues/190. I don't have evidence that parsing layers in different orders will result in different performance, but I could see it resulting in different memory allocations, so I think making sure the order is the same is a good idea to rule out any potential for variability due to that. Could you add sorting before the test?

java/src/main/java/com/mlt/converter/mvt/MvtUtils.java
```diff
@@ -50,6 +61,23 @@ public static JtsMvt decodeMvt2Fast(ByteArrayInputStream mvtTile) throws IOExcep
...
+
+    return vectorTile.layers;
```
springmeyer, line 70: I propose iterating all features in both the MVT and MLT benchmarks and returning a count of some kind.

springmeyer (changes requested):
Had a chance to quickly do a first pass review. Requesting some reworking of the code being benchmarked to be more comparable between MLT and MVT decoding (see inline comments).

Also:

 - Would be great to pull in org.springmeyer.vector-tile-java via a package rather than vendoring: I think gradle supports pulling direct from GitHub right? If not, I can package it to maven. Let me know your preference
 - I would really like to see, before merging, separation of the benchmark code and optimizations in different PRs. So we could have the benchmark land first and then see the benchmark results change in a followup PR to optimize the Java code. Does that sound good as the eventual path forward once you have time to work on this again?

java/src/jmh/java/com/mlt/BingMapsDecoderBenchmark.java
mactrem: The `FeatureTable` is the in-memory MLT format which i'm proposing as part of the [spec](https://github.com/maplibre/maplibre-tile-spec/tree/main/spec#2-in-memory-format). So basically the spec is divided into the storage and in-memory format like it is the case for example with Parquet and Arrow. This MLT in-memory format can be directly processed like it is the case with the MVT in-memory representation but in a more efficient way. For example you can directly filter the data, process the geometries or in some cases directly copy the geometries into GL buffers. So the geometries are fully decoded and usable by (next next generation) map renderer. By not calling loadGeometry the (most) important and dominant part of the decoding is left out (such as transforming the command/parameter integer into vertices, also delta, zigZag, varint decoding). Even if the data are lazy decoded, in my experience (nearly) all geometries get materialized at some point in time before rendering (at least in data/styles  i'm using) and you pay the costs anyway. Or am I missing something in relation to the lazy decoding or do you have a different experience? 

But of course I also understand the problem that we cannot currently use this column-oriented in-memory format directly in MapLibre, as this would mean too much refactoring effort and we therefore have to convert to another row-oriented in-memory representation. What do think if we you have two benchmarks, one which shows the potential of the format by decoding in the proposed in-memory format which is representative for next generation map renderer (or also for map renderer which using different in-memory representations) and benchmarks for the currently used MapLibre GL JS in-memory representation. 
In general because of my research background i'm in particular interested in the "next generation" use case, so i'm also totally fine if we don't merge this benchmarks. On the other hand i also think just showing the second benchmarks where we have this additional transformation into a different in-memory representation doesn't fully show the potential of the format. What do you think about it?

java/src/main/java/com/mlt/converter/mvt/MvtUtils.java
mactrem: But that's also the point of the proposed MLT storage and in-memory format and why they are so tightly coupled and may have better decoding performance. Because you have to do less work on decoding and you can also work on compressed data

mactrem:
> * Would be great to pull in org.springmeyer.vector-tile-java via a package rather than vendoring: I think gradle supports pulling direct from GitHub right? If not, I can package it to maven. Let me know your preference

yes, it is currently still work in progress. So my plan was to access it from maven. Do you think you could publish it to maven as it could be also interesting for other developers, as it shows clear performance advantage over the existing MVT libraries?

springmeyer:
Okay sounds good. I'll publish to maven and ping you back here when it is available.

java/src/jmh/java/com/mlt/BingMapsDecoderBenchmark.java
mactrem: Some additional comments on the proposed in-memory format. The in-memory format allows map rendering libraries on the CPU as well on the GPU random (constant time) access to every attribute in the tile. The step of transforming to a random access representation is also currently quite expensive and has to be optimized.  So the MLT in-memory format has basically the same capabilities like the Arrow or Velox format but tailored for the map rendering use case. The GeoArrow format for example, which is basically a spatial extension to Arrow,  can also be directly processed and rendered by Deck.GL see https://github.com/geoarrow/deck.gl-layers

mactrem:
@springmeyer @nyurik i will now continue working on the integration of MLT in MapLibre GL JS. Can we merge this PR, so that i can build on top of that?

louwers:
@mactrem Does it make sense to merge this? I can help you fix the merge conflict.

mactrem:
> @mactrem Does it make sense to merge this? I can help you fix the merge conflict.

@louwers this should be already included in the repo based on a different PR

louwers:
OK good because that does not look like a fun merge conflict to solve.

Closing!

PR #240, opened 2024-07-10
add @mapbox/point-geometry to dependencies

### LINKED ISSUE #239: Missing JS dependency
I was reading up on the project of this project to a writeup of my own, how is `@mapbox/point-geometry` a dependency, shows up in your lock file, but not a dependency? I'm assuming this is just a mistake or are you planning on removing it?

https://github.com/search?q=repo%3Amaplibre%2Fmaplibre-tile-spec%20mapbox%2Fpoint-geometry&type=code

springmeyer:
Currently `@mapbox/point-geometry` is a runtime/production dependency of the JS decoder in order to be compatible with [maplibre-gl-js](https://github.com/maplibre/maplibre-gl-js/blob/c75919f19d39ab7a7fdb7e30a00212e2ce0ba5b1/package.json#L22) which depends on it for the vertex storage format.

But we were not explicitly depending upon it. This only worked because other dependencies were pulling it in transitively (either maplibre-gl-js or @mapbox/vector-tile).

Fixes #239

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -10,6 +10,7 @@
       "license": "(MIT OR Apache-2.0)",
       "dependencies": {
         "@bufbuild/protobuf": "^1.10.0",
+        "@mapbox/point-geometry": "^0.1.0",
         "@types/bytebuffer": "^5.0.49",
         "bitset": "^5.1.1",
         "bytebuffer": "^5.0.1"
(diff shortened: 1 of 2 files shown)


wipfli (commented):
I think you should only list direct dependencies in your package.json.

springmeyer:
I'm not actively working on this project, so I won't have bandwidth to work on fixes like this, closing.

PR #251, opened 2024-07-20
Pure Java FSST port

### LINKED ISSUE #150: Address portability issue around FSST encoding support
## Context

Currently [FSST encoding](https://github.com/cwida/fsst) is used in this project.

The FSST implementation we depend on in the java code is a JNI module (which compiles a C++ module and loads it at runtime): https://github.com/maplibre/maplibre-tile-spec/tree/main/java/resources.

Up until very recently were having to manually compile and link locally which was brittle. So this was recently improved to run automatically as part of the java compile: #62 

## Problem

Limitations remain:

  - While theres is working and tested support for macOS (x86 and arm64), linux amd64, and windows x86_64, there are likely other platforms and architectures that are not supported (where the compile or linking may be broken)
  - The loading of the `FsstWrapper.so` is currently relative to the working directory. This means that the CLI tools and the java unit + benchmarks only work if run from within the `java` directory

So this needs to be improved. 

## Ideas for solutions

Potential solutions to consider:
 - Moving to a pure Java FSST implementation
 - Changing the `FsstWrapper.so` loading logic to be relative to the java file that loads it (or making other packaging fixes to make it portable)
 - Your idea here?

msbarry:
I ported https://github.com/cwida/fsst to Java so we can eventually eliminate the native build/JNI dependency. For now you can choose which to run with by setting `FSST=java` `FSST=jni` environmental variable, or `FSST=debug` to run both and analyze performance and compressed size difference and print a summary when the JVM exits. It still defaults to the JNI implementation for now.

It doesn't product byte-for-byte equivalent output, but in general this version produces smaller output sizes. When I run with `FSST=debug` over all of the tiles in the `test` directory, the Java version produces smaller output 98.7% of the time, and the average output is of the Java version is 5% smaller than JNI.

I added jmh benchmarks for small (30 byte), medium (600 bytes), large (23kb), and extra large (230kb) inputs:

```
Benchmark                            Mode  Cnt     Score     Error  Units
FsstBenchmark.encodeExtraLargeJava  thrpt    5   197.412 ±   1.429  ops/s
FsstBenchmark.encodeExtraLargeJni   thrpt    5   485.555 ±   4.583  ops/s
FsstBenchmark.encodeLargeJava       thrpt    5   321.528 ±   2.831  ops/s
FsstBenchmark.encodeLargeJni        thrpt    5   734.146 ±   9.440  ops/s
FsstBenchmark.encodeMediumJava      thrpt    5  3684.809 ± 214.504  ops/s
FsstBenchmark.encodeMediumJni       thrpt    5  5592.609 ± 508.552  ops/s
FsstBenchmark.encodeSmallJava       thrpt    5  8796.862 ±  75.670  ops/s
FsstBenchmark.encodeSmallJni        thrpt    5  7805.358 ± 634.229  ops/s
```

I wouldn't expect the Java version to be as fast especially for large inputs version since the C++ version has a lot of SIMD optimizations. 40%+ the throughput seems good for now until the java vector API lands and the Java version could switch over to use that.

I'd propose we let this burn-in for a bit in this repo, then eventually extract it to a maplibre/fsst-java repo for others to use as well.

Fixes #150


java/src/jmh/java/com/mlt/converter/encodings/fsst/FsstBenchmark.java
```diff
@@ -0,0 +1,81 @@
...
+@Threads(value = 1)
+@Warmup(iterations = 5)
+@Measurement(iterations = 5)
+@Fork(value = 1)
+public class FsstBenchmark {
+
+  // ~30 bytes
+  private static final byte[] SMALL =
+      Base64.getDecoder().decode("SGV1YmFjaExpcHBlS2V0dGJhY2hTdGV2ZXI=");
+  // ~600 bytes
```
msbarry, line 23: I extracted these from the FSST encoder when it was encoding some example tiles in the `test` directory. Let me know if you want me to put them somewhere else.

java/src/main/java/com/mlt/converter/encodings/fsst/ByteArrayList.java
```diff
@@ -0,0 +1,58 @@
+package com.mlt.converter.encodings.fsst;
+
+import java.util.Arrays;
+
+/**
+ * Lightweight arraylist of primitive bytes as an alaternative to {@link
+ * java.io.ByteArrayOutputStream} since it has synchronization overhead.
+ */
+class ByteArrayList {
```
msbarry: This has a noticeable performance improvement over ByteArrayOutputStream when creating byte arrays. I normally use HPPC for this, would you be open to adding a depenedency on that?

java/src/main/java/com/mlt/converter/encodings/fsst/SymbolTable.java
```diff
@@ -1,3 +1,47 @@
...
+        + symbols.length
+        + "], symbolLengths=int["
+        + symbolLengths.length
+        + "], compressedData=byte["
+        + compressedData.length
+        + "]}";
+  }
+
+  public int weight() {
+    return symbols.length + symbolLengths.length + compressedData.length;
```
msbarry, line 31: I assume when computing the total compressed size that symbol lengths each use 1 byte (not 4) - is this correct?

java/src/main/java/com/mlt/converter/encodings/fsst/SymbolTable.java
```diff
@@ -1,3 +1,47 @@
...
+            && Arrays.equals(compressedData, that.compressedData));
+  }
+
+  @Override
+  public int hashCode() {
+    int result = Arrays.hashCode(symbols);
+    result = 31 * result + Arrays.hashCode(symbolLengths);
+    result = 31 * result + Arrays.hashCode(compressedData);
+    return result;
+  }
```
msbarry, line 49: Intellij was complaining that default equals and hashcode aren't correct when raw arrays are fields in a record so I generated them.

java/src/main/java/com/mlt/converter/encodings/fsst/Fsst.java
```diff
@@ -0,0 +1,46 @@
...
+      // handling
+      int symbolIndex = compressedData[i] & 0xFF;
+      // 255 is our escape byte -> take the next symbol as it is
+      if (symbolIndex == 255) {
+        decodedData.write(compressedData[++i] & 0xFF);
+      } else if (symbolIndex < symbolLengths.length) {
+        int symbolLength = symbolLengths[symbolIndex];
+        int symbolOffset = symbolOffsets[symbolIndex];
+        byte[] symbolBytes = new byte[symbolLength];
+        for (int j = 0; j < symbolLength; j++) {
```
mactrem: I think there is no need for using a ByteBuffer class in that case for the symbolBuffer, so we could get rid of that extra loop by directly writing to the ByteArrayOutputStream with an specified offset or am I missing something?

java/src/main/java/com/mlt/converter/encodings/fsst/ByteArrayList.java
mactrem: Yes, we can introduce HPPC

java/src/main/java/com/mlt/converter/encodings/fsst/Fsst.java
mactrem: In general in MLT the size of the decoded buffer is also part of the tile, so that we can preallocate the buffer. Therefore, we could get rid of the dynamic allocation overhead based on the  ByteArrayOutputStream

mactrem (commented):
Thanks, great work! Just small comments, then we can merge it

java/src/main/java/com/mlt/converter/encodings/fsst/SymbolTable.java
mactrem: yes that is correct

java/src/main/java/com/mlt/converter/encodings/fsst/Fsst.java
msbarry: Good catch, I copied decoder over without any changes but it's a lot cleaner now with that fix.  I added a benchmark for that as well, looks like it's about 4x faster (decompresses 400-500MB/s vs. ~100-150MB/s with old version on my M1 macbook)

I added a new `decode` method that takes an expected length, but I couldn't figure out how to get that expected length in the callers so I just left the old one too and marked it as deprecated so we can update to use the new one in a followup PR?

mactrem: Yes, we can change it in the next PR

mactrem (approved):

mactrem:
@msbarry Just a small note. In the [BtrBlocks paper](https://www.cs.cit.tum.de/fileadmin/w00cfj/dis/papers/btrblocks.pdf)
they achieve about 15 GB/s throughput in the fsst decoding of Strings. The next time i find time I will have look at their implementation. 

Bildschirmfoto 2024-07-26 um 21 13 04

mactrem:
At first glance, the difference of 30x is difficult to explain on such simple decoding between the java and c++ implementation. For sure their implementation is ahead of time compiled because of C++ and they probably preallocating the array and for sure some micro optimizations are also possible, but lets see ...

msbarry:
I got the 400MB/s number from the benchmark that decodes ~140kb of compressed data into 230kb of decompressed data at about 2000 operations per second - so it's reading 280MB/s and writing 400-500MB/s.

If you remove writing to the output array and just scan the input for escape symbols, it's around ~1.5GB/s - looks like most of the slowdown from there is copying symbols from the symbol byte array to the output (including looking up the symbols offset and length).

I'd be curious to see an apples to apples comparison against equivalent C++ code to see how much is java/native difference vs. if using a different algorithm?

PR #252, opened 2024-07-23
Use import instead of require on @mapbox/point-geometry

### LINKED ISSUE #239: Missing JS dependency

wipfli:
Uses an ES6 import for `@mapbox/point-geometry` and uses the latest version of the package which is also ES6 I think.

Closes #239

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -10,6 +10,7 @@
       "license": "(MIT OR Apache-2.0)",
       "dependencies": {
         "@bufbuild/protobuf": "^1.10.0",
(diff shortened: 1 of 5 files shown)


HarelM:
This is no longer relevant. Closing.

PR #259, opened 2024-07-24
Feature/fast_pfor_ts

Gh0stD3x:
Fixed issues with the VarInt encoding/decoding where an exception would speed up the process


mactrem:
@Gh0stD3x can you please provide some benchmark results in the PR which compares the current decoding performance of FastPfor with Varint?

tom-nom-nom:
Benchmarks were already included in the README.md file but here they are aswell

benchmarks

springmeyer:
@Gh0stD3x Could you contribute this to the `js/` folder instead of `encodings` folder? After #45 the fastpfor TS implementation is here: https://github.com/maplibre/maplibre-tile-spec/tree/main/js/src/encodings/fastpfor and has seen fixes since your original contribution: https://github.com/maplibre/maplibre-tile-spec/pull/133. So, what I recommend doing is re-creating a PR against main that modifies files, as needed, in the `js/src/encodings` folder. Sound good?

PR #383, opened 2024-11-13
Add optional pre triangulation for polygons when creating MLT 

FF-98:
## Summary
- add option to triangulate polygons using earcut when converting from MVT to MLT

Adjustments in:
- MltConverter
- GeometryEncoder
- VectorizedGeometryDecoder


java/build.gradle
```diff
@@ -27,11 +27,13 @@ dependencies {
     implementation 'org.apache.orc:orc-core:1.8.1'
     implementation 'com.github.davidmoten:hilbert-curve:0.2.3'
     implementation 'com.carrotsearch:hppc:0.10.0'
+    implementation "io.github.earcut4j:earcut4j:2.2.2"
```
msbarry, line 30: We already depend on JTS, which has a few built-in polygon triangulation utilities: https://github.com/locationtech/jts/tree/master/modules/core/src/main/java/org/locationtech/jts/triangulate/polygon - and I think `PolygonEarClipper` is similar to earcut.  Do you have any quality/performance comparison of those vs. earcut4j? If earcut4j performs better then adding a new dependency seems fine, but if not I'd default to the dependencies we already have.

java/src/main/java/com/mlt/converter/MltConverter.java
```diff
@@ -266,7 +272,8 @@ private static byte[] encodePropertyColumns(
           FeatureTableOptimizations featureTableOptimizations,
           List<Feature> sortedFeatures,
           List<Feature> mvtFeatures,
-          PhysicalLevelTechnique physicalLevelTechnique) {
+          PhysicalLevelTechnique physicalLevelTechnique,
+          Boolean triangulatePolygons) {
```
msbarry, line 276: use a primitive `boolean` here

java/src/main/java/com/mlt/converter/encodings/GeometryEncoder.java
```diff
@@ -33,12 +34,15 @@ private GeometryEncoder() {}
   public static EncodedGeometryColumn encodeGeometryColumn(
       List<Geometry> geometries,
       PhysicalLevelTechnique physicalLevelTechnique,
-      SortSettings sortSettings) {
+      SortSettings sortSettings,
+      boolean triangulatePolygons) {
```
msbarry, line 38: Could we make it possible this take the triangulated polygons as an input to this function instead of embedding the triangulation inside of it?

On planetiler I'm considering some options for fixing invalid polygons that produce triangulated geometries as an intermediate step and I'd like for it to at least be possible to pass those triangulated polygons into this routine without forcing them to be computed again.

java/src/main/java/com/mlt/converter/triangulation/TriangulatedPolygon.java
```diff
@@ -0,0 +1,23 @@
+package com.mlt.converter.triangulation;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TriangulatedPolygon {
```
msbarry, line 6: If I'm reading build.gradle correctly, it looks like we're compiling against java 17. If that's right could this class just be `record TriangulatedPolygon(List<Integer> indexBuffer, int numTrianglesPerPolygon) {}` ?

java/src/main/java/com/mlt/converter/triangulation/VectorTileConverter.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  private void triangulatePolygons(List<VectorTileDecoder.Feature> decodedTile) {
+    for (VectorTileDecoder.Feature feature : decodedTile) {
+      var geometry = feature.getGeometry().toString();
+      if (geometry.contains(Geometry.TYPENAME_MULTIPOLYGON.toUpperCase())) {
+        triangulateMultiPolygon((MultiPolygon) feature.getGeometry());
+      } else if (geometry.contains(Geometry.TYPENAME_POLYGON.toUpperCase())) {
+        triangulatePolygon((Polygon) feature.getGeometry());
+      }
```
msbarry, line 89: Can we change these places that switch based on type of geometry to:

```suggestion
      if (feature.getGeometry() instanceof MultiPolygon multiPolygon) {
        triangulateMultiPolygon(multiPolygon);
      } else if (feature.getGeometry() instanceof Polygon polygon) {
        triangulateMultiPolygon(polygon);
      }
```

java/src/main/java/com/mlt/converter/triangulation/TriangulationUtils.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+import org.apache.commons.lang3.ArrayUtils;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.MultiPolygon;
+import org.locationtech.jts.geom.Polygon;
+
+public class TriangulationUtils {
+  private TriangulationUtils() {}
+
+  public static TriangulatedPolygon triangulatePolygon(Polygon polygon) {
```
msbarry, line 14: A `Polygon` can have holes, and a `MultiPolygon` just contains many `Polygon`s (which may each have holes). Can we change this to handle those cases and use the `poly.getExteriorRing()` `poly.getNumInteriorRing()` `poly.getInteriorRingN(int n)` helpers? Let's add some tests for polygons with holes as well.

java/src/main/java/com/mlt/converter/triangulation/TriangulationUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    var holeIndex = 0;
+
+    ArrayList<Double> multiPolygonCoordinates = new ArrayList<>();
+    ArrayList<Integer> holeIndices = new ArrayList<>();
+
+    for (int i = 0; i < multiPolygon.getNumGeometries(); i++) {
+      // assertion: first polygon defines the outer linear ring and the other polygons define its
+      // holes!
+      if (i == 0) {
+        holeIndex = multiPolygon.getGeometryN(i).getCoordinates().length;
```
msbarry, line 35: Each call to getCoordinates may be transforming a packed double[] CoordinateSequence to a `Coordinate[]` array. Is it possible to just extract the `CoordinateSequence` once and use `double getX(int idx)`, `double getY(int idx)` and `double getZ(int idx)` (only when coordinateSequence.getDimension() > 2)

java/src/test/java/com/mlt/earcut/EarCutTriangulationTest.java
```diff
@@ -0,0 +1,158 @@
...
+    innerPolygonCoordinates[0] = new Coordinate(0.25, 0.25);
+    innerPolygonCoordinates[1] = new Coordinate(0.75, 0.25);
+    innerPolygonCoordinates[2] = new Coordinate(0.75, 0.75);
+    innerPolygonCoordinates[3] = new Coordinate(0.25, 0.75);
+
+    when(innerPolygon.getCoordinates()).thenReturn(innerPolygonCoordinates);
+
+    when(multiPolygon.getNumGeometries()).thenReturn(2);
+    when(multiPolygon.getGeometryN(0)).thenReturn(outerPolygon);
+    when(multiPolygon.getGeometryN(1)).thenReturn(innerPolygon);
```
msbarry, line 68: I'd suggest just creating the actual MultiPolygon/Polygon classes instead of using mocks.  One easy way is to use `new WKTReader().read("POLYGON ((30 10, 40 40, 20 40, 10 20, 30 10))")` to decode the well-known-text representation of a geometry

HarelM:
@TimSylvester @louwers please check if this PR was already addressed and if we want to allow this option as part of the CLI.
The current conflict list indicate this will probably won't be possible to just merge this PR.
Closing this one.
@FF-98 feel free to open a new one against the new code.

PR #440, opened 2025-01-27
Update fastpfor

Gh0stD3x:
My latest changes to the FastPFOR decoder. Please use `FastPFORDecoder` now as the main interface now.


CommanderStorm:
main has diverged a lot and has a lot of changes done to the fpf code. sorry that this did not get a review

PR #441, opened 2025-02-06
First cut at native C++ MLT loader

TimSylvester:
Implements all the necessary types and encodings to load the tiles in `test/expected`.

Currently builds only with CMake.

Depends on protozero (for tileset metadata decoding), FastPFoR, and (optionally) nlohmann/json.

When dumped to GeoJSON, results match the output of the nodejs test, after being modified slightly to produce that output.


cpp/include/mlt/metadata/stream.hpp
```diff
@@ -0,0 +1,246 @@
...
+#include <mlt/util/buffer_stream.hpp>
+#include <mlt/util/noncopyable.hpp>
+#include <mlt/util/varint.hpp>
+
+#include <optional>
+#include <memory>
+
+namespace mlt::metadata::stream {
+
+enum class DictionaryType {
```
louwers: Maybe you want to use something like magic_enum? https://github.com/Neargye/magic_enum

cpp/.pre-commit-config.yaml
```diff
@@ -0,0 +1,24 @@
+repos:
```
louwers: This file needs to be in the repository root I think.

cpp/src/mlt/decode/geometry.hpp
```diff
@@ -0,0 +1,430 @@
...
+
+namespace mlt::decoder {
+
+class GeometryDecoder {
+    GeometryFactory geometryFactory;
+
+public:
+    using GeometryColumn = geometry::GeometryColumn;
+
+    GeometryColumn decodeGeometryColumn(BufferStream& tileData,
```
louwers: Doesn't really feel like this function need to be a method. Consider making it a free standing function?

Also applies to the other methods of this class.

cpp/src/mlt/decode/geometry.hpp
```diff
@@ -0,0 +1,430 @@
...
+        intDecoder.decodeIntStream<std::uint32_t, std::uint32_t, GeometryType>(
+            tileData, geomColumn.geometryTypes, *geomTypeMetadata);
+
+        for (std::uint32_t i = 1; i < numStreams; ++i) {
+            const auto geomStreamMetadata = StreamMetadata::decode(tileData);
+            switch (geomStreamMetadata->getPhysicalStreamType()) {
+                case PhysicalStreamType::LENGTH: {
+                    if (!geomStreamMetadata->getLogicalStreamType() ||
+                        !geomStreamMetadata->getLogicalStreamType()->getLengthType()) {
+                        throw std::runtime_error("Length stream missing logical type: " + column.name);
```
louwers, line 100: Instead of throwing exceptions you could consider using `std::expected`. Makes it a bit easier to test as well as to handle specific errors.

cpp/src/mlt/decode/int_template.hpp
```diff
@@ -0,0 +1,226 @@
...
+            }
+            break;
+        case LogicalLevelTechnique::COMPONENTWISE_DELTA:
+            if constexpr (std::is_same_v<TTarget, std::int32_t> || std::is_same_v<TTarget, std::uint32_t>) {
+                out.resize(values.size());
+                std::ranges::transform(values, out.begin(), [](auto x) { return static_cast<TTarget>(x); });
+                vectorized::decodeComponentwiseDeltaVec2(out);
+                break;
+            }
+            throw std::runtime_error("Logical level technique COMPONENTWISE_DELTA not implemented for 64-bit values");
```
louwers, line 86: I think this can be a `static_assert` instead.

cpp/src/mlt/metadata/stream.cpp
```diff
@@ -0,0 +1,81 @@
...
+    const auto logicalLevelTechnique1 = static_cast<LogicalLevelTechnique>(encodingsHeader >> 5);
+    const auto logicalLevelTechnique2 = static_cast<LogicalLevelTechnique>((encodingsHeader >> 2) & 0x7);
+    const auto physicalLevelTechnique = static_cast<PhysicalLevelTechnique>(encodingsHeader & 0x3);
+
+    using namespace util::decoding;
+    const auto [numValues, byteLength] = decodeVarints<std::uint32_t, 2>(buffer);
+
+    return {
+        physicalStreamType,
+        std::move(logicalStreamType),
```
louwers, line 98: No need for `std::move` due to RVO.

cpp/src/mlt/decode/int.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+template <typename T>
+std::vector<T>& IntegerDecoder::getTempBuffer() noexcept {
+    // Can we make these covariant and eliminate the casts?
+    if constexpr (std::is_trivially_assignable_v<decltype(impl->buffer4)::reference, T>) {
+        return reinterpret_cast<std::vector<T>&>(impl->buffer4);
+    } else if constexpr (std::is_trivially_assignable_v<decltype(impl->buffer8)::reference, T>) {
+        return static_cast<std::vector<T>&>(impl->buffer8);
+    }
+    // Ideally this would be a `static_assert`, but that fails even if never reached by template expansion.
```
louwers: You could assign the if constexpr values to a constexpr bool and then do `static_assert(!a && !b)` right?

cpp/src/mlt/decode/property.hpp
```diff
@@ -0,0 +1,191 @@
...
+#include <variant>
+#include <vector>
+#include "mlt/properties.hpp"
+#include "mlt/util/packed_bitset.hpp"
+
+namespace mlt::decoder {
+
+class PropertyDecoder {
+public:
+    PropertyDecoder(IntegerDecoder& intDecoder_, StringDecoder& stringDecoder_) noexcept(false)
```
louwers: Could you remove all the `noexcept(false)`? That's the default I think.

cpp/src/mlt/util/rle.hpp
```diff
@@ -0,0 +1,189 @@
...
+        while (position < numValues) {
+            // if we are out of values, read more
+            if (remainingValues == 0) {
+                readHeader();
+            }
+            // how many do we read out of this block?
+            size_t count = std::min(static_cast<size_t>(numValues - position), remainingValues);
+            uint64_t consumed = 0;
+            if (repeating) {
+#if SUPPORT_SKIP_NULL
```
louwers: Is there a good reason to make this configurable?

js/test/unit/decoder/MltDecoder.spec.ts
```diff
@@ -205,6 +205,22 @@ describe("MltDecoder", () => {
 
 });
 
+import { MapLibreTile } from "../../../src/data/MapLibreTile";
```
louwers: Should be at the top of the file.

cpp/src/mlt/decode/int_template.hpp
```diff
@@ -0,0 +1,226 @@
...
+#include <mlt/decode/int.hpp>
+
+#include <cassert>
+#include <stdexcept>
+#include <string>
+#include <utility>
+
+namespace mlt::decoder {
+
+#pragma region Public
```
louwers, line 15: What's this?

cpp/test/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+option(MLT_WITH_JSON "Include JSON support" ON)
+if(MLT_WITH_JSON)
+    target_include_directories(mlt-cpp-test PUBLIC "${json_SOURCE_DIR}/include")
+    target_compile_definitions(mlt-cpp-test PUBLIC MLT_WITH_JSON=1)
+endif(MLT_WITH_JSON)
+
+#target_include_directories(mlt-cpp-test PRIVATE "${PROJECT_SOURCE_DIR}/../include")
+
+target_compile_definitions(mlt-cpp-test PRIVATE PROTOZERO_USE_VIEW=std::string_view)
```
louwers: Maybe this should be set either in the build config or in the source code, but not both?

cpp/include/mlt/metadata/stream.hpp
TimSylvester: Why? We don't have need for the string equivalents so far, so I'm not seeing what value that would add to justify brining in more stuff.

cpp/src/mlt/decode/int_template.hpp
TimSylvester: That condition is only required for that branch in the switch, while the `static_assert` fails for all invocations.  Perhaps if I moved the switch case to a separate method, but the problem will go away when that functionality is implemented.

cpp/src/mlt/metadata/stream.cpp
TimSylvester: Only if it's returned alone, RVO doesn't apply to an initializer list passed to a constructor of the object being returned.

image

cpp/src/mlt/decode/int.cpp
TimSylvester: Indeed it does, clever.  And at least clang 14 is smart enough to not complain about a missing return value.

cpp/src/mlt/decode/property.hpp
TimSylvester: I've been using them to keep track of what I have and haven't evaluated, and I find it a helpful reminder to consider exception safety, but if you think it's too verbose, that's fair.

cpp/src/mlt/util/rle.hpp
TimSylvester: I couldn't tell if it was actually used while porting it over.  I'm 90% sure it isn't now that I've finished most of the decoder branches, so yeah I'll remove it.

cpp/src/mlt/decode/int_template.hpp
TimSylvester: Folding/navigation

image

cpp/src/mlt/decode/geometry.hpp
TimSylvester: The class instance has two members that would have to be passed around to each.  Still an option, but there are already more arguments than I'd like.
Which reminds me, I didn't finish out the factory pattern from the Java implementation, though I don't know if it's going to be useful.

cpp/src/mlt/decode/geometry.hpp
TimSylvester: That looks to be C++23, but I suppose we could patch it in...
Ideally I'd prefer avoiding exceptions altogether, but with STL that's impractical, not to mention dependencies.  I'm open to `expected` or similar, depending on how much extra if/else noise it ends up needing.

louwers (changes requested):
Please add some build/test instructions to cpp/README.md

cpp/.clang-format
```diff
@@ -0,0 +1,23 @@
+---
```
louwers: This file can be deleted, I added it to the root of the repo.

louwers:
> @louwers Should I move `.clang-tidy` to the root as well?

You can leave it in `cpp` for now, there is not so much C++ code for Java.

cpp/src/mlt/decode/int.cpp
```diff
@@ -0,0 +1,51 @@
...
+IntegerDecoder::~IntegerDecoder() noexcept = default;
+
+template <typename T>
+std::vector<T>& IntegerDecoder::getTempBuffer() noexcept {
+    // Can we make these covariant and eliminate the casts?
+    constexpr auto is4 = std::is_trivially_assignable_v<typename decltype(impl->buffer4)::reference, T>;
+    constexpr auto is8 = std::is_trivially_assignable_v<typename decltype(impl->buffer8)::reference, T>;
+    static_assert(is4 || is8);
+    if constexpr (is4) {
+        return reinterpret_cast<std::vector<T>&>(impl->buffer4);
```
louwers: Can you cast a `std::vector<std::uint32_t>` to a `std::vector<std::int32_t>`?

TimSylvester: I should revisit that anyway...

TimSylvester: Intermediate results all use the contents of a `uint8_t` vector now.

cpp/include/mlt/metadata/stream.hpp
```diff
@@ -10,26 +10,28 @@
 
 namespace mlt::metadata::stream {
 
-enum class DictionaryType {
+enum class DictionaryType : std::uint32_t {
```
louwers, line 13: Nit: no need to specify values because 0, 1, 2, 3, etc. is the default.

louwers:
Just noticed that vendor/fsst contains a lot of test/benchmark data and even an mp4 video.

```
 38M	./paper/dbtext
 68K	./paper/results
172K	./paper/zstd
 38M	./paper
 57M
```

We should probably only copy the relevant files instead of adding a submodule.

TimSylvester:
@louwers Can we use `sparse-checkout` to exclude those?  I hate to create something we have to maintain.

TimSylvester:
That seems to work.  If the goal is to keep from using up space on the CI hosts, anyway. 

[186466d5](https://github.com/maplibre/maplibre-tile-spec/pull/441/commits/186466d5663367f6dfacdb503e9014d05fd08fb2)

sjg-wdw:
I see @louwers has a requested change.  Has that been done?

TimSylvester:
I think so, all the comment issues are resolved.

louwers:
I don't think we're building this on CI right now right?

TimSylvester:
Building and testing, but using CMake for easy integration with MLN.  I haven't taken the time to figure out how to use `just` like everything else is.

[https://github.com/maplibre/maplibre-tile-spec/actions/runs/17109683704/job/48527713249?pr=441#step:4:45](https://github.com/maplibre/maplibre-tile-spec/actions/runs/17109683704/job/48527713249?pr=441#step:4:45)

nyurik (approved):
lets merge this to help with moving forward - ok as it is not being used anywhere just yet

nyurik:
I got a crash, trying to follow readme:

```bash
$ pwd
<....>/maplibre-tile-spec/cpp

$ git submodule init && git submodule update
$ cmake -GNinja -Bbuild -S. && cmake --build build --target mlt-cpp-test mlt-cpp-json
$ build/tool/mlt-cpp-json ../test/expected/bing/4-12-6.mlt

terminate called after throwing an instance of 'std::runtime_error'
  what():  Unexpected end of buffer
Aborted (core dumped)
```

I tried the same steps as in the CI, and they all passed.

nyurik:
This list has been moved to #530

My review questions, with most of them likely resulting in smaller tiles and less complexity

- [ ] Current CPP implementation has two integers for each block: `metadataSize` and `tileDataSize`.  Neither seem to be needed because we agreed to have 1 metadata + 1 feature table for each layer.  Since both are in bytes, it should save at least 4 bytes per layer, probably more.
- [ ] Why is there unused `featureTableBodySize` value in each `FeatureTable` (layer)?
- [ ] Why is the `version` field present in the data block? Shouldn't it be the very first byte of the metadata block? Note that the version byte could also be used as an indicator of the layer type (raster, vector, ...) - so we don't have to support just a single value in this position from the start. Also, better call it `LayerType`. Extendability here would simply add more types to this enum, allowing us to grow the spec without the breaking changes.
- [ ] Why is extent info also part of the `FeatureTable`?  This would make sense to also store in the metadata.
- [ ] Why are there unused `maxTileExtent` and `numFeatures` in the `FeatureTable`?
- [ ] The  `featureTableId` does not seem to be needed either if we always have a single metadata+data pair.  If we eventually introduce tileset-wide shared metadata, we can use a different version value.
- [ ] The 4096 tile extent is so common, I would recommend dedicating a separate `LayerType` for that -- saves another 2 bytes per layer. This way we can have the most common vector type as one value, and the custom vector type would specify more params like extent, etc.
- [ ] Why do we use strings to identify column types, e.g. `id` and `geometry`? We would have to store it in each layer's metadata now, adding mandatory 10 bytes per layer?  Also, if the user defines a column with the same name, what happens?
  - [ ] if column type is predefined as an enum, e.g. `id = 1, geometry = 2, one property = 3`, the `numStreams` value should not be needed for the known types.
  - [ ] the geometry type (const) vs arbitrary geometry type should be two or more enum values

nyurik (changes requested):
see issues above

cpp/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+module(name = "maplibre-tile-spec")
```
louwers: Should be moved to the root I think.

cpp/src/mlt/geometry_vector.cpp
```diff
@@ -0,0 +1,545 @@
...
+#include <mlt/util/stl.hpp>
+
+namespace mlt::geometry {
+
+#if !defined(MLT_GEOMETRY_CHECKS)
+#define MLT_GEOMETRY_CHECKS 1
+#endif
+namespace {
+#if MLT_GEOMETRY_CHECKS
+void checkBuffer(std::size_t index, std::size_t size, std::string_view name) {
```
louwers: Suggestion: make this an inline function with if constexpr, then the macro is not needed.

cpp/include/mlt/metadata/tileset.hpp
```diff
@@ -0,0 +1,256 @@
...
+#include <variant>
+#include <vector>
+
+namespace mlt {
+struct BufferStream;
+}
+
+namespace mlt::metadata::tileset {
+
+// Based on maplibre-tile-spec:spec/schema/mlt_tileset_metadata.proto commit
```
louwers: Can you link to https://maplibre.org/maplibre-tile-spec/specification/ instead?

TimSylvester:
Most of these are the result of following the JS implementation and necessary to read the test files, so I'll let @mactrem speak to the design.

We discussed eliminating support for multiple `FeatureTables`, but then decided to merge as-is and apply changes from there, and I have [some concerns](https://github.com/maplibre/maplibre-tile-spec/pull/479#discussion_r2307659806) and want feedback.

> * [ ]  Why is there unused `featureTableBodySize` value in each `FeatureTable` (layer)?

IIRC, one goal is to be able to easily skip decoding of layers which are not used, so we need to be able to jump past it as soon as we get to the name.  This implementation doesn't yet do that, though.  If we do the above, then we can can drop this size.

cpp/MODULE.bazel
TimSylvester: My longer-term goal was to use `just` for the cpp build as well, and allow Bazel to build it from `vendor`, but I ended up needing to add this to satsfy Bazel.  Possibly I missed something and it's not really necessary?

TimSylvester:
> I got a crash, trying to follow readme:

I hadn't re-encoded all the test tiles in the new format, it should work now.  Anything still using the old format won't, of course.

nyurik:
I tried to document all the aspects of this code from the data structure perspective, outlined my concerns, and began some top-level specification refactoring in https://github.com/maplibre/maplibre-tile-spec/discussions/530

nyurik:
sigh, even the precommit CI is having issues updating your PR :)  (maybe add it as an allowed contributor?)  Or at least lets get the CI to pass, merge it, and will do another PR to fix all this

louwers (approved):
Let's merge it, what do you think @TimSylvester?

nyurik:
lets merge

PR #442, opened 2025-02-07
fix(ts): use the correct vertex count for multi-polygons with vertex offsets

TimSylvester:
I noticed this while building the c++ version.  It looks like a bug, but it isn't covered by any of the tests.

Compare to the [Java code](https://github.com/maplibre/maplibre-tile-spec/blob/d449bdb8e9c2e8451a5afaa61751bc523465a487/java/src/main/java/com/mlt/decoder/GeometryDecoder.java#L260-L269) for reference.


DoFabien:
Hi,
For awareness, the current TS implementation already uses ring-specific counts in the multipolygon path (`ts/src/vector/geometry/geometryVectorConverter.ts`), so the issue does not appear there. The logic has been in place since commit `99f2c56f4` (2025-04-02);

HarelM:
@nyurik @CommanderStorm @louwers should this PR be merged?
Is this case covered by the synthetic tests?

HarelM:
Oh, wait, I just saw that this change is in the typescript code and not Java.
@Turtelll is this the same as your recent change to multipypolygon fix?

Turtelll:
No I didn't change anything regarding multipypolygon. I just used  the topologyVector to extract the ring and part offsets there were no changes made to production code

HarelM:
Are you sure? I see changes in the production code as part of this PR with regards to counter increments, am I missing anything? 
https://github.com/maplibre/maplibre-tile-spec/pull/1018/changes

Turtelll:
These are regarding MULTIPOINT

HarelM:
Is it? the code here is under the  `else if (geometryType === GeometryType.MULTIPOLYGON)` statement...

HarelM:
Oh, you mean the other PR is about multi point.
Fair, but it is possible that there's still a bug in multipolygon as well and not just in multi point?

Turtelll:
No i don't think so. 
- geometryOffsetsCounter++ - advances for each MULTIPOLYGON geometry
- partOffsetCounter++ - advances for each polygon within the MULTIPOLYGON
- ringOffsetsCounter++ - advances for each ring

All the different configurations of the synthetic tests files confirm correct advancements

sjg-wdw:
Did this ever get resolved?

HarelM:
Not yet, on my todo list.

HarelM:
My bad, I thought this was referring to a different issue. 
I don't know if this fix is correct or not, I need a failing use case to check that. 
Was that added as a synthetic test?

TimSylvester:
I don't have a test case, I never even had that code running locally.  From looking at the encoder, it looks like vertex offsets are present only if it's smaller to use dictionary encoding on the vertices, which I guess means that the same vertices are reused significantly.  It looks like that branch isn't covered in the Java tests, either.

HarelM:
- This is no longer needed, replaced by tests in #1552

PR #479, opened 2025-04-29
first pass

TimSylvester:


spec/README.md
```diff
@@ -145,22 +147,21 @@ This had the advantage that encodings can be reused and implementation of encode
 |--------------|--------------------------------|--------------------------------------------|
 | Date         | Int32                          | number of days since Unix epoch            |
 | Timestamp    | Int64                          | number of milliseconds since Unix epoch    |
-| RangeMap     | RangeMap (Map<vec2<Double>, T> | for storing linear referencing information |
+| RangeMap     | RangeMap (Map<vec2<Double>, T>)| for storing linear referencing information |
```
nyurik: perhaps this for conciseness?   Also, note that C++ types might be a bit confusing for some. I, for example, not sure how `List<UInt8>` is different from `vec` - is there a type guide here?
 
```suggestion
| RangeMap     | Map<vec2<Double>, T> | for storing linear referencing information |
```

TimSylvester: `List` implies arbitrary length, `vec2` is always 2 items.

`vec2` is GLSL and Rust terminology, C++ would use `array<double,2>` (or `pair` or `tuple`).

I suppose familiarity with all these things is assumed, but it could be confusing.  There's a glossary at the end which needs to be filled out, and perhaps some of these should be added as well.

nyurik:
@mactrem any thoughts on this?  This is mostly a documentation change, and you are the foremost expert on this :)

spec/assets/FeatureTableSchema.mermaid
```diff
@@ -0,0 +1,143 @@
...
+    class ComplexField {
+      +ComplexType physicalType %% oneof i.e., physicalType XOR logicalType
+      +LogicalComplexType logicalType
+      +VarInt childCount %% Present only if CHILD_TYPES is set in columnOptions
+      +Field[] children
+    }
+
+    %% ---------------- String ------------------
+    class String {
+      +VarInt length
```
nyurik, line 87: weird - the the preview does not contain the `+` symbol

```suggestion
      VarInt length
```

spec/assets/FeatureTableSchema.mermaid
```diff
@@ -0,0 +1,143 @@
...
+      +ColumnOptions options %% VarInt
+      +String name
+      +ScalarColumn scalarType %% oneof i.e., scalarType XOR complexType
+      +ComplexColumn complexType
+    }
+
+    %% ---------------- ScalarColumn ----------------
+    class ScalarColumn {
+      +ScalarColumnOptions options %% VarInt
+
```
nyurik: Is the empty line needed here?

```suggestion
```

spec/README.md
```diff
@@ -69,50 +65,57 @@ This physical streams are further divided into logical streams that add addition
...
 
 ##### Tile Metadata
 
-In addition to the metadata that applies to the entire tileset,
-every tile has three additional types of sections with information about its structure.
+Each `FeatureTable` is preceded by a `FeatureTableMetadata` describing it.
+
+A tile consists of any number of:
+
+- The size of the `FeatureTableMetadata` to follow, varint-encoded
```
nyurik, line 81: Why do we need the size of the entire feature table metadata as well as the entire feature table data here?  Also, if you say below that `each pair may contain one or more layers.  This allows the tile to be built by simple concatenation of separate results.` -- how can they (layers?) be concatenated if you have a size field for the tables?  A bit confusing if layers and feature tables are the same concept here?

For context, the practice of prepending the size of the upcoming data is a common pattern, but it should only be used in a case where no other indicator exists.  For example, an integer value could indicate the size of the following string -- but we can assume the string is NOT `nil-terminated` -- otherwise we would have two, possibly conflicting indicators of the string end.  The same logic applies to concatenated layers (feature tables?) -- if we know the overall size of the payload (i.e. HTTP end of stream, or file size), we do not need to know how many layers are coming up - we can simply read them until the end, expecting the end of the final layer to coincide with the end of the stream/file.

spec/README.md
```diff
@@ -132,7 +135,7 @@ Complex types are composed of scalar types.
 | DataType         | Logical Types           | Description                                        | Layout        |
 |------------------|-------------------------|----------------------------------------------------|---------------|
 | List             | Binary (List<UInt8>)    |                                                    | Variable-Size |
-| Map              | RangeMap (Map<vec2d, T> | additional key stream -> length, key, data streams | Variable-Size |
+| Map              | Map<vec2d, T>           | additional key stream -> length, key, data streams | Variable-Size |
 | Struct           |                         |                                                    |               |
 | Vec2<T>, Vec3<T> | Geometry, GeometryZ     |                                                    | Fixed-Size    |
```
nyurik, line 140: tbh, i am a bit concerned about the Vec3 here. The Z-value (i.e. elevation) usually has vastly different scale from the x and y values.  So listing them as `Vec3<T>` which implies the same `T` type may result in a massive inefficiency that at the same time has very limited usage. On the other hand, handling Z value in the same way as just another the per-vertix `M` value might allow significant flexibility - that `M` value could be elevation (with its own scale), road width, range-based RLE-encoded attribute, etc.  Do note that Z value here is not a hard requirement to ship in v1 if there are any concerns with it, as it does not exist in MVT.  I might be totally misunderstanding it though.

spec/README.md
```diff
@@ -145,22 +148,21 @@ This had the advantage that encodings can be reused and implementation of encode
 |--------------|--------------------------------|--------------------------------------------|
 | Date         | Int32                          | number of days since Unix epoch            |
 | Timestamp    | Int64                          | number of milliseconds since Unix epoch    |
-| RangeMap     | RangeMap (Map<vec2<Double>, T> | for storing linear referencing information |
+| RangeMap     | Map<vec2<Double>, T>           | for storing linear referencing information |
```
nyurik, line 152: if this is the percentage-based range encoding we have been discussing, it is a bit confusing - are you saying it's a map of a two-value double -> T? If the double is the resulting value, it seem to be an overkill -- Overture ranges for example are percentage based, and depending on the needed precision, we may even want to have a per metadata (per column?) "scaling factor" for these values

mactrem: In the following, my thoughts on the support for linear referencing in MLT.

### Context  
Overture supports **linear referencing (geometric scoping)** in two ways:  
- Using the `at` property  
- Using the `between` property  

📖 See: [Geometric Scoping – Linear Referencing](https://docs.overturemaps.org/schema/concepts/scoping-rules/#geometric-scoping-linear-referencing)

---

### Proposal for v1  
For the **initial implementation**, I recommend focusing on the `between` property and encoding ranges using a **double-pair–based encoding**:  

- Introduce a **range stream** before the data stream, storing interleaved start and end values (as percentages).  
- These ranges define the applicability of the corresponding values in the data stream.  
- Compress the double values using **ALP encoding**.  

Applying a **scaling factor** (stored in metadata) to reduce values to integers would improve efficiency. However, it also introduces complexity, since the generic `RangeMap` structure would need explicit handling in metadata. So again thinking about a concept of how such a specilized handling fits into the metadata (time that currently probably nobady has!)

One limitation of that approach is that MultiPart geometries (MultiLineStrings) are not supported, so v1 would focus on single part geometries (LineString)
 
In my optinionm the goal for v1 should be  to introduce this novel linear referencing feature for vector tiles in a **simple, working form**, while leaving room for optimizations (informed by benchmarking) in future versions.


---

### Handling the `at` Property  
For v1, I  would not add specialized support for the `at` property. Instead, it can be represented in simpler ways:  
- Store the same value twice in the range stream.  
- Or encode it as an **m-value**.  

This avoids extra complexity in the first version while still making `at` usable and again optimizing it in v2. 

---

### General Considerations  
Full support in the rendering stack will take additional time anyway
- The **Style Spec** needs to be updated.  
- The **renderer** must also be adapted.

spec/assets/FeatureTableSchema.mermaid
TimSylvester: Some things were missing a `+`

spec/README.md
TimSylvester: There are two different ways to build the header.  The more straightforward way, and the way it has been so far, using a single header followed by the columns it describes is a little more efficient.  But at Markus' request, for the benefit of the Microsoft design, it also supports layers (or groups of layers) being built independently and concatenated without additional processing.

nyurik: @TimSylvester I am aware of that -- that was my request to make tile layers concatenatable.  I am still a bit confused by the design, possibly due to the naming change.  In MVT, the structure is just a sequence of layers, where each layer is independent. Here, it seems the structure is two levels, so may need a bit more explanation and/or an image

TimSylvester: Oh, sorry, I misremembered.  It wasn't my intent to rename anything, so that may be a mistake, though I'm not sure what you're referring to specifically.  The intent is that the data sent by a tile server could be multiple `Tile`s, but I could add another level to the diagram to make that more explicit.

Alternatively, we could simplify by always interleaving so that a tile is only a sequence of metadata+layer pairs, though I would need to update some things and re-encode the test data.

nyurik: I think the interleave is a clearer pattern, and yes, lets clarify the docs plus maybe a simple diagram to make sure we are on the same page. Thx!

TimSylvester: The existing code does a lot of work to combine things across layers and even tiles when building the metadata, and I'm hesitant to tear this apart to make it produce only one layer at a time.

https://github.com/WetDogWeather/maplibre-tile-spec/blob/5ad7288366577ca7dd7b542d4f787d03f2e67b21/java/src/main/java/com/mlt/converter/MltConverter.java#L35

@mactrem Do you have any guidance here?

louwers (approved):

louwers (commented):
This PR cleans things up a lot, but some parts don't read very spec-y yet.

> Note: Further evaluation is needed to determine if the [latest research findings](https://arxiv.org/pdf/2306.15374.pdf) can be used to enable random access on delta encoded values as well

Or the design goals. For example: "To be future-proof it is important that the file can be parallel processed on the GPU for the usage within a GPU compute shader."

This reads like an academic paper, not like a specification.

The current implementations in `ENCODINGS.md` also do not belong in a specification.

I think all non-normative writing should be moved to a separate documentation  1.0 is released.

louwers:
Once this is merge I will make a separate PR this week to do some copy editing.

nyurik:
merged, lets do the copy-editing :)

PR #486, opened 2025-05-26
refactor: get_data_type_from_column in decode_id_column

Weixing-Zhang:
This PR was originally intended to include the full implementation of decode_column_id, but I realized that StreamMetadataDecoder.decode does not currently produce the same result as the Java implementation when given the same input. Therefore, I am submitting this smaller PR first to isolate that issue and keep changes manageable.

Implemented:
- Separated scalar type extraction logic into a standalone helper function to simplify the main decoding flow.
- Added an unit test

### DIFF
--- a/rust/mlt/src/decoder/decode.rs
+++ b/rust/mlt/src/decoder/decode.rs
@@ -6,10 +6,10 @@ use geo_types::Geometry;
 use zigzag::ZigZag;
 
 use crate::data::MapLibreTile;
-use crate::decoder::helpers::decode_boolean_rle;
+use crate::decoder::helpers::{decode_boolean_rle, get_data_type_from_column};
 use crate::decoder::varint;
 use crate::encoder::geometry::GeometryScaling;
-use crate::metadata::proto_tileset::{column, scalar_column, Column, ScalarType, TileSetMetadata};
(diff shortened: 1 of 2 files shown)


nyurik:
did you push the right code?  I only see the removal of an empty line

Weixing-Zhang:
@nyurik This is just a placeholder. I am working on it now so others are aware.

rust/mlt/src/vector/vector_type.rs
```diff
@@ -0,0 +1,8 @@
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
nyurik: why do you need this file? it doesn't seem to be related to the rest of the code?

jjcfrancisco (approved):
Brilliant work @Weixing-Zhang

nyurik (approved):

rust/mlt/src/vector/vector_type.rs
Weixing-Zhang: I do not why you still can see this. It was included in the original PR but has been removed `outdated`.

nyurik: must be caching issue, all good now

PR #488, opened 2025-05-31
fix: remove unintended cursor advancement

Weixing-Zhang:
Cursor advancement was resulting in incorrect reads and has therefore been removed.


nyurik:
I'm ok to merge this type of changes at this stage, but we should try to do test driven development for this as the product matures, otherwise it is not very clear how this pr changes the behavior of the code, and which tests used to fall but don't fail any more

nyurik (approved):

PR #491, opened 2025-06-03
fix: StreamMetadata decoder unhandled types

Weixing-Zhang:
Context:
Previously, the following decoding path caused StreamMetadata::decode to be skipped entirely:
```
let property_column_names: Option<&String> = self.config.as_ref().and_then(|cfg| {
    cfg.feature_table_decoding
        .as_ref()
        .and_then(|ftd| ftd.get(&feature_table_metadata.name))
});

if property_column_names.is_none() {
    self.tile.advance(*feature_table_body_size as usize);
    continue;
}
```
As a result, the decoder was never exercised in practice. Upon investigation:
- The present physical stream type is currently unhandled.
- Decoded metadata values appear to be inconsistent or incorrect.

To-be-implemented:
- [x] Add a handler for the `Present` stream type
- [x] Change `LogicalStreamType` to be Option because None is not a variant in the enum, but it can be None.
- [x] Add missing None for `LogicalLevelTechnique` and `PhysicalLevelTechnique` to match the enum
- [x] Add offset tracking for debugging
- [x] Add unit tests

### DIFF
--- a/rust/mlt/src/decoder/decode.rs
+++ b/rust/mlt/src/decoder/decode.rs
@@ -25,15 +25,26 @@ pub struct Config {
 pub struct Decoder {
     pub tile: Bytes,
     pub config: Option<Config>,
+    pub original_tile_size: usize,
 }
 
 impl Decoder {
     pub fn new(tile: Vec<u8>, config: Option<Config>) -> Self {
+        let tile = Bytes::from(tile);
         Self {
-            tile: Bytes::from(tile),
+            original_tile_size: tile.len(),
+            tile,
             config,
         }
     }
+
+    // Returns the offset of the tile data for debugging purposes
+    // NOTE: This is a temporary workaround to track the offset,
+    // and should be removed once `Bytes` supports offset tracking internally.
+    pub fn offset(&self) -> usize {
+        self.original_tile_size - self.tile.len()
+    }
+
     #[expect(unused_variables)]
     pub fn decode(&mut self, tile_metadata: &TileSetMetadata) -> MltResult<MapLibreTile> {
         while self.tile.has_remaining() {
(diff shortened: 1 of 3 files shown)


rust/mlt/src/decoder/decode.rs
```diff
@@ -25,15 +25,25 @@ pub struct Config {
 pub struct Decoder {
     pub tile: Bytes,
     pub config: Option<Config>,
+    pub original_tile_length: usize,
 }
 
 impl Decoder {
     pub fn new(tile: Vec<u8>, config: Option<Config>) -> Self {
+        let bytes = Bytes::from(tile);
```
nyurik: It is a common practice in Rust to re-use variables. Also, if you change the order of fields, you don't need an extra var. The second change might be less readable so up to you.  Also, I think `size` is more concise than `length`

```rust
        let tile = Bytes::from(tile);
        Self {
            original_tile_length: tile.len();
            tile,
            config,
        }
```

Weixing-Zhang: Done. Thank you for the suggestions.

Weixing-Zhang:
@jjcfrancisco The major changes I made were adding the None variant to some of the enums because we are using the enum index to read the types. Additionally, I added the missing present type. 

Please take a look when you are available. I will implement the `Bytes` wrapper, and this PR is not blocking me.

jjcfrancisco:
> @jjcfrancisco The major changes I made were adding the None variant to some of the enums because we are using the enum index to read the types. Additionally, I added the missing present type.
> 
> Please take a look when you are available. I will implement the `Bytes` wrapper, and this PR is not blocking me.

Looks good 👍

Issue #495, opened 2025-06-07
Improve offset tracking in Rust implementation


Weixing-Zhang:
Context:
Our current implementation uses the `bytes` crate, which does not expose an internal offset, only the remaining length. Please correct me if I am wrong, but I could not find any support for this in the `bytes` documentation. This limitation makes it difficult to align or debug Rust against the java and ts implementations in the same repo.

I can think of the following 3 options:
- Create a lightweight wrapper around Bytes to track the offset manually
- Replace Bytes with Cursor<Vec<u8>> for native position tracking
- Upstream a patch or extension trait to the bytes crate (if maintainers are open)

Im inclined to the lightweight wrapper, as it should be straightforward to implement and flexible enough to retain the native functionality of the bytes crate while giving us the offset tracking we need.

nyurik:
@Weixing-Zhang could you clarify the use case?  Perhaps an example of the sudo-code you want to be able to do?

Weixing-Zhang:
@nyurik @jjcfrancisco 

The primary use case is during development, where we need to compare the behavior of our Rust implementation with others—especially in the decoding stage. 

For example, [java](https://github.com/maplibre/maplibre-tile-spec/blob/54fe468c0de2b9710957deb7746bdcd17f1e7698/java/src/main/java/com/mlt/decoder/MltDecoder.java#L63C13-L63C84), [ts](https://github.com/maplibre/maplibre-tile-spec/blob/54fe468c0de2b9710957deb7746bdcd17f1e7698/ts/src/mltDecoder.ts#L60C9-L60C28), and [js](https://github.com/maplibre/maplibre-tile-spec/blob/54fe468c0de2b9710957deb7746bdcd17f1e7698/js/src/decoder/MltDecoder.ts#L23) implementations all use an offset to track the current reading position. 

However, in our rust implementation, we don’t currently have this since we are using `bytes`. The `bytes` crate simplifies byte access and eliminates the need for manual index management, but as far as I know, it doesn't expose the offset position. **_Adding it directly to the MLT decoder doesn't solve the problem, as the offset is being consumed in various scopes throughout the code_.**

An example can be as simple as follows. However, it does add maintenance overhead like `slice` because we lose access to the full apis of `bytes` unless we re-expose or reimplement the needed methods ourselves.

```
use bytes::{Buf, Bytes};

pub struct TrackingBytes {
    inner: Bytes,
    original_size: usize,
}

impl TrackingBytes {
[...]

PR #496, opened 2025-06-09
feat: bytes wrapper for tile with offset

Weixing-Zhang:
Context:
- Reported issue at https://github.com/maplibre/maplibre-tile-spec/issues/495

Todo:
- [x] Add a bytes wrapper
- [x] Add unit test
- [x] Replace the use of bytes with the wrapper

### DIFF
--- a/rust/mlt/src/decoder/tracked_bytes.rs
+++ b/rust/mlt/src/decoder/tracked_bytes.rs
@@ -44,8 +44,7 @@ mod tests {
 
     #[test]
     fn test_tracked_bytes_offset_and_reads() {
-        let data = vec![0x10, 0x20, 0x30, 0x40];
-        let mut tile = TrackedBytes::new(data);
+        let mut tile: TrackedBytes = [0x10, 0x20, 0x30, 0x40].as_slice().into();
 
         // Initial offset should be 0
         assert_eq!(tile.offset(), 0);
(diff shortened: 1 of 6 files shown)


rust/mlt/src/decoder/tracked_bytes.rs
```diff
@@ -0,0 +1,66 @@
+use std::ops::{Deref, DerefMut};
+
+use bytes::{Buf, Bytes};
+
+pub struct TrackedBytes(usize, Bytes);
+
+impl TrackedBytes {
+    pub fn new(data: Vec<u8>) -> Self {
+        let bytes = Bytes::from(data);
```
nyurik: Bytes already has a bunch of `From` [trait implementations](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#impl-From%3C%26%5Bu8%5D%3E-for-Bytes), no need to hard-code one of them. This code would allow calling `new` with anything that can be converted into bytes.

```suggestion
    pub fn new<T: Into<Bytes>>(data: T) -> Self {
        let bytes = data.into();
```

Weixing-Zhang: Thank you so much for pointing this out. It is indeed more generic now.

nyurik: hehe, well... that's the whole idea of `generics` :rofl:

Weixing-Zhang:
@nyurik @jjcfrancisco Please take a look whenever you have time. 

I will be working on the `decode_boolean_rle` as it didn't return the expected results the last time I checked.

jjcfrancisco (approved):

jjcfrancisco:
Happy with the work :)

PR #499, opened 2025-06-12
feat: implement decode int stream

Weixing-Zhang:
Todo:
- [ ] Add missing steps before decoding column ids
- [ ] Implement  `decode_int_stream` function
- [ ] Add required unit tests

### DIFF
--- a/rust/mlt/src/decoder/decode.rs
+++ b/rust/mlt/src/decoder/decode.rs
@@ -10,7 +10,7 @@ use crate::decoder::helpers::{decode_boolean_rle, get_data_type_from_column};
 use crate::decoder::tracked_bytes::TrackedBytes;
 use crate::decoder::varint;
 use crate::encoder::geometry::GeometryScaling;
-use crate::metadata::proto_tileset::{Column, TileSetMetadata};
+use crate::metadata::proto_tileset::{Column, ScalarType, TileSetMetadata};
 use crate::metadata::stream::StreamMetadata;
 use crate::{MltError, MltResult};
 
@@ -63,16 +63,16 @@ impl Decoder {
                     ))
                 })?;
 
-            let property_column_names: Option<&String> = self.config.as_ref().and_then(|cfg| {
-                cfg.feature_table_decoding
-                    .as_ref()
-                    .and_then(|ftd| ftd.get(&feature_table_metadata.name))
-            });
+            // let property_column_names: Option<&String> = self.config.as_ref().and_then(|cfg| {
+            //     cfg.feature_table_decoding
+            //         .as_ref()
+            //         .and_then(|ftd| ftd.get(&feature_table_metadata.name))
+            // });
 
-            if property_column_names.is_none() {
-                self.tile.advance(*feature_table_body_size as usize);
-                continue;
-            }
+            // if property_column_names.is_none() {
+            //     self.tile.advance(*feature_table_body_size as usize);
+            //     continue;
+            // }
 
             let extent = infos
                 .get(2)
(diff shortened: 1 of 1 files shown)


rust/mlt/src/decoder/integer_decoder.rs
```diff
@@ -1,18 +1,29 @@
...
 use crate::metadata::stream::StreamMetadata;
+use crate::metadata::stream_encoding::PhysicalLevelTechnique;
 
 pub fn decode_int_stream(
     tile: &mut TrackedBytes,
     stream_metadata: &StreamMetadata,
     is_signed: bool,
 ) -> Vec<i32> {
-    todo!();
+    if stream_metadata.physical.technique == PhysicalLevelTechnique::FastPfor {
```
nyurik: would it make sense to have some sort of an iterator here?  I don't know the performance implications, but allocating a new `Vec` on every return might also be a significant cost.  No idea what's the best path though

rust/mlt/src/decoder/integer_decoder.rs
```diff
@@ -0,0 +1,92 @@
+use crate::decoder::tracked_bytes::TrackedBytes;
+use crate::decoder::varint;
+use crate::metadata::stream::StreamMetadata;
+use crate::metadata::stream_encoding::{
+    DictionaryType, LengthType, Logical, LogicalLevelTechnique, LogicalStreamType, OffsetType,
+    Physical, PhysicalLevelTechnique, PhysicalStreamType,
+};
+use fastpfor::cpp::Codec32 as _;
+use fastpfor::cpp::FastPFor128Codec;
```
nyurik: if this is only needed for a test, move it to the test mod

Weixing-Zhang: It is for decoding. The PR is still in development, but the main logic for fastpfor decoding and encoding is completed. I need to spend some time unifying the use of the different types across the functions. 

Please ignore the notification for now. I will ping you and Francis after the PR is ready.

Weixing-Zhang:
@nyurik Could you please review this PR when you have time? Thank you.

nyurik:
@jjcfrancisco any thoughts about this PR? if not, will merge shortly

jjcfrancisco:
No comments, impressive work! The only comment I have is from [this previous message](https://osmus.slack.com/archives/C05GNNSK3MK/p1747901956292339?thread_ts=1747675030.025629&cid=C05GNNSK3MK) from a while back:

> ...you could do the following methods within the integer stream decoder:
> * getVectorTypeIntStream
> * decodeConstIntStream
> * decodeIntStream
> 
> Once those parts are done, you could look into the id column decoder and I could get started with the geometry decoder. Then the property column decoder it's up for grabs.

Are we still aiming for this @Weixing-Zhang or have new ideas/thoughts/routes? Also @nyurik would like to do errors first?

jjcfrancisco (approved):

Weixing-Zhang:
> No comments, impressive work! The only comment I have is from [this previous message](https://osmus.slack.com/archives/C05GNNSK3MK/p1747901956292339?thread_ts=1747675030.025629&cid=C05GNNSK3MK) from a while back:
> 
> > ...you could do the following methods within the integer stream decoder:
> > 
> > * getVectorTypeIntStream
> > * decodeConstIntStream
> > * decodeIntStream
> > 
> > Once those parts are done, you could look into the id column decoder and I could get started with the geometry decoder. Then the property column decoder it's up for grabs.
> 
> Are we still aiming for this @Weixing-Zhang or have new ideas/thoughts/routes? Also @nyurik would like to do errors first?

That sounds good to me. As a follow-up, I will probably refactor the errors first, so there will be less refactoring work needed for this.

Weixing-Zhang:
> awesome work!
> 
> One thing we will need to do relatively soon (separate PR) is to refactor errors - it is a very bad practice to do memory allocation on error because 1) its expensive, and not all errors are shown to the user/logged - often the caller may want to handle it somehow, 2) callers cannot handle it well because they have to do string parsing.
> 
> So instead, we should start creating different errors that take the actual parts of the error message. See `thiserror` crate for detals.

Totally. Let me refactor them in the next PR so there will be less refactoring needed for this later on.

PR #505, opened 2025-08-04
chore(rust): improve build and warnings

nyurik:
There are still warnings, but at least they will be more visible.

In the future, we should make all this compilable without any warnings.


jjcfrancisco:
I can see plenty of warnings but I understand that's ok for now.

PR #508, opened 2025-08-20
refactor: address post-merge review comments for PR#506

Weixing-Zhang:
- [ ] Address post-merge comments from @nyurik


rust/mlt/src/metadata/stream.rs
```diff
@@ -38,42 +38,30 @@ impl StreamMetadata {
         // let stream_type = tile
         //     .get(offset.position() as usize)
         //     .ok_or(MltError::DecodeError("Failed to read...".into()))?;
+
+        #[inline]
+        fn enum_from_code<T: TryFrom<u8>>(code: u8, kind: &'static str) -> Result<T, MltError> {
```
nyurik: why not simply create several errors instead of this?  How many different `T` types do you expect?  If only a few, lets just have dedicated errors for each, and that will simplify everything

Weixing-Zhang: This is not an error wrapper but helper function to convert byte to enum for each physical or technical stream type.
```
PhysicalStreamType::try_from(stream_type >> 4).map_err(|_| {
                MltError::InvalidDiscriminant {
                    kind: "PhysicalStreamType",
                    code: stream_type >> 4,
                }
            })?;
```
 I will remove those error wrappers in this PR.

nyurik: I meant that this helper doesn't add much value, e.g. instead of `enum_from_code(subtype_code, "DictionaryType")` we can use `MltError::InvalidPhysicalStreamType(code)` -- much simpler IMO

Weixing-Zhang: done. thank you.

rust/mlt/src/decoder/vectorized/helpers.rs
```diff
@@ -14,7 +14,10 @@ pub fn decode_componentwise_delta_vec2s<T: ZigZag>(data: &[T::UInt]) -> Result<V
         });
     }
     if len % 2 != 0 {
-        return Err(MltError::invalid_value_multiple(2, len));
+        return Err(MltError::InvalidValueMultiple {
```
nyurik, line 17: I wonder if a more descriptive error is better here, e.g. `UnexpectedDeltaVecSize` or something else like that?  This way it gives better context of what's wrong with the data stream for future debugging - rather than "unexpected multiple" - which could apply to anything...  We don't have to fix it now -- totally up to you on this one.

rust/mlt/src/error.rs
```diff
@@ -124,98 +135,13 @@ pub enum MltError {
...
-    Logical,
-}
-
-#[derive(Debug)]
-pub enum TechniqueDiscriminant {
-    Physical(PhysicalLevelTechnique),
-    Logical(LogicalLevelTechnique),
-}
-
 //----------------- Varint failures -----------------
```
nyurik: lets not do these C-style separator comments -- if its a description of this enum, just use the `///` -- this way they will be compiled into documentation.

Weixing-Zhang: done

rust/mlt/src/decoder/vectorized/helpers.rs
Weixing-Zhang: done

PR #510, opened 2025-08-25
chore(rust): fix clippy lints

nyurik:


rust/mlt/src/converter/mvt.rs
```diff
@@ -29,7 +30,7 @@ pub fn decode_mvt(mvt_tile: &[u8]) -> MapVectorTile {
                     .keys
                     .iter()
                     .zip(layer.values.iter())
-                    .map(|(k, v)| (k.replace("_", ":"), Value::from(v.clone())))
+                    .map(|(k, v)| (k.replace('_', ":"), Value::from(v.clone())))
```
Weixing-Zhang, line 33: nit: why not k.replace('_', ':'). Instead, just change one of them.

Weixing-Zhang:
Just out of curiosity, did you run`cargo clippy --fix` to fix the lint issues?

rust/mlt/src/converter/mvt.rs
nyurik: @Weixing-Zhang see https://doc.rust-lang.org/std/primitive.str.html#method.replace

pattern can be one of multiple things, whereas the second param is only `&str`

nyurik:
no, mostly by hand.  Some rust lints do get fixed that way, but relatively a small percentage

jjcfrancisco (approved):

PR #513, opened 2025-08-29
feat(rust): rle & morton partial decode methods

jjcfrancisco:


HarelM:
@nyurik @CommanderStorm is this PR still relevant given all the changes you made to the rust code?

nyurik:
I'll have to look, maybe some tests?

CommanderStorm:
some of this work would be interesting, but our codebase has diverged too far for this to be semi-mergable.

I will keep for example the fast delta in mind, for future optimisations.
Would be an interesting optimisation if you want to run this by our new benchmarks 😉 
Lets close the PR for now

PR #517, opened 2025-09-04
refactor: reorganize decode vectorized code

Weixing-Zhang:
As I began working on the decode_const_int_stream function, I referred to the TypeScript implementation. Instead of moving the stream-related code into the vectorized folder, I placed it in integer_stream.rs, which I find more structured and better organized.


jjcfrancisco (approved):
Yes, makes sense.

PR #525, opened 2025-09-14
massive spec rewrite

nyurik:
using both human and AI, rewrote most of the spec to be more readable, and flagged some unclear and possibly incorrect moments.

### DIFF
--- a/mkdocs/docs/encodings.md
+++ b/mkdocs/docs/encodings.md
@@ -2,153 +2,138 @@
 
 ## Plain
 
-No compression technique is applied on the data.
-Depending on the data type, the data are stored in the following format:
+No compression is applied to the data.
+Depending on the data type, values are stored in the following formats:
 
-- Boolean: Least-significant bit ordering within bytes
-- Integer: Little-Endian byte order
-- Long: Little-Endian byte order
-- Float: IEEE754 floating-point numbers in Little-Endian byte order
-- Double: IEEE754 floating-point numbers in Little-Endian byte order
-- String: length and data streams
+- **Boolean**: Least-significant bit ordering within bytes
+- **Integer**: Little-Endian byte order
+- **Long**: Little-Endian byte order
+- **Float**: IEEE754 floating-point numbers in Little-Endian byte order
+- **Double**: IEEE754 floating-point numbers in Little-Endian byte order
+- **String**: Length and data streams
 
 ## Boolean-RLE
 
-This encoding is used for compressing boolean columns.
-Least-significant bit numbering (bit-endianness) is used as bit order.
-Have a look at the [ORC specification](https://orc.apache.org/specification/ORCv1/#boolean-run-length-encoding)
-for implementation details.
+This encoding compresses boolean columns using least-significant bit numbering (bit-endianness).
+Refer to the [ORC specification](https://orc.apache.org/specification/ORCv1/#boolean-run-length-encoding) for implementation details.
 
 ## Byte-RLE
 
-This encoding is used for compressing byte streams such as the `GeometryType` stream of the Geometry column.
-Have a look at the [ORC specification](https://orc.apache.org/specification/ORCv1/#byte-run-length-encoding)
-for implementation details.
+This encoding compresses byte streams, such as the `GeometryType` stream in Geometry columns.
+Refer to the [ORC specification](https://orc.apache.org/specification/ORCv1/#byte-run-length-encoding) for implementation details.
 
 ## Dictionary Encoding
 
-Dictionary encoding is used to compactly represent repeated values and can be applied to `String` and `Geometry` columns.
-In addition to the actual distinct values stored in the `dictionary` stream, a separate `data` stream for the indices
-into the dictionary is used.
+Dictionary encoding compactly represents repeated values and can be applied to `String` and `Geometry` columns.
+Distinct values are stored in a `dictionary` stream, while a separate `data` stream stores indices into the dictionary.
 
 ### String Dictionary Encoding
 
 The dictionary stream contains distinct UTF-8 encoded string values.
-The data stream contains the indices for the dictionary encoded as `UInt32`.
-A dictionary-encoded nullable string column contains of the following streams and order: `Present`, `Length`, `Dictionary`, `Data`.
+The data stream contains dictionary indices encoded as `UInt32`.
+A dictionary-encoded nullable string column consists of the following streams in order:
+`Present`, `Length`, `Dictionary`, `Data`
 
-All streams can be further compressed by recursively applying lightweight encodings:
+All streams can be further compressed using lightweight encodings:
 
-- Present Stream: Boolean-RLE
-- Length and Data Stream: see Integer encodings
-- Dictionary: see FSST Dictionary
+- **Present Stream**: Boolean-RLE
+- **Length and Data Stream**: See Integer encodings
+- **Dictionary**: See FSST Dictionary
 
 #### FSST Dictionary Encoding
 
-Dictionary encoding needs fully repeating strings to reduce size.
-However, the attributes of geospatial data often contains strings which share a common prefix but are not completely equal
-such as the localized country names.
-FSST replaces frequently occurring substrings while maintaining support for efficient scans and random look-ups.
-It is used in MLT to further compress the UTF-8 encoded dictionary values.
-A FSST Dictionary-encoded nullable string column contains of the following streams and order:
+Dictionary encoding requires fully repeating strings to reduce size. However, geospatial attributes often contain strings with common prefixes that are not identical (e.g., localized country names).
+FSST replaces frequently occurring substrings while supporting efficient scans and random lookups. It further compresses UTF-8 encoded dictionary values in MLT.
+An FSST dictionary-encoded nullable string column consists of the following streams in order:
 `Present`, `SymbolLength`, `SymbolTable`, `String Length`, `Dictionary` (Compressed Corpus)
-For implementation details see [the following paper](https://www.vldb.org/pvldb/vol13/p2649-boncz.pdf)
+For implementation details, refer to [this paper](https://www.vldb.org/pvldb/vol13/p2649-boncz.pdf).
 
-Available implementations:
+**Available implementations**:
 
-- C++: [https://github.com/cwida/fsst](https://github.com/cwida/fsst)
-- Java: Work in Progress
-- Js/WebAssembly decoder: Work in Progress (simple to implement)
+- **C++**: [https://github.com/cwida/fsst](https://github.com/cwida/fsst)
+- **Java**: Work in progress
+- **JS/WebAssembly decoder**: Work in progress (simple to implement)
 
 #### Shared Dictionary Encoding
 
-A Shared dictionary encoding can be used, to share a common dictionary between the columns.
-In addition, nested fields can use a shared dictionary encoding, to share a common dictionary between the fields.
-This can be applied for example on localized values of the name:* columns of an OSM dataset that can be identical across fields.
-If a shared dictionary encoding is used for nested fields, all fields that use the shared dictionary
-must be grouped one behind the other in the file and prefixed with the dictionary.
+Shared dictionary encoding allows multiple columns to share a common dictionary. Nested fields can also use a shared dictionary (e.g., for localized values like `name:*` columns in an OSM dataset).
+If used for nested fields, all fields sharing the dictionary must be grouped together in the file and prefixed with the dictionary.
 
-Shared dictionary-encoded nullable string columns consists of the following streams and order:
+A shared dictionary-encoded nullable string column consists of the following streams in order:
 `Length`, `Dictionary`, `Present1`, `Data1`, `Present2`, `Data2`
 
-Shared FSST Dictionary-encoded nullable string columns consists of the following streams and order:
+A shared FSST dictionary-encoded nullable string column consists of the following streams in order:
 `SymbolLength`, `SymbolTable`, `String Length`, `Dictionary` (Compressed Corpus), `Present1`, `Data1`, `Present2`, `Data2`
 
 ### Vertex Dictionary Encoding
 
-Uses an additional `VertexOffsets` stream for storing the indices to the coordinates of the vertices
-in the `VertexBuffer` stream. The vertices in the VertexBuffer are sorted on a Hilbert-Curve
-and delta-encoded in combination with a null suppression technique.
+Uses an additional `VertexOffsets` stream to store indices for vertex coordinates in the `VertexBuffer` stream.
+Vertices in the VertexBuffer are sorted using a Hilbert curve and delta-encoded with null suppression.
 
 #### Morton Vertex Dictionary Encoding
 
-The coordinates of the `VertexBuffer` are transformed to a 1D integer using the Morton Code.
-Then the data are sorted according to the Morton code and further compressed by applying
-an Integer compression technique.
+`VertexBuffer` coordinates are transformed into a 1D integer using Morton code.
+The data is sorted by Morton code and further compressed using integer compression techniques.
 
 ## Integer Encodings
 
-In general most data in MLT are stored in the form of arrays of integers.
-Using efficient and fast algorithms for compressing arrays of integers is therefore crucial.
+Most data in MLT is stored as integer arrays, making efficient integer compression crucial.
 
-### Logical Level Technique
+### Logical-Level Techniques
 
-Integers are encoded based on the two lightweight compression techniques delta and rle as well
-as a combination of both schemes called delta-rle.
+Integers are encoded using delta encoding, run-length encoding (RLE), or a combination of both (delta-RLE).
 
-#### Delta
+#### Delta Encoding
 
-Compute the differences between consecutive elements of a sequence e.g. x2 - x1, x3 - x2, ...
-Is always used in combination with a physical level technique to reduce the number of bits needed to store the
-delta values.
+Computes differences between consecutive elements (e.g., x₂ - x₁, x₃ - x₂, ...).
+Used with physical-level techniques to reduce the number of bits required for storing delta values.
 
-#### RLE
+#### Run-Length Encoding (RLE)
 
-See [https://en.wikipedia.org/wiki/Run-length_encoding](https://en.wikipedia.org/wiki/Run-length_encoding) for a basic explanation
-All runs and values are stored in two separate buffers.
-If the values are unsigned integers ZigZag encoding is applied to the values buffer.
-Both buffers are then further compressed using a null suppression technique.
+Refer to [Wikipedia](https://en.wikipedia.org/wiki/Run-length_encoding) for a basic explanation.
+Runs and values are stored in separate buffers. For unsigned integers, ZigZag encoding is applied to the values buffer. Both buffers are further compressed using null suppression.
 
-#### Combining delta and rle (Delta-RLE)
+#### Delta-RLE
 
-Delta-compress the values and then apply RLE encoding.
-This technique is efficient for ascending sequences like `id`s.
+Applies delta encoding followed by RLE. Efficient for ascending sequences like `id` fields.
 
-### Physical Level Technique (Null Suppression)
+### Physical-Level Techniques (Null Suppression)
 
-####  ZigZag encoding
+#### ZigZag Encoding
 
-For encoding unsigned integers, ZigZag encoding is used for both of the following null suppression techniques.
-[ZigZag](https://en.wikipedia.org/wiki/Variable-length_quantity#Zigzag_encoding) encoding uses the least significant bit for encoding the sign.
+Used for encoding unsigned integers in null suppression techniques.
+[ZigZag encoding](https://en.wikipedia.org/wiki/Variable-length_quantity#Zigzag_encoding) uses the least significant bit to represent the sign.
 
-#### Varint encoding
+#### Varint Encoding
 
-Byte-aligned null suppression technique try to compress an integer using a minimal number of bytes.
-For implementation details see [Protobuf](https://protobuf.dev/programming-guides/encoding/#varints).
+A byte-aligned null suppression technique that compresses integers using a minimal number of bytes.
+For implementation details, refer to [Protobuf](https://protobuf.dev/programming-guides/encoding/#varints).
 
 #### SIMD-FastPFOR
 
-Bit-aligned null suppression technique that compresses an integer using a minimal number of bits
-by using a patched approach to store exceptions (outliers) in a separate section to keep the overall bit width small.
+A bit-aligned null suppression technique that compresses integers using a minimal number of bits.
+Uses a patched approach to store exceptions (outliers) separately, keeping the overall bit width small.
 
-see [https://arxiv.org/pdf/1209.2137.pdf](https://arxiv.org/pdf/1209.2137.pdf) and [https://ayende.com/blog/199524-C/integer-compression-the-fastpfor-code](https://ayende.com/blog/199524-C/integer-compression-the-fastpfor-code)
+Refer to:
+- [https://arxiv.org/pdf/1209.2137.pdf](https://arxiv.org/pdf/1209.2137.pdf)
+- [https://ayende.com/blog/199524-C/integer-compression-the-fastpfor-code](https://ayende.com/blog/199524-C/integer-compression-the-fastpfor-code)
 
-Available implementations:
+**Available implementations**:
 
-- C++: [https://github.com/lemire/FastPFor](https://github.com/lemire/FastPFor)
-- Java: [https://github.com/lemire/JavaFastPFOR](https://github.com/lemire/JavaFastPFOR)
-- C#: [https://github.com/Genbox/CSharpFastPFOR](https://github.com/Genbox/CSharpFastPFOR)
-- Js/WebAssembly: Work in Progress (higher implementation complexity)
+- **C++**: [https://github.com/fast-pack/FastPFOR](https://github.com/fast-pack/FastPFOR)
+- **Java**: [https://github.com/fast-pack/JavaFastPFOR](https://github.com/fast-pack/JavaFastPFOR)
+- **C#**: [https://github.com/Genbox/CSharpFastPFOR](https://github.com/Genbox/CSharpFastPFOR)
+- **JS/WebAssembly**: Work in progress (higher implementation complexity)
 
-## Float encodings
+## Float Encodings
 
-### Adaptive Lossless floating-Point Compression (ALP)
+### Adaptive Lossless Floating-Point Compression (ALP)
 
-See [https://dl.acm.org/doi/pdf/10.1145/3626717](https://dl.acm.org/doi/pdf/10.1145/3626717) for a detailed explanation.
-ALP is a lossless floating-point compression technique that is used to compress the float values in MLT.
+A lossless floating-point compression technique used in MLT.
+For a detailed explanation, refer to [this paper](https://dl.acm.org/doi/pdf/10.1145/3626717).
 
-Available implementations:
+**Available implementations**:
 
-- C++: [https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp](https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp)
-- Java: Work in Progress
-- JS/WebAssembly: Work in Progress
+- **C++**: [https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp](https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp)
+- **Java**: Work in progress
+- **JS/WebAssembly**: Work in progress
--- a/mkdocs/docs/specification.md
+++ b/mkdocs/docs/specification.md
@@ -6,88 +6,79 @@
 
 ### 1.1 Basics
 
-MLT (MapLibre Tile) contains information about a specific geographic region - a tile. Each tile is a collection of `FeatureTables`, which are equivalent to Layers in the [MVT specification](https://github.com/mapbox/vector-tile-spec).
+An MLT (MapLibre Tile) contains information about a specific geographic region, known as a tile. Each tile is a collection of `FeatureTables`, which are equivalent to `Layers` in the [MVT specification](https://github.com/mapbox/vector-tile-spec).
 
-Each `FeatureTable` contains thematically grouped vector data, known as Features. Features in one FeatureTable share a common set of attribute columns (properties) and usually share the same geometry type (but not required to).
+A `FeatureTable` contains thematically grouped vector data, known as `Features`. Features within a single FeatureTable share a common set of attribute columns (properties) and typically share the same geometry type (though this is not strictly required).
 
-The visual appearance of a tile on a map is usually controlled by [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/), which specifies how the features in the FeatureTables are rendered.
+Each `FeatureTable` is preceded by a `FeatureTableMetadata` that describes FeatureTable's structure.
 
-Each feature must have a `geometry` column, an optional `id` column, and optional property columns. The geometry column type is based on the OGC's Simple Feature Access Model (SFA) without the support for GeometryCollection types.
+The visual appearance of a tile is usually defined by a [MapLibre Style](https://maplibre.org/maplibre-style-spec/), which specifies how features are rendered.
 
-The geometries are not restricted to one geometry type, but this is recommended for reasons of efficiency.
-As in MVT, the geometry coordinates are encoded as integers in a vector tile grid coordinates.
+Each feature must have a `geometry` column, an optional `id` column, and optional property columns. The geometry column type is based on the OGC's Simple Feature Access Model (SFA), excluding support for `GeometryCollection` types.
+
+While geometries are not restricted to a single type, using one type per table is recommended for efficiency. As in MVT, geometry coordinates are encoded as integers within vector tile grid coordinates.
 
 > [!NOTE]
-> "column", "field", and "property" are used interchangeably in this document.
+> The terms `column`, `field`, and `property` are used interchangeably in this document.
 
-### 1.2 File Layout
+### 1.2 Tile Layout
 
-A FeatureTable in the MLT specification is based on a tabular, column-oriented layout.
-It uses different lightweight compression schemes to efficiently encode the values of the columns.
-A FeatureTable consists of a optional `id` column, a mandatory `geometry` column and optional property columns.
-Because there is no tile header, `FeatureTable`s can be concatenated on the fly.
+A `FeatureTable` in the MLT specification uses a tabular, column-oriented layout. It employs various lightweight compression schemes to encode column values efficiently.
 
-A logical column is separated into several physical `streams` (sub-columns) inspired by the ORC file format which are stored next to each other. A stream is a sequence of values of a known length in a continuous chunk of memory, all having the same type. A stream has some additional metadata such as the size of the stream and the encoding type.
-For example, a nullable string property column can have a `present stream`, which indicates the presence of a value,
-a `length` stream, which describes the number of characters of every string, and a `data stream`, with the actual
-UTF-8 encoded string values.
+A FeatureTable consists of a mandatory `geometry` column, an optional `id` column, and optional property columns. The absence of a single header at the beginning of the tile allows `FeatureTable`s to be constructed independently, and even concatenated on the fly.
 
-MLT distinguishes between the following type of streams:
+A logical column is separated into several physical `streams` (sub-columns), inspired by the ORC file format. These streams are stored contiguously. A stream is a sequence of values of a known length in a continuous memory chunk, all sharing the same type. Streams include additional metadata, such as their size and encoding type.
 
-- **present**:
-  Is used for a efficient encoding of sparse columns since it indicates the presence of a value base on a bit flag.
-  This stream can be omitted if the column is not declared as nullable in the `FieldMetadata`.
-- **data**:
-  Stores the actual data of a column such as `boolean`, `int`, `float`, or `string` values of the feature properties.
-  Dictionary-encoded values, string values, or the coordinate values of a geometry column are stored in this type of stream.
-  In addition to the optional present stream, this is the only used stream type for fixed-sized data types such as `boolean`, `int`, and `float`.
-- **length**:
-  Specifies the number of elements for variable-sized data types such as strings or list.
-- **offset**:
-  Stores the offsets into a data stream when dictionary encoding is used such as for strings or vertices.
+For example, a nullable string property column might have:
+- A **`present` stream** (a bit flag indicating the presence of a value).
+- A **`length` stream** (describing the number of characters for each string).
+- A **`data` stream** (containing the actual UTF-8 encoded string values).
 
-This physical streams are further divided into logical streams that add additional semantics of how to interpret the data:
+MLT defines the following stream types:
+
+- **Present**: Enables efficient encoding of sparse columns by indicating value presence via a bit flag. This stream can be omitted if the column is not nullable (as declared in the `FieldMetadata`).
+- **Data**: Stores the actual column data (e.g., `boolean`, `int`, `float`, or `string` values for feature properties, dictionary-encoded values, or geometry coordinates). For fixed-size data types (`boolean`, `int`, `float`), this is the only required stream besides the optional `present` stream.
+- **Length**: Specifies the number of elements for variable-sized data types like strings or lists.
+- **Offset**: Stores offsets into a data stream when using dictionary encoding (e.g., for strings or vertices).
+
+These physical streams are further categorized into logical streams that define how to interpret the data:
 
 ![](assets/spec/LogicalStreams.png)
 
 #### 1.2.1 Metadata
 
-##### Tileset Metadata
+##### Tileset Metadata <span class="experimental"/>
 
-Global metadata for the tileset is stored separately, in [JSON format](assets/spec/mlt_tileset_metadata.json).
+> [!NOTE]
+> Tileset metadata was initially implemented as a size reduction experiment. This feature is not currently supported.
 
-The tileset metadata defines information for the full tileset is the equivalent to the TileJSON spec, which is commonly used in combination with MVT.
+Global metadata for the entire tileset is stored separately in a [JSON file](assets/spec/mlt_tileset_metadata.json).
 
-By specifying the information once per tileset, the definition of redundant (shared) metadata per tile will be avoided,
-which can take some relevant space of the overall size on small tiles.
+This tileset metadata provides information for the full tileset and is the equivalent of the TileJSON spec commonly used with MVT and other tile types. By defining this information once per tileset, we avoid redundant metadata in each tile, saving significant space, especially for small tiles.
 
 ##### Tile Metadata
 
-Each `FeatureTable` is preceded by a `FeatureTableMetadata` describing it.
+Each `FeatureTable` is preceded by a `FeatureTableMetadata` section describing it.
 
-A tile consists of any number of:
+> [!CAUTION]
+> This is not clear, and possibly incorrect. Why any number? Should the size of the upcoming metadata and table be part of that structure?
 
-- The size of the `FeatureTableMetadata` to follow, varint-encoded
-- The size of the `FeatureTable` to follow, varint-encoded
-- One `FeatureTableMetadata`
-- One `FeatureTable`
+A FeatureTable consists of any number of the following sequences:
+- The size of the upcoming `FeatureTableMetadata` (varint-encoded).
+- The size of the upcoming `FeatureTable` (varint-encoded).
+- One `FeatureTableMetadata` section.
+- One `FeatureTable` section.
 
-The `FeatureTableMetadata` is described in detail below.
-
-Each `FeatureTableMetadata`/`FeatureTable` pair may contain one or more layers.  This allows the tile to be built by simple concatenation of separate results.
+This structure allows a tile to be built by simply concatenating separate results. The `FeatureTableMetadata` is described in detail below.
 
 Within a `FeatureTable`, additional metadata describes the structure of each part:
 
-- **FieldMetadata**: contains information about the number of streams the field is divided and the vector type for efficient decoding into the in-memory format. Every Field (column) section is preceded by a `FieldMetadata` section.
-- **StreamMetdata**: contains information about the Stream such as the used encoding schemes
-  and the number of values. Every Stream section is preceded by a `StreamMetadata` section.
+- **FieldMetadata**: Contains information about a field (column), including the number of streams it comprises and its vector type for efficient decoding into the in-memory format. Every field section is preceded by a `FieldMetadata` section.
+- **StreamMetadata**: Contains information about a stream, such as the encoding scheme used and the number of values. Every stream section is preceded by a `StreamMetadata` section.
 
-Since for every `Field`, a `FieldMetadata` section has to be present, even for fields which are not included in one
-specific tile, no `id` is needed.
-The absence of an entire field in a tile can be identified by a zero value for the number of streams.
-All integers in the metadata sections are `Varint`-encoded if u32 is used or bit-packed in case of an u8 data type.
+Since every `Field` has a `FieldMetadata` section, even for fields absent in a specific tile, no `id` is needed. A field's absence is indicated by a zero value for its number of streams. All integers in metadata sections are `Varint`-encoded (for u32) or bit-packed (for u8).
 
-``` mermaid
+```mermaid
 ---
 title: FeatureTableSchema
 config:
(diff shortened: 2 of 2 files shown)


mkdocs/docs/encodings.md
```diff
@@ -2,153 +2,138 @@
...
+This encoding compresses byte streams, such as the `GeometryType` stream in Geometry columns.
+Refer to the [ORC specification](https://orc.apache.org/specification/ORCv1/#byte-run-length-encoding) for implementation details.
 
 ## Dictionary Encoding
 
-Dictionary encoding is used to compactly represent repeated values and can be applied to `String` and `Geometry` columns.
-In addition to the actual distinct values stored in the `dictionary` stream, a separate `data` stream for the indices
-into the dictionary is used.
+Dictionary encoding compactly represents repeated values and can be applied to `String` and `Geometry` columns.
+Distinct values are stored in a `dictionary` stream, while a separate `data` stream stores indices into the dictionary.
```
CommanderStorm, line 32: Maybe a connecting sentence?

```suggestion
Distinct values are stored in a `dictionary` stream, while a separate `data` stream stores indices into the dictionary.
We allow the following encodings of this concept:
```

mkdocs/docs/encodings.md
```diff
@@ -2,153 +2,138 @@
...
 
 The dictionary stream contains distinct UTF-8 encoded string values.
-The data stream contains the indices for the dictionary encoded as `UInt32`.
-A dictionary-encoded nullable string column contains of the following streams and order: `Present`, `Length`, `Dictionary`, `Data`.
+The data stream contains dictionary indices encoded as `UInt32`.
+A dictionary-encoded nullable string column consists of the following streams in order:
+`Present`, `Length`, `Dictionary`, `Data`
 
-All streams can be further compressed by recursively applying lightweight encodings:
+All streams can be further compressed using lightweight encodings:
```
CommanderStorm, line 49: Is "can" here correct?
What is the alternative encoding if this is not chosen?
Let's either make this a MUST or specify what is the alternative

mkdocs/docs/encodings.md
```diff
@@ -2,153 +2,138 @@
...
+- **JS/WebAssembly decoder**: Work in progress (simple to implement)
 
 #### Shared Dictionary Encoding
 
-A Shared dictionary encoding can be used, to share a common dictionary between the columns.
-In addition, nested fields can use a shared dictionary encoding, to share a common dictionary between the fields.
-This can be applied for example on localized values of the name:* columns of an OSM dataset that can be identical across fields.
-If a shared dictionary encoding is used for nested fields, all fields that use the shared dictionary
-must be grouped one behind the other in the file and prefixed with the dictionary.
+Shared dictionary encoding allows multiple columns to share a common dictionary. Nested fields can also use a shared dictionary (e.g., for localized values like `name:*` columns in an OSM dataset).
```
CommanderStorm:
```suggestion
Shared dictionary encoding allows multiple columns to share a common dictionary.
Nested fields can also use a shared dictionary (e.g., for localized values like `name:*` columns in an OSM dataset).
```

mkdocs/docs/encodings.md
```diff
@@ -2,153 +2,138 @@
...
 
-#### RLE
+#### Run-Length Encoding (RLE)
 
-See [https://en.wikipedia.org/wiki/Run-length_encoding](https://en.wikipedia.org/wiki/Run-length_encoding) for a basic explanation
-All runs and values are stored in two separate buffers.
-If the values are unsigned integers ZigZag encoding is applied to the values buffer.
-Both buffers are then further compressed using a null suppression technique.
+Refer to [Wikipedia](https://en.wikipedia.org/wiki/Run-length_encoding) for a basic explanation.
+Runs and values are stored in separate buffers. For unsigned integers, ZigZag encoding is applied to the values buffer. Both buffers are further compressed using null suppression.
```
CommanderStorm:
```suggestion
Runs and values are stored in separate buffers.
For unsigned integers, ZigZag encoding is applied to the values buffer.
Both buffers are further compressed using null suppression.
```

mkdocs/docs/encodings.md
```diff
@@ -2,153 +2,138 @@
...
 
-#### Combining delta and rle (Delta-RLE)
+#### Delta-RLE
 
-Delta-compress the values and then apply RLE encoding.
-This technique is efficient for ascending sequences like `id`s.
+Applies delta encoding followed by RLE. Efficient for ascending sequences like `id` fields.
```
CommanderStorm:
```suggestion
Applies delta encoding followed by RLE.
Efficient for ascending sequences like `id` fields.
```

mkdocs/docs/encodings.md
```diff
@@ -2,153 +2,138 @@
...
+Integers are encoded using delta encoding, run-length encoding (RLE), or a combination of both (delta-RLE).
 
-#### Delta
+#### Delta Encoding
 
-Compute the differences between consecutive elements of a sequence e.g. x2 - x1, x3 - x2, ...
-Is always used in combination with a physical level technique to reduce the number of bits needed to store the
-delta values.
+Computes differences between consecutive elements (e.g., x₂ - x₁, x₃ - x₂, ...).
+Used with physical-level techniques to reduce the number of bits required for storing delta values.
```
CommanderStorm: It is not clear that we refer to the techniques below.
When I was originally reading this, I thought this was a convoluted way of saying: "Values are bitpacked".

Linking those sections would likely help ^^

mkdocs/docs/encodings.md
```diff
@@ -2,153 +2,138 @@
...
+A lossless floating-point compression technique used in MLT.
+For a detailed explanation, refer to [this paper](https://dl.acm.org/doi/pdf/10.1145/3626717).
 
-Available implementations:
+**Available implementations**:
 
-- C++: [https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp](https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp)
-- Java: Work in Progress
-- JS/WebAssembly: Work in Progress
+- **C++**: [https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp](https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp)
```
CommanderStorm: We do this pattern a few times in the article.
I think we should do the following instead in those cases


```suggestion
- **C++**: <https://github.com/duckdb/duckdb/tree/16830ded9b4882b90f35c2d7e2d740547ae7d3fc/src/include/duckdb/storage/compression/alp>
```

CommanderStorm (commented):
Much nicer to read. I have left a few nits, but nothing actually wrong

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
 
 ### 1.1 Basics
 
-MLT (MapLibre Tile) contains information about a specific geographic region - a tile. Each tile is a collection of `FeatureTables`, which are equivalent to Layers in the [MVT specification](https://github.com/mapbox/vector-tile-spec).
+An MLT (MapLibre Tile) contains information about a specific geographic region, known as a tile. Each tile is a collection of `FeatureTables`, which are equivalent to `Layers` in the [MVT specification](https://github.com/mapbox/vector-tile-spec).
```
CommanderStorm:
```suggestion
An MLT (MapLibre Tile) contains information about a specific geographic region, known as a tile.
Each tile is a collection of `FeatureTables`, which are equivalent to `Layers` in the [MVT specification](https://github.com/mapbox/vector-tile-spec).
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
-Each `FeatureTable` contains thematically grouped vector data, known as Features. Features in one FeatureTable share a common set of attribute columns (properties) and usually share the same geometry type (but not required to).
+A `FeatureTable` contains thematically grouped vector data, known as `Features`. Features within a single FeatureTable share a common set of attribute columns (properties) and typically share the same geometry type (though this is not strictly required).
```
CommanderStorm:
```suggestion
A `FeatureTable` contains thematically grouped vector data, known as `Features`.
Features within a single FeatureTable share a common set of attribute columns (properties) and typically share the same geometry type (though this is not strictly required).
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
-The visual appearance of a tile on a map is usually controlled by [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/), which specifies how the features in the FeatureTables are rendered.
+Each `FeatureTable` is preceded by a `FeatureTableMetadata` that describes FeatureTable's structure.
```
CommanderStorm:
```suggestion
Each `FeatureTable` is preceded by a `FeatureTableMetadata` that describes `FeatureTable`'s structure.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
-Each feature must have a `geometry` column, an optional `id` column, and optional property columns. The geometry column type is based on the OGC's Simple Feature Access Model (SFA) without the support for GeometryCollection types.
+The visual appearance of a tile is usually defined by a [MapLibre Style](https://maplibre.org/maplibre-style-spec/), which specifies how features are rendered.
 
-The geometries are not restricted to one geometry type, but this is recommended for reasons of efficiency.
-As in MVT, the geometry coordinates are encoded as integers in a vector tile grid coordinates.
+Each feature must have a `geometry` column, an optional `id` column, and optional property columns. The geometry column type is based on the OGC's Simple Feature Access Model (SFA), excluding support for `GeometryCollection` types.
```
CommanderStorm: How about this way of writing this

```suggestion
Each feature must have
- a `geometry` column (type based on the OGC's Simple Feature Access Model (SFA), excluding support for `GeometryCollection` types)
- an optional `id` column
- optional property columns
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
+
+While geometries are not restricted to a single type, using one type per table is recommended for efficiency. As in MVT, geometry coordinates are encoded as integers within vector tile grid coordinates.
```
CommanderStorm:
```suggestion
While geometries are not restricted to a single type, using one type per table is recommended for efficiency.
As in MVT, geometry coordinates are encoded as integers within vector tile grid coordinates.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
 > [!NOTE]
-> "column", "field", and "property" are used interchangeably in this document.
+> The terms `column`, `field`, and `property` are used interchangeably in this document.
+
+### 1.2 Tile Layout
 
-### 1.2 File Layout
+A `FeatureTable` in the MLT specification uses a tabular, column-oriented layout. It employs various lightweight compression schemes to encode column values efficiently.
```
CommanderStorm:
```suggestion
A `FeatureTable` in the MLT specification uses a tabular, column-oriented layout.
It employs various lightweight compression schemes to encode column values efficiently.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
-A FeatureTable in the MLT specification is based on a tabular, column-oriented layout.
-It uses different lightweight compression schemes to efficiently encode the values of the columns.
-A FeatureTable consists of a optional `id` column, a mandatory `geometry` column and optional property columns.
-Because there is no tile header, `FeatureTable`s can be concatenated on the fly.
+A FeatureTable consists of a mandatory `geometry` column, an optional `id` column, and optional property columns. The absence of a single header at the beginning of the tile allows `FeatureTable`s to be constructed independently, and even concatenated on the fly.
```
CommanderStorm:
```suggestion
A FeatureTable consists of a mandatory `geometry` column, an optional `id` column, and optional property columns.
The absence of a single header at the beginning of the tile allows `FeatureTable`s to be constructed independently, and even concatenated on the fly.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
-A logical column is separated into several physical `streams` (sub-columns) inspired by the ORC file format which are stored next to each other. A stream is a sequence of values of a known length in a continuous chunk of memory, all having the same type. A stream has some additional metadata such as the size of the stream and the encoding type.
-For example, a nullable string property column can have a `present stream`, which indicates the presence of a value,
-a `length` stream, which describes the number of characters of every string, and a `data stream`, with the actual
-UTF-8 encoded string values.
+A logical column is separated into several physical `streams` (sub-columns), inspired by the ORC file format. These streams are stored contiguously. A stream is a sequence of values of a known length in a continuous memory chunk, all sharing the same type. Streams include additional metadata, such as their size and encoding type.
```
CommanderStorm:
```suggestion
A logical column is separated into several physical `streams` (sub-columns), inspired by the ORC file format.
These streams are stored contiguously.
A stream is a sequence of values of a known length in a continuous memory chunk, all sharing the same type.
Streams include additional metadata, such as their size and encoding type.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
-  Specifies the number of elements for variable-sized data types such as strings or list.
-- **offset**:
-  Stores the offsets into a data stream when dictionary encoding is used such as for strings or vertices.
+MLT defines the following stream types:
 
-This physical streams are further divided into logical streams that add additional semantics of how to interpret the data:
+- **Present**: Enables efficient encoding of sparse columns by indicating value presence via a bit flag. This stream can be omitted if the column is not nullable (as declared in the `FieldMetadata`).
+- **Data**: Stores the actual column data (e.g., `boolean`, `int`, `float`, or `string` values for feature properties, dictionary-encoded values, or geometry coordinates). For fixed-size data types (`boolean`, `int`, `float`), this is the only required stream besides the optional `present` stream.
+- **Length**: Specifies the number of elements for variable-sized data types like strings or lists.
+- **Offset**: Stores offsets into a data stream when using dictionary encoding (e.g., for strings or vertices).
```
CommanderStorm:
```suggestion
- **Present**:
  Enables efficient encoding of sparse columns by indicating value presence via a bit flag.
  This stream can be omitted if the column is not nullable (as declared in the `FieldMetadata`).
- **Data**:
  Stores the actual column data (e.g., `boolean`, `int`, `float`, or `string` values for feature properties, dictionary-encoded values, or geometry coordinates).
  For fixed-size data types (`boolean`, `int`, `float`), this is the only required stream besides the optional `present` stream.
- **Length**:
  Specifies the number of elements for variable-sized data types like strings or lists.
- **Offset**:
  Stores offsets into a data stream when using dictionary encoding (e.g., for strings or vertices).
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 ![](assets/spec/LogicalStreams.png)
 
 #### 1.2.1 Metadata
 
-##### Tileset Metadata
+##### Tileset Metadata <span class="experimental"/>
 
-Global metadata for the tileset is stored separately, in [JSON format](assets/spec/mlt_tileset_metadata.json).
+> [!NOTE]
+> Tileset metadata was initially implemented as a size reduction experiment. This feature is not currently supported.
```
CommanderStorm:
```suggestion
> Tileset metadata was initially implemented as a size reduction experiment.
> This feature is not currently supported.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
-The tileset metadata defines information for the full tileset is the equivalent to the TileJSON spec, which is commonly used in combination with MVT.
+Global metadata for the entire tileset is stored separately in a [JSON file](assets/spec/mlt_tileset_metadata.json).
 
-By specifying the information once per tileset, the definition of redundant (shared) metadata per tile will be avoided,
-which can take some relevant space of the overall size on small tiles.
+This tileset metadata provides information for the full tileset and is the equivalent of the TileJSON spec commonly used with MVT and other tile types. By defining this information once per tileset, we avoid redundant metadata in each tile, saving significant space, especially for small tiles.
```
CommanderStorm:
```suggestion
This tileset metadata provides information for the full tileset and is the equivalent of the TileJSON spec commonly used with MVT and other tile types.
By defining this information once per tileset, we avoid redundant metadata in each tile, saving significant space, especially for small tiles.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
 ##### Tile Metadata
 
-Each `FeatureTable` is preceded by a `FeatureTableMetadata` describing it.
+Each `FeatureTable` is preceded by a `FeatureTableMetadata` section describing it.
 
-A tile consists of any number of:
+> [!CAUTION]
+> This is not clear, and possibly incorrect. Why any number? Should the size of the upcoming metadata and table be part of that structure?
```
CommanderStorm:
```suggestion
> [!CAUTION]
> This is not clear, and possibly incorrect.
> Why any number?
> Should the size of the upcoming metadata and table be part of that structure?
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
+A FeatureTable consists of any number of the following sequences:
+- The size of the upcoming `FeatureTableMetadata` (varint-encoded).
+- The size of the upcoming `FeatureTable` (varint-encoded).
+- One `FeatureTableMetadata` section.
+- One `FeatureTable` section.
 
-The `FeatureTableMetadata` is described in detail below.
-
-Each `FeatureTableMetadata`/`FeatureTable` pair may contain one or more layers.  This allows the tile to be built by simple concatenation of separate results.
+This structure allows a tile to be built by simply concatenating separate results. The `FeatureTableMetadata` is described in detail below.
```
CommanderStorm:
```suggestion
This structure allows a tile to be built by simply concatenating separate results.
The `FeatureTableMetadata` is described in detail below.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
 Within a `FeatureTable`, additional metadata describes the structure of each part:
 
-- **FieldMetadata**: contains information about the number of streams the field is divided and the vector type for efficient decoding into the in-memory format. Every Field (column) section is preceded by a `FieldMetadata` section.
-- **StreamMetdata**: contains information about the Stream such as the used encoding schemes
-  and the number of values. Every Stream section is preceded by a `StreamMetadata` section.
+- **FieldMetadata**: Contains information about a field (column), including the number of streams it comprises and its vector type for efficient decoding into the in-memory format. Every field section is preceded by a `FieldMetadata` section.
+- **StreamMetadata**: Contains information about a stream, such as the encoding scheme used and the number of values. Every stream section is preceded by a `StreamMetadata` section.
```
CommanderStorm:
```suggestion
- **FieldMetadata**:
  Contains information about a field (column), including the number of streams it comprises and its vector type for efficient decoding into the in-memory format.
  Every field section is preceded by a `FieldMetadata` section.
- **StreamMetadata**:
  Contains information about a stream, such as the encoding scheme used and the number of values.
  Every stream section is preceded by a `StreamMetadata` section.
```

mkdocs/docs/specification.md
```diff
@@ -6,88 +6,79 @@
...
 
-Since for every `Field`, a `FieldMetadata` section has to be present, even for fields which are not included in one
-specific tile, no `id` is needed.
-The absence of an entire field in a tile can be identified by a zero value for the number of streams.
-All integers in the metadata sections are `Varint`-encoded if u32 is used or bit-packed in case of an u8 data type.
+Since every `Field` has a `FieldMetadata` section, even for fields absent in a specific tile, no `id` is needed. A field's absence is indicated by a zero value for its number of streams. All integers in metadata sections are `Varint`-encoded (for u32) or bit-packed (for u8).
```
CommanderStorm:
```suggestion
Since every `Field` has a `FieldMetadata` section, even for fields absent in a specific tile, no `id` is needed.
A field's absence is indicated by a zero value for its number of streams.
All integers in metadata sections are `Varint`-encoded (for u32) or bit-packed (for u8).
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
+#### 1.2.2 Type System
 
-#### 1.2.2 Type system
-
-The types system of MLT is divided into physical and logical types.
-The physical types define the layout of the data in the storage format, while the logical types give additional semantics to the physical types.
-Based on this separation, the encoding and decoding of the data can be simplified.
-This reduces the complexity for implementing encoder and decoder as well allow the encodings to be reused.
+The MLT type system distinguishes between physical and logical types. Physical types define the data layout in storage, while logical types add semantic meaning. This separation simplifies encoder and decoder implementation and allows encoding schemes to be reused.
```
CommanderStorm:
```suggestion
The MLT type system distinguishes between physical and logical types.
Physical types define the data layout in storage, while logical types add semantic meaning.
This separation simplifies encoder and decoder implementation and allows encoding schemes to be reused.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
 ##### 1.2.2.1 Physical Types
 
-The physical types define the layout of the data in the storage format.
-Both scalar and complex types can be divided into fixed-size binaries and variable-size binaries.
-Variable-size binaries have an additional length stream which specifies the size of each element.
-All fixed-size binaries have the same bit (boolean) or byte width and therefore no additional length stream is used.
+Physical types define the data layout in storage. Both scalar and complex types can be categorized as fixed-size or variable-size binaries. Variable-size binaries require an additional length stream to specify the size of each element. Fixed-size binaries have a consistent bit (boolean) or byte width and thus require no length stream.
```
CommanderStorm:
```suggestion
Physical types define the data layout in storage.
Both scalar and complex types can be categorized as fixed-size or variable-size binaries.
Variable-size binaries require an additional length stream to specify the size of each element.
Fixed-size binaries have a consistent bit (boolean) or byte width and thus require no length stream.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
+| --------- | -------------------- | -------------------------------------------------- | ------------- |
+| List      | Binary (List<UInt8>) |                                                    | Variable-Size |
+| Map       | Map<vec2d, T>        | Additional key stream -> length, key, data streams | Variable-Size |
+| Struct    |                      |                                                    |               |
+| Vec2<T>, Vec3<T> | Geometry, GeometryZ |                                                    | Fixed-Size    |
 
 ##### 1.2.2.2 Logical Types <span class="experimental"/>
 
+> [!CAUTION]
+> Original text had `encodings can be reused` text which is unclear. What is "encodings" in this context?
```
CommanderStorm:
```suggestion
> [!CAUTION]
> Original text had `encodings can be reused` text which is unclear.
> What is "encodings" in this context?
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
-In addition, nested fields can use a shared dictionary encoding, to share a common dictionary between the fields.
-This can be applied, for example, on localized values of the name:* columns of an OSM dataset that can be identical across fields.
-If a shared dictionary encoding is used for nested fields, all fields that use the shared dictionary
-must be grouped sequentially in the file and prefixed with the dictionary.
+For nested properties (e.g., structs, lists), a [present/length](https://arxiv.org/pdf/2304.05028.pdf) pair encoding is chosen over the Dremel encoding for its simpler implementation and faster decoding into the in-memory format.
+
+Every nullable field has an additional `present` stream. Every collection type field (e.g., a list) has an additional `length` stream specifying its length. As in ORC, nested fields are flattened based on a pre-order traversal.
+
+Nested fields can also use shared dictionary encoding to share a common dictionary (e.g., for localized `name:*` columns in an OSM dataset). Fields using a shared dictionary must be grouped sequentially in the file and prefixed by the dictionary.
```
CommanderStorm:
```suggestion
Every nullable field has an additional `present` stream.
Every collection type field (e.g., a list) has an additional `length` stream specifying its length.
As in ORC, nested fields are flattened based on a pre-order traversal.

Nested fields can also use shared dictionary encoding to share a common dictionary (e.g., for localized `name:*` columns in an OSM dataset).
Fields using a shared dictionary must be grouped sequentially in the file and prefixed by the dictionary.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
 ##### RangeMap <span class="experimental"/>
 
-RangeMaps are an efficient way to encode linear referencing information, as used for example in [Overture Maps](https://docs.overturemaps.org/overview/feature-model/scoping-rules#geometric-scoping-linear-referencing).
-RangesSets store the range values and data values in two separate streams.
-The min and max values for the ranges are stored as interleaved double values in a separate range stream.
+RangeMaps efficiently encode linear referencing information, as used in [Overture Maps](https://docs.overturemaps.org/overview/feature-model/scoping-rules#geometric-scoping-linear-referencing). RangeSets store range values and data values in two separate streams. The min and max values for the ranges are stored as interleaved double values in a separate range stream.
```
CommanderStorm: We are a bit inconsistent which values get code styling, but this likely does not matter that much

```suggestion
`RangeMaps` efficiently encode linear referencing information, as used in [Overture Maps](https://docs.overturemaps.org/overview/feature-model/scoping-rules#geometric-scoping-linear-referencing).
`RangeSets` store range values and data values in two separate streams. The min and max values for the ranges are stored as interleaved double values in a separate range stream.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
 #### 1.2.3 Encoding Schemes
 
-MLT uses different lightweight compression schemes for the space efficient storage and fast decoding of the data types.
-To further reduce the size of a column the encodings can be recursively cascaded (hybrid encodings) up to a certain level.
-For example the resulting integer columns of a dictionary encoding can be further compressed based on one of the Integer encoding schemes.
-The following encoding pool for specific data types was selected based on the analytics on the efficiency
-in terms of the compression ratio and decoding speed on test datasets such as the OpenMapTiles schema and Bing Maps based tilesets.
+MLT uses various lightweight compression schemes for space-efficient storage and fast decoding. Encodings can be recursively cascaded (hybrid encodings) to a certain degree. For example, integer columns resulting from dictionary encoding can be further compressed using integer encoding schemes.
```
CommanderStorm:
```suggestion
MLT uses various lightweight compression schemes for space-efficient storage and fast decoding.
Encodings can be recursively cascaded (hybrid encodings) to a certain degree.
For example, integer columns resulting from dictionary encoding can be further compressed using integer encoding schemes.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
+| String    | Plain, Dictionary, [FSST](https://www.vldb.org/pvldb/vol13/p2649-boncz.pdf) Dictionary | | |
+| Geometry  | Plain, Dictionary, Morton-Dictionary | | |
 
 > [!NOTE]
 > `ALP`, `FSST`, and `FastPFOR` encodings are <span class="experimental"></span>.
 
-Since SIMD-FastPFOR generally produces smaller data streams and is faster to decode, it should be preferred over Varint encoding.
-Varint encoding is mainly added to the encoding pool for compatibility reasons and it's simpler implementation compared to SIMD-FastPFOR.
-Varint encoding also can be more efficient when used in combination with heavyweight compression schemes like GZip.
+SIMD-FastPFOR is generally preferred over Varint encoding due to its smaller output and faster decoding speed. Varint encoding is included mainly for compatibility and simplicity, and it can be more efficient when combined with heavyweight compression like GZip.
```
CommanderStorm:
```suggestion
SIMD-FastPFOR is generally preferred over Varint encoding due to its smaller output and faster decoding speed.
Varint encoding is included mainly for compatibility and simplicity, and it can be more efficient when combined with heavyweight compression like GZip.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
-To find the best coding schemes for a specific data type from the coding pool, using a brute force approach by testing all available schemes
-with respect to the resulting column size would be too costly. The recommended approach improves the selection time by using
-the selection strategy described in the [BTRBlocks](https://www.cs.cit.tum.de/fileadmin/w00cfj/dis/papers/btrblocks.pdf) paper:
+A brute-force search for the best encoding scheme is too costly. Instead, we recommend the selection strategy from the [BTRBlocks](https://www.cs.cit.tum.de/fileadmin/w00cfj/dis/papers/btrblocks.pdf) paper:
```
CommanderStorm:
```suggestion
A brute-force search for the best encoding scheme is too costly.
Instead, we recommend the selection strategy from the [BTRBlocks](https://www.cs.cit.tum.de/fileadmin/w00cfj/dis/papers/btrblocks.pdf) paper:
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
-- First, calculation of data metrics to early exclude encodings. For example if the average run length is less than 2,
-  RLE encoding is not a candidate for the encodings pool
-- Second, use a sampling-based algorithm for choosing the best compression scheme out of the pool by taking randomly parts
-  of the data with a total size of 1% of the full data and apply the encoding schemes selected in step 1. Apply the encoding scheme
-  on the data which produces the smallest output
+- Calculate data metrics to exclude unsuitable encodings early (e.g., exclude RLE if the average run length is less than 2).
+- Use a sampling-based algorithm: randomly select parts of the data totaling ~1% of the full dataset and apply the candidate encodings from step 1. Choose the scheme that produces the smallest output.
```
CommanderStorm:
```suggestion
- Use a sampling-based algorithm: randomly select parts of the data totaling ~1% of the full dataset and apply the candidate encodings from step 1.
  Choose the scheme that produces the smallest output.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
 #### 1.2.4 FeatureTable Layout
 
 ##### ID Column
 
-No `id` column is mandatory.  If an identifier column is included, it should be u64 or a narrower integer type for compatibility with MVT.
-Narrow the column to u32, if possible, to enable the use of FastPfor128 encoding.
+An `id` column is not mandatory. If included, it should be a u64 or narrower integer type (u32 if possible) for MVT compatibility. A narrower type enables the use of efficient encodings like FastPfor128.
```
CommanderStorm:
```suggestion
An `id` column is not mandatory.
If included, it should be a u64 or narrower integer type (u32 if possible) for MVT compatibility.
A narrower type enables the use of efficient encodings like FastPfor128.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
 ##### Geometry Column
 
-The main idea is to use a structure of arrays (data-oriented design) layout for the geometries.
-The `x`, `y`, and optional `z` coordinates are stored interleaved in a `VertexBuffer` so that they can be efficiently processed
-on the CPU or even directly copied to GPU buffers. If the `z` coordinate is not relevant for rendering, it can be stored separately as
-an m-coordinate (see vertex-scoped properties).
+The geometry column uses a Structure of Arrays (SoA) layout (data-oriented design). The `x`, `y`, and optional `z` coordinates are stored interleaved in a `VertexBuffer` for efficient CPU processing and direct copying to GPU buffers. If the `z` coordinate is not needed for rendering, it can be stored separately as an `M-value` (see vertex-scoped properties).
```
CommanderStorm:
```suggestion
The geometry column uses a Structure of Arrays (SoA) layout (data-oriented design).
The `x`, `y`, and optional `z` coordinates are stored interleaved in a `VertexBuffer` for efficient CPU processing and direct copying to GPU buffers.
If the `z` coordinate is not needed for rendering, it can be stored separately as an `M-value` (see vertex-scoped properties).
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
-The information about the geometry of the features is separated in different streams and is partly inspired
-by the [geoarrow](https://github.com/geoarrow/geoarrow) specification.
-Using separate streams for describing the geometry of a feature enables a better optimization of the compression
-and faster processing. In addition, pre-tessellated polygon meshes can be stored directly in the file to avoid the time-consuming triangulation step.
+The geometry information is separated into different streams, partly inspired by the [geoarrow](https://github.com/geoarrow/geoarrow) specification. This separation enables better compression optimization and faster processing. Pre-tessellated polygon meshes can also be stored directly to avoid runtime triangulation.
```
CommanderStorm:
```suggestion
The geometry information is separated into different streams, partly inspired by the [geoarrow](https://github.com/geoarrow/geoarrow) specification.
This separation enables better compression optimization and faster processing.
Pre-tessellated polygon meshes can also be stored directly to avoid runtime triangulation.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
+
+Depending on the geometry type, the following streams are used in addition to `GeometryType`:
+- **Point**: VertexBuffer
+- **LineString**: NumParts, VertexBuffer
+- **Polygon**: NumParts (Polygon), NumRings (LinearRing), VertexBuffer
+- **MultiPoint**: NumGeometries, VertexBuffer
+- **MultiLineString**: NumGeometries, NumParts (LineString), VertexBuffer
+- **MultiPolygon**: NumGeometries, NumParts (Polygon), NumRings (LinearRing), VertexBuffer
+
+An additional `VertexOffsets` stream is present when using Dictionary or Morton-Dictionary encoding. If geometries (mainly polygons) are pre-tessellated for direct GPU use, `NumTriangles` and `IndexBuffer` streams must be provided.
```
CommanderStorm:
```suggestion
An additional `VertexOffsets` stream is present when using Dictionary or Morton-Dictionary encoding.
If geometries (mainly polygons) are pre-tessellated for direct GPU use, `NumTriangles` and `IndexBuffer` streams must be provided.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
-The properties of a feature are divided into `feature-scoped` and `vertex-scoped` properties.
-The values of feature-scoped properties are related to a specific Feature, which means there is one value in
-a property column per feature.
+Feature properties are divided into `feature-scoped` and `vertex-scoped` properties.
+- **Feature-scoped**: One value per feature.
+- **Vertex-scoped**: One value per vertex in the VertexBuffer per feature (modeling M-coordinates from GIS).
+
+> [!NOTE]
+> TODO: Would it make sense to place vertex-scoped properties AFTER feature scoped ones? I suspect some implementations may act of feature-scoped properties first, and possibly even ignore vertex-scoped, at least for now (esp since vertex-scoped ones are still experimental)
```
CommanderStorm:
```suggestion
> [!NOTE]
> TODO: Would it make sense to place vertex-scoped properties AFTER feature scoped ones?
> I suspect some implementations may act of feature-scoped properties first, and possibly even ignore vertex-scoped, at least for now (esp since vertex-scoped ones are still experimental)
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
-The values of a vertex-scoped properties are related to each vertex, which means there is one value in a property column
-per vertex in the VertexBuffer. This allows to model what is known as M-coordinates in GIS applications.
-Vertex-scoped properties have to be grouped together and are placed before the feature-scoped properties in the FeatureTable.
-The scope of a property column is specified in the tileset metadata document based on the `ColumnScope` enum.
+Vertex-scoped properties must be grouped together and placed before feature-scoped properties in the FeatureTable. A property's scope is defined in the tileset metadata using the `ColumnScope` enum.
```
CommanderStorm:
```suggestion
Vertex-scoped properties must be grouped together and placed before feature-scoped properties in the FeatureTable.
A property's scope is defined in the tileset metadata using the `ColumnScope` enum.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
-
-- blue boxes: columns that have only logical character and are not persisted.
-  Based on the TileSet metadata the fields can be reconstructed from the streams.
-- white boxes: metadata which describe the structure of the data. The metadata is divided into
-  FeatureTable, Stream (SM) and Feature (FM) metadata
-- yellow boxes: streams which contains the actual data
+The following examples illustrate the layout of a `FeatureTable` in storage. The color scheme is:
+- **Blue boxes**: Logical constructs, not persisted. Fields are reconstructed from streams based on TileSet metadata.
+- **White boxes**: Metadata describing data structure (FeatureTable, Stream (SM), Feature (FM) metadata).
+- **Yellow boxes**: Streams containing the actual data.
```
CommanderStorm:
```suggestion
- **Blue boxes**:
  Logical constructs, not persisted.
  Fields are reconstructed from streams based on TileSet metadata.
- **White boxes**:
  Metadata describing data structure (FeatureTable, Stream (SM), Feature (FM) metadata).
- **Yellow boxes**:
  Streams containing the actual data.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
-Encoding of a `FeatureTable` with an `id` field, a `LineString` geometry field and the flat feature scoped properties class and subclass:
+Encoding of a `FeatureTable` with an `id` field, a `LineString` geometry field, and the flat feature-scoped properties `class` and `subclass`:
 ![img_4.png](assets/spec/img_4.png)
 
-#### 1.3.3 MultiPolygon with flat properties
+#### 1.3.3 MultiPolygon with Flat Properties
 
-Encoding of a `FeatureTable` with a `id` field, `MultiPolygon` geometry field and flat feature scoped property fields.
-Because vertex dictionary encoding is used a `VertexOffsets` stream is present:
+Encoding of a `FeatureTable` with an `id` field, a `MultiPolygon` geometry field, and flat feature-scoped property fields. A `VertexOffsets` stream is present due to vertex dictionary encoding:
```
CommanderStorm:
```suggestion
Encoding of a `FeatureTable` with an `id` field, a `MultiPolygon` geometry field, and flat feature-scoped property fields.
A `VertexOffsets` stream is present due to vertex dictionary encoding:
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 ![img_5.png](assets/spec/img_5.png)
 
-#### 1.3.4 Vertex-scoped and feature-scoped properties
+#### 1.3.4 Vertex-Scoped and Feature-Scoped Properties
 
-Example layout for encoding of vertex-scoped and feature scoped properties.
-All vertex-scoped properties have to be grouped together and placed before the feature-scoped properties
-in the file. Since the `id` colum in this example is not `nullable`, the present stream can be omitted.
+Example layout encoding vertex-scoped and feature-scoped properties. All vertex-scoped properties are grouped together and placed before feature-scoped properties. The `id` column is not nullable, so its present stream is omitted.
```
CommanderStorm:
```suggestion
Example layout encoding vertex-scoped and feature-scoped properties.
All vertex-scoped properties are grouped together and placed before feature-scoped properties.
The `id` column is not nullable, so its present stream is omitted.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 ![img_7.png](assets/spec/img_7.png)
 
 ### Sorting
 
-Choosing the right column for sorting the features can have significant impact on the size
-of the `FeatureTable`. To take full advantage of the columnar layout, sorting is crucial.
-To test every layer for every possible sorting order of every column is too costly.
+Choosing the right column to sort features by can significantly reduce the size of the `FeatureTable`. Sorting is crucial for leveraging the columnar layout fully. Exhaustively testing every possible sorting order for every column in every layer is computationally expensive. See recommended heuristic in the [encoding schemes](#123-encoding-schemes).
```
CommanderStorm:
```suggestion
Choosing the right column to sort features by can significantly reduce the size of the `FeatureTable`.
Sorting is crucial for leveraging the columnar layout fully.
Exhaustively testing every possible sorting order for every column in every layer is computationally expensive.
See recommended heuristic in the [encoding schemes](#123-encoding-schemes).
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
-
-By using a columnar memory layout (data-oriented design), for the MLT in-memory format, we can overcome these issues and
-additionally take advantage of an improved cache utilization for subsequent access and fast SIMD instructions for the processing of the data.
-The MLT in-memory format incorporates and extends the ideas from the analytical in-memory formats Apache Arrow, Velox, and the DuckDB execution
-format and is tailored for the visualization use case. To be future-proof it is important that the file can be parallel processed on the GPU for the
-usage within a GPU compute shader.
+> The following is a high-level overview; the in-memory format will be explained in more detail later.
 
-The main design goals of the MLT in-memory format are:
+The record-oriented, array-of-structures in-memory model used by libraries processing Mapbox Vector Tiles incurs considerable overhead. This includes creating many small objects (increasing memory allocation load) and placing additional strain on garbage collectors in browsers.
```
CommanderStorm:
```suggestion
The record-oriented, array-of-structures in-memory model used by libraries processing Mapbox Vector Tiles incurs considerable overhead.
This includes creating many small objects (increasing memory allocation load) and placing additional strain on garbage collectors in browsers.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
 
-- Define a platform-agnostic representation to avoid the expensive materialization costs in particular for strings
-- Enable advanced CPU throughput by optimizing the memory layout for cache locality and SIMD instructions
-- Allow random (mainly constant time) access to all data, so it can also be parallel processed on the GPU (compute shader)
-- Provide compressed data structure which can be directly processed without decoding
-- Provide the geometries of a tile in a representation that can be (directly) loaded into GPU buffers with little or no additional processing
+MLT uses a columnar memory layout (data-oriented design) for its in-memory format to overcome these issues. This approach improves cache utilization for subsequent data access and enables the use of fast SIMD instructions. The MLT in-memory format incorporates ideas from analytical in-memory formats like Apache Arrow, Velox, and the DuckDB execution format, tailored for visualization use cases. It is also designed for future parallel processing on the GPU within compute shaders.
```
CommanderStorm:
```suggestion
MLT uses a columnar memory layout (data-oriented design) for its in-memory format to overcome these issues.
This approach improves cache utilization for subsequent data access and enables the use of fast SIMD instructions.
The MLT in-memory format incorporates ideas from analytical in-memory formats like Apache Arrow, Velox, and the DuckDB execution format, tailored for visualization use cases.
It is also designed for future parallel processing on the GPU within compute shaders.
```

mkdocs/docs/specification.md
```diff
@@ -246,246 +237,197 @@ classDiagram
...
-an auxiliary buffer containing offset values will be utilized.
+The main design goals for the MLT in-memory format are:
+- Define a platform-agnostic representation to avoid expensive materialization costs, especially for strings.
+- Maximize CPU throughput by optimizing memory layout for cache locality and SIMD instructions.
+- Allow random (preferably constant-time) access to all data for parallel processing on GPUs (compute shaders).
+- Provide compressed data structures that can be processed directly without full decoding.
+- Provide tile geometries in a representation that can be loaded into GPU buffers with minimal additional processing.
 
-The MLT in-memory format supports the following vectors:
+Data is stored in contiguous memory buffers called **vectors**, accompanied by metadata and an optional null bitmap. The storage format includes a `VectorType` field in the metadata to instruct the decoder which vector type to use for a specific field. An auxiliary offset buffer enables random access to variable-sized data types like strings or lists.
```
CommanderStorm:
```suggestion
Data is stored in contiguous memory buffers called **vectors**, accompanied by metadata and an optional null bitmap.
The storage format includes a `VectorType` field in the metadata to instruct the decoder which vector type to use for a specific field.
An auxiliary offset buffer enables random access to variable-sized data types like strings or lists.
```

CommanderStorm (commented):
Sorry to not this much, should be easy enough to apply.

Most of these are newline changes to make review simpler

mkdocs/docs/encodings.md
nyurik: thx, fixed

mkdocs/docs/encodings.md
nyurik: not sure what you mean. Delta encoding is fairly well defined in the industry, and at one point I even made this https://crates.io/crates/delta-encoding - what kind of a change you think would help?

mkdocs/docs/specification.md
nyurik: Do we want to mark all instances of `FeatureTable` with back ticks, or just the first occurrence of each?

mkdocs/docs/encodings.md
CommanderStorm: I am not refering to delta encoding.
That should be clear to anybody.

I am refering to the line
> Used with physical-level techniques to reduce the number of bits required for storing delta values.

Given that "physical-level techniques" are only mentioned later in the spec, either linking that or noting that this is explained later would help

nyurik: done

mkdocs/docs/encodings.md
nyurik: I don't have an answer to that - perhaps we should add a TODO warning? I would like to merge this before we start doing the meaning changes

CommanderStorm:
```suggestion
> [!NOTE]
> TODO: Is "can" (=> MAY) here the right terminoligy?
> This implies that streams might be encoded by one of the options below, but also other options.
> 
> If yes, what is the alternative options?
> If no, clarify this misunderstanding.

All streams can be further compressed using lightweight encodings:
```

mkdocs/docs/encodings.md
```diff
@@ -4,155 +4,153 @@
...
+A dictionary-encoded nullable string column consists of the following streams in order:
+`Present`, `Length`, `Dictionary`, `Data`
 
-All streams can be further compressed by recursively applying lightweight encodings:
+> [!NOTE]
+> TODO: Is "can" (=> MAY) here the right terminoligy?
+> This implies that streams might be encoded by one of the options below, but also other options.
+>
+> If yes, what is the alternative options?
+> If no, clarify this misunderstanding.
```
mactrem, line 47: I would prefer "can" because present, length and data streams can be also encoded using [plain encoding](https://github.com/maplibre/maplibre-tile-spec/blob/f48a66001c57e1113212dbef13e8b797568eaea1/mkdocs/docs/encodings.md#plain), also FSST Dictionary is an optional optimization

mkdocs/docs/specification.md
```diff
@@ -248,245 +270,237 @@ classDiagram
...
+- **Feature-scoped**: One value per feature.
+- **Vertex-scoped**: One value per vertex in the VertexBuffer per feature (modeling M-coordinates from GIS).
 
-The values of a vertex-scoped properties are related to each vertex, which means there is one value in a property column
-per vertex in the VertexBuffer. This allows to model what is known as M-coordinates in GIS applications.
-Vertex-scoped properties have to be grouped together and are placed before the feature-scoped properties in the FeatureTable.
-The scope of a property column is specified in the tileset metadata document based on the `ColumnScope` enum.
+> [!NOTE]
+> TODO: Would it make sense to place vertex-scoped properties AFTER feature scoped ones?
+> I suspect some implementations may act of feature-scoped properties first, and possibly even ignore vertex-scoped, at least for now (esp since vertex-scoped ones are still experimental)
```
mactrem, line 421: yes, since we have agreed that m-coordinates are experimentel in version 1 we could change the order

mkdocs/docs/specification.md
```diff
@@ -248,245 +270,237 @@ classDiagram
...
 
 ### Logical Types <span class="experimental"></span>
 
+> [!CAUTION]
+> Original text had `encodings can be reused` text which is unclear.
+> What is "encodings" in this context?
 
-Add additional semantics on top of the physical types.
-This had the advantage that encodings can be reused and implementation of encoders and decoders can be simplified.
+Logical types add semantics on top of physical types, enabling code reuse and simplifying encoder/decoder implementation.
```
mactrem, line 314: This changes are correct, it is more precise to use the term `code reuse`. The context are all encodings defined in the [Encodings spec]( https://github.com/maplibre/maplibre-tile-spec/blob/f48a66001c57e1113212dbef13e8b797568eaea1/mkdocs/docs/encodings.md). For example for the Date and Timestamp logical types the interger encodings can be reused and no dedicated seperate encodings have to be defined for this types

mkdocs/docs/specification.md
```diff
@@ -248,245 +270,237 @@ classDiagram
...
+| Data Type | Logical Level Technique      | Physical Level Technique            | Description |
+| --------- | ---------------------------- | ----------------------------------- | ----------- |
+| Boolean   | [Boolean RLE](https://orc.apache.org/specification/ORCv1/#boolean-run-length-encoding) | | |
+| Integer   | Plain, RLE, Delta, Delta-RLE | [SIMD-FastPFOR](https://arxiv.org/pdf/1209.2137.pdf), [Varint](https://protobuf.dev/programming-guides/encoding/#varints) | |
+| Float     | Plain, RLE, Dictionary, [ALP](https://dl.acm.org/doi/pdf/10.1145/3626717) | | |
+| String    | Plain, Dictionary, [FSST](https://www.vldb.org/pvldb/vol13/p2649-boncz.pdf) Dictionary | | |
+| Geometry  | Plain, Dictionary, Morton-Dictionary | | |
 
 > [!NOTE]
 > `ALP`, `FSST`, and `FastPFOR` encodings are <span class="experimental"></span>.
```
mactrem, line 359: `experimental` is not rendered correctly. However, i would not call the FSST and FastPfor encodings experimental. On the TypeScript decoder side FSST is already working but FastPFOR not yet, but we are working on that. The Java decoder, which for sure is not the important one, is supporting both FSST and FastPfor already. @TimSylvester what is the status on the MLT CPP  decoder are you already decoding FSST and FastPfor encodings?

mkdocs/docs/specification.md
```diff
@@ -497,11 +511,8 @@ The MLT in-memory format supports the following vectors:
 - [Run-End Encoded (REE) Vectors](https://arrow.apache.org/docs/format/Columnar.html#run-end-encoded-layout)
 
 > [!NOTE]
-> Further evaluation is needed to determine if the [latest research findings](https://arxiv.org/pdf/2306.15374.pdf)
-> can be used to enable random access on delta encoded values as well
+> Further evaluation is needed to determine if [recent research](https://arxiv.org/pdf/2306.15374.pdf) can enable random access on delta-encoded values.
```
mactrem, line 514: we can remove that line completely because such research statements should not be part of a final spec

mkdocs/docs/specification.md
```diff
@@ -8,88 +8,110 @@
...
+Each `FeatureTable` is preceded by a `FeatureTableMetadata` section describing it.
 
-- The size of the `FeatureTableMetadata` to follow, varint-encoded
-- The size of the `FeatureTable` to follow, varint-encoded
-- One `FeatureTableMetadata`
-- One `FeatureTable`
+> [!CAUTION]
+> This is not clear, and possibly incorrect.
+> Why any number?
+> Should the size of the upcoming metadata and table be part of that structure?
```
mactrem, line 90: @TimSylvester can answer that best

mactrem:
Thanks for your work @nyurik!

mkdocs/docs/encodings.md
```diff
@@ -4,155 +4,153 @@
...
-#### Combining delta and rle (Delta-RLE)
+### Delta-RLE
 
-Delta-compress the values and then apply RLE encoding.
-This technique is efficient for ascending sequences like `id`s.
+Applies delta encoding followed by RLE.
+Efficient for ascending sequences like `id` fields.
 
-### Physical Level Technique (Null Suppression)
+## Physical-Level Techniques
```
louwers, line 116: This section is still unclear. @mactrem could you explain?

mkdocs/docs/encodings.md
```diff
@@ -4,155 +4,153 @@
...
+Null suppression techniques are used to compress integer arrays by reducing the number of bits required to store each integer.
 
-####  ZigZag encoding
+### ZigZag Encoding
 
-For encoding unsigned integers, ZigZag encoding is used for both of the following null suppression techniques.
-[ZigZag](https://en.wikipedia.org/wiki/Variable-length_quantity#Zigzag_encoding) encoding uses the least significant bit for encoding the sign.
+Used for encoding unsigned integers in null suppression techniques.
```
louwers, line 121: Original was "ZigZag encoding is used for both of the following", but now it lists only one? @mactrem is there another one?

mkdocs/docs/encodings.md
louwers: Also shouldn't Null Suppression be in the heading?

mkdocs/docs/encodings.md
```diff
@@ -4,155 +4,153 @@
...
+A byte-aligned null suppression technique that compresses integers using a minimal number of bytes.
+For implementation details, refer to [Protobuf](https://protobuf.dev/programming-guides/encoding/#varints).
 
-#### SIMD-FastPFOR
+### SIMD-FastPFOR
 
-Bit-aligned null suppression technique that compresses an integer using a minimal number of bits
-by using a patched approach to store exceptions (outliers) in a separate section to keep the overall bit width small.
+A bit-aligned null suppression technique that compresses integers using a minimal number of bits.
+Uses a patched approach to store exceptions (outliers) separately, keeping the overall bit width small.
```
louwers, line 132: Here 'in a separate section' is changed to 'separately'. Not sure if that is an improvement.

mkdocs/docs/encodings.md
mactrem: What do you mean by "lists only one"? Varint Encoding and SIMD-FastPFOR are categorized as null suppression techniques and zigZag encoding can be applied to both of these techniques

mkdocs/docs/encodings.md
mactrem: In our context null suppression, so removing leading zeros from small integer, is the only applied physical level technique for integers, so both can be basically used interchangeable

mkdocs/docs/specification.md
louwers: Let's do this in another PR, I think we should put it on a separate page.

CommanderStorm (approved):

PR #531, opened 2025-09-17
Add basic Bazel target for C++ library

louwers:


TimSylvester:
I just found this in the PRs after missing the notification about it, sorry.

The `.lock` file shouldn't be included, should it?

louwers:
@TimSylvester I think generally it is recommended to commit lock files, but for some reason it is not common in the Bazel ecosystem. So let's do some cargo culting, removed it!

TimSylvester (approved):

PR #535, opened 2025-09-18
feat(rust): experiment with tests, `nom`, ownership

nyurik:
* introduce auto-generated two variants of each struct - owned and borrowed (zero copy)


rust/mlt-nom/src/structures/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+            .ok_or(NomError(Error::new(input, ErrorKind::Fail)))?;
+        let (input, value) = take(size)(input)?;
+
+        let layer = match tag {
+            1 => {
+                let (data, meta) = FeatureMetaTable::parse(value)?;
+                Layer::Layer(FeatureTable::parse(data, meta)?)
+            }
+            tag => Layer::Unknown(Unknown { tag, value }),
+        };
```
CommanderStorm, line 45: could we add a comment why this exists? Seems a bit strange to only have one option (tag=1) for something.

rust/mlt-nom/src/structures/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub fn parse(input: &[u8]) -> IResult<&[u8], Layer<'_>> {
+        let (input, size) = utils::parse_varint_usize(input)?;
+
+        // tag is a varint, but we know fewer than 127 tags for now,
+        // so we can use a faster u8 and fail if it is bigger than 127.
+        let (input, tag) = utils::parse_u8(input)?;
+        // 1 byte must be parsed for the tag, so if size is 0, it's invalid
+        let size = size
+            .checked_sub(1)
+            .ok_or(NomError(Error::new(input, ErrorKind::Fail)))?;
```
CommanderStorm, line 35: nit: A bit of a nicer error message than just "fail" would be nice, if not it is also fine.

rust/mlt-nom/src/structures/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+        let (input, value) = take(size)(input)?;
```
CommanderStorm, line 36: nit: `value` seems like bad naming, not sure what is happening below, though.

rust/mlt-nom/src/structures/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+/// Unknown layer data
+#[borrowme]
+#[derive(Debug, PartialEq)]
+pub struct Unknown<'a> {
+    pub tag: u8,
+    #[borrowme(borrow_with = Vec::as_slice)]
+    pub value: &'a [u8],
```
CommanderStorm, line 57: I am not sure if storing the actual value of the data if unknown is smart, but fine

nyurik: `&[u8]` is effectively the size of the unknown block to be skipped. Most of the time its not needed, but it has near zero cost, and may be used by users for things like dumping into bin files for debugging  purposes

rust/mlt-nom/src/structures/mod.rs
nyurik: i try to use all `nom` parsers as returning `input` (i.e. the data that hasn't been eaten yet), and the value - the result of parsing. Technically, the value could be of any eaten type, so for consistency i call them `value`

PR #540, opened 2025-09-23
feat(rust): add tests files and rust code to use them

nyurik:
These test files probably should not be added to the repo, but instead the CI should generate them on the fly. For now, adding them here to help with testing. @Weixing-Zhang and @jjcfrancisco feel free to create PRs against this branch until we can get all tests to pass - i.e. each bin file should be parsable based on the corresponding metadata, and should match the json versions


nyurik:
this is now dynamically generated as part of CI with #541

PR #544, opened 2025-09-24
refactor(rust): improves boolean decoder organization

Weixing-Zhang:
Context
- The PR was originally intended to enable the `test_decode_fixtures` unit test, but this was not possible at the current stage. As the work progressed, it became apparent that separating the boolean decoder would help with future enabling efforts.

Changes
- Separated the boolean decoder from the helper function
- Update result type for int_decoder
- Relocated unit tests accordingly

Related Issues
- References issue #569 regarding Java encoder boolean truncation bug

### DIFF
--- a/rust/mlt/src/decoder/mod.rs
+++ b/rust/mlt/src/decoder/mod.rs
@@ -1,3 +1,4 @@
+pub mod boolean;
 mod decode;
 mod helpers;
 pub mod integer;
@@ -77,7 +78,7 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "not all parsing has been implemented yet"]
+    #[ignore = "not all parsing has been implemented yet - some boolean test fixtures have incorrect JSON output due to Java encoder bug (see issue #569)"]
     fn test_decode_fixtures() {
         for (name, path) in &get_bin_fixtures() {
             let meta = fs::read(path.with_extension("meta.bin")).expect(name);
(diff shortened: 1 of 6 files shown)


rust/mlt/src/decoder/mod.rs
```diff
@@ -82,9 +82,19 @@ mod tests {
...
 
             // TODO(Weixing): implement the rest of the decoding
+            eprintln!("----------------------------------------------");
+            eprintln!("fixture name = {name}");
+            eprintln!("fixture path = {}", path.display());
+            eprintln!("decoded meta = {:#?}", meta);
+            eprintln!("raw data stream (hex): {:02x?}", data);
+
+            // read the expected json file
+            let expected = fs::read_to_string(path.with_extension("json")).expect(name);
```
nyurik: note that the file content might have newline at the end.  It might make sense to compare as values rather than as strings - or maybe to parse the json file and convert it back to string?

rust/mlt/src/decoder/mod.rs
```diff
@@ -81,30 +83,45 @@ mod tests {
...
+            eprintln!("raw data stream (hex): {:02x?}", data);
 
+            // read the expected json file
             let expected = fs::read_to_string(path.with_extension("json")).expect(name);
-            assert_eq!(
-                serde_json::to_string(&result).expect(name),
-                expected,
-                "case {name}"
-            );
+            eprintln!("expected: {}", expected);
```
nyurik: just FYI - an easy way to debug is to add `dbg!(expected);` - much shorer

rust/mlt/src/decoder/mod.rs
Weixing-Zhang: Those were for initial testing, but they did not work out. See below.

```
The PR was originally intended to enable the test_decode_fixtures unit test, but this was not possible at the current stage. As the work progressed, it became apparent that separating the boolean decoder would help with future enabling efforts.
```

jjcfrancisco (approved):

PR #547, opened 2025-09-24
Implement new LTV top-level tile format

TimSylvester:
- Implement new LTV top-level tile format
- Column and field encoding unchanged
- Eliminate support for separate metadata output
- Refactor decoding methods to `MltDecoder`
- Simplify column mappings
- Remove some vectorized cruft
- Disable checking of expected files, they are not re-encoded as this is an intermediate step


TimSylvester:
I'm going to update the column/field encodings first.

PR #548, opened 2025-09-25
Reduce encoded size of column metadata

TimSylvester:
- Collapses the encoding of physical/logical,scalar/complex types into a single value range.
- Eliminates stream counts from scalar property columns other than string
- Eliminates present streams from non-nullable columns

### DIFF
--- a/java/src/main/java/org/maplibre/mlt/converter/encodings/MltTypeMap.java
+++ b/java/src/main/java/org/maplibre/mlt/converter/encodings/MltTypeMap.java
@@ -1,0 +1,201 @@
+package org.maplibre.mlt.converter.encodings;
+
+import java.util.Optional;
+import javax.annotation.Nullable;
+import org.maplibre.mlt.metadata.tileset.MltTilesetMetadata;
+
+public class MltTypeMap {
+  public static class Tag0x01 {
+    /// Produces the unique type encoding for a `Column` or `Field`
+    public static Optional<Integer> encodeColumnType(
+        @Nullable MltTilesetMetadata.ScalarType physicalScalarType,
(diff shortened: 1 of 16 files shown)


java/src/main/java/org/maplibre/mlt/converter/encodings/MltTypeMap.java
```diff
@@ -0,0 +1,201 @@
...
+            case FLOAT -> Optional.of(isNullable ? 14 : 15);
+            case DOUBLE -> Optional.of(isNullable ? 16 : 17);
+            case STRING -> Optional.of(isNullable ? 18 : 19);
+            default -> Optional.empty();
+          };
+        }
+      } else if (logicalScalarType != null) {
+        if (logicalScalarType == MltTilesetMetadata.LogicalScalarType.ID) {
+          return Optional.of(isNullable ? (hasLongIDs ? 20 : 21) : (hasLongIDs ? 22 : 23));
+        }
```
nyurik, line 37: can you make all the un-named types to use the first set of values, e.g.:

* ID / non-nullable / int32 = 0
* ID / non-nullable / int64 = 1
* ID / nullable / int32 = 2
* ID / nullable / int64 = 3
* Mix Geometry (always non-nullable) = 4
* skip until 10 just in case we figure out that non-mixed (all points / all polylines /...) would make sense before we finish with 0x01
* boolean prop / non-nullable = 10
* boolean prop / nullable = 11
* ...

P.S. note that int32/int64 might be an optimization on the stream type layer, i.e. if all values are int32, we can use fastpfor32 vs if they are int64, use some bigger compression - in which case instead of `ID / int32` it becomes `ID / fastpfor32` vs `ID / RLE` vs ...

nyurik (approved):

PR #549, opened 2025-09-25
Add encoding server

adrian-cojocaru:
This PR adds a development server that acts as a proxy for style/source/tile requests and encodes tile responses to MLT.

- add server mode to the java cli encoder (avoid the jvm startup cost)
- tested with the native [implementation](https://github.com/maplibre/maplibre-native/pull/3246)
- fix exception in `GeometryEncoder.java`

### DIFF
--- a/java/encoding-server/server.mjs
+++ b/java/encoding-server/server.mjs
@@ -1,0 +1,16 @@
+import express from 'express';
+
+import config from './config.mjs'
+import { convertStyleRequest, convertSourceRequest, convertTileRequest, runCLISetup } from './convert.mjs'
+
+const app = express();
+
+app.use('/style', convertStyleRequest);
+app.use('/source', convertSourceRequest);
+app.use('/tile', convertTileRequest);
+
+app.listen(config.port, config.host, () => {
+  runCLISetup();
+
+  console.log(`Server started on port ${config.port}`);
+});
--- a/java/src/main/java/org/maplibre/mlt/converter/encodings/GeometryEncoder.java
+++ b/java/src/main/java/org/maplibre/mlt/converter/encodings/GeometryEncoder.java
@@ -618,10 +618,10 @@ public class GeometryEncoder {
     }
 
     // TODO: get rid of that separate calculation
-    var minVertexValue =
+    var minVertexValue = vertexBuffer.isEmpty() ? Integer.MAX_VALUE :
         Collections.min(
             vertexBuffer.stream().flatMapToInt(v -> IntStream.of(v.x(), v.y())).boxed().toList());
-    var maxVertexValue =
+    var maxVertexValue = vertexBuffer.isEmpty() ? -Integer.MAX_VALUE :
         Collections.max(
             vertexBuffer.stream().flatMapToInt(v -> IntStream.of(v.x(), v.y())).boxed().toList());
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.class
 *.log
+.DS_Store
 .idea/
 .vscode/
 bazel-*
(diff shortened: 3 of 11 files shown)


nyurik:
thx!  My biggest concern with this is if this is the overall path we want to pursue?  Java codebase was mostly used for experimentations, but we are actively working on the Rust encoder and decoder, and once ready, it will be easily integratable into Martin and other tooling...  I'm just worried this code might go unmaintained soon... maybe

sjg-wdw:
We have a working Java path so we'll use that.

adrian-cojocaru:
@nyurik It's mostly a node app using a cli tool. It can be adapted to use anything since it's main focus was to handle requests and delegate the encoding to an actual encoder.

nyurik:
we should fix merge conflict if we want to proceed with this

sjg-wdw:
He's off for a few days.  Should look at it when he's back.

java/src/main/java/org/maplibre/mlt/converter/encodings/GeometryEncoder.java
```diff
@@ -624,11 +624,21 @@ public static EncodedGeometryColumn encodeGeometryColumn(
 
     // TODO: get rid of that separate calculation
     var minVertexValue =
-        Collections.min(
-            vertexBuffer.stream().flatMapToInt(v -> IntStream.of(v.x(), v.y())).boxed().toList());
+        vertexBuffer.isEmpty()
```
TimSylvester: This should no longer be necessary, as layers with no features are skipped at a higher level.

java/encoding-server/server.mjs
HarelM, line 1: Why not use something more java oriented? Feels weird to have javascript as part of a java project as I would expext people familiar with java to use something more familier if they needed to maintain this...

sjg-wdw: Adrian broke this into two pieces.  The Java part is more focused on translation and closer to the CLI implementation.
As to why the server portion is NodeJS, I believe he just finds Java networking annoying.  Which is fair.

louwers: The way I see it, it's just a tool for development. And it works pretty well for that.

It is not a robust, feature rich tile server to be used in production like Martin.

HarelM: I personally don't mind, but I'm not an expert java developer, and I would assume keeping this folder with java code only would make future maintenance easier, but I might be wrong...

louwers: The CLI is decoupled from the Java library, so if someone wants to write a better JVM-based tile server, they can do it.

sjg-wdw: We have yet to hear a compelling case for realtime MVT to MLT translation.  If there is one, then this functionality should probably go into Martin.  In the mean time it's meant for debugging and development.

HarelM: I haven't looked deeply into the code, but for development, this seems like a lot of code...

sjg-wdw: I might be missing something, but I just see a handful of source and a lot of the usual NodeJS cruft.

HarelM: I think that it might just be misplaced,  maybe, if this can be beneficial to other maintainers (not just java), maybe it should be moved to a different folder?

sjg-wdw: Probably, yeah.  Do we want to do that now or as part of the larger reorganization to split the spec out from the tools?

HarelM: Since this is a new file(s), I would consider do only this as part of this PR and later on move other stuff as part of a different PR. My 2 cents.

PR #550, opened 2025-09-25
fix buffer size calculations

TimSylvester:


nyurik (approved):
Copilot seems to have correctly suggested a change. Otherwise seems good

TimSylvester:
I'm sticking with the minimal change, int will always be 4 bytes in java

PR #556, opened 2025-09-29
Add test case for 5-byte varint case

TimSylvester:
Maybe I'm missing something here, but there seems to be a problem with the varint decoding, though it's surprising that we didn't run into it yet.  This is the failing test case based on troubleshooting an error I got on a real tile.

@mactrem Is there a canonical set of test cases for the varint encoding we're using?  ~~There seem to be several slightly different variations on it.~~

I found the link to [Bazel](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/util/VarInt.java).  There doesn't seem to be any testing of that varint code in the [test tree](http://github.com/bazelbuild/bazel/tree/master/src/test/java/com/google/devtools/build/lib/util).

But the decode implementation isn't quite the same, so I'll consider that confirmation.

### DIFF
--- a/java/src/main/java/org/maplibre/mlt/converter/MltConverter.java
+++ b/java/src/main/java/org/maplibre/mlt/converter/MltConverter.java
@@ -486,7 +486,8 @@ public class MltConverter {
           PhysicalLevelTechnique physicalLevelTechnique,
           boolean encodePolygonOutlines,
           @Nullable URI tessellateSource,
-          @Nullable HashMap<String, Triple<byte[], byte[], String>> rawStreamData) {
+          @Nullable HashMap<String, Triple<byte[], byte[], String>> rawStreamData)
+          throws IOException {
     /*
      * Following simple strategy is currently used for ordering the features when sorting is enabled:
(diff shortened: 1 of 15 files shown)


nyurik (approved):

mactrem:
@TimSylvester  Since the Java decoder was built only for research purposes, its test coverage is limited. Maintaining two separate decoder implementations was too time consuming, so I focused exclusively on making the JavaScript decoder fast and stable. The JavaScript version should be fine, with tests covering cases that involve more than four bytes see https://github.com/mactrem/mlt-evaluation/blob/main/ts/test/unit/encodings/integerDecodingUtils.spec.ts.

mactrem:
Also there are many end-to-end tests with various tilesets like different OpenMapTiles datasets, Overture Maps, SwissTopo, ... which are testing against MVT and all tests passing successfully. Therefore, this Js codebase should be quite stable see for example https://github.com/mactrem/mlt-evaluation/blob/main/ts/test/unit/mltDecoder.spec.ts

PR #559, opened 2025-09-30
Make the compare option more useful

TimSylvester:
- When using the `--printmlt` or `--printmvt` options, print the property values in key order and omit null values from the MLT, so that the outputs can be usefully diff-ed.
- Split the compare option into `--compare-geom` and `--compare-prop`.
- Disregard null properties when comparing property collections.
- Consider identical values of different types to be equivalent.
- Show the differences in a more useful format, each on a new line so that they align.


TimSylvester:
@copilot re-review

nyurik (approved):

PR #560, opened 2025-10-01
Add implementation status page, deem MLT stable

louwers:
This PR removes the warning in the README that MLT is "still in active research and development".

Instead, it adds a note that the specification is deemed stable as of October 2025, but explains that MLT is a living standard with experimental features that may undergo changes in the future. Also, it links to a new page with the current status of implementations and intergrations.


TimSylvester (approved):

mactrem (approved):
Lgtm. @louwers thanks for adding the reference to the paper!

PR #577, opened 2025-10-08
feat: add `generate-expected-mlt` to root justfile

nyurik:


nyurik:
made this part of #576

PR #579, opened 2025-10-08
feat(rust): update mlt-nom to tag0x01

nyurik:
massive rewrite of the rust/mlt-nom implementation - now supports "raw parsing" (mostly skips over columns), plus offers a way to "materialize" individual columns as needed, plus a way to serialize it back into a file (WIP), and lots of testing

# Concepts
Several basic types keep the same patterns:
* `Raw...` structs that holds references to the original input data, with its lifetime tied to the input `&[u8]` slice. No decoding is done at this stage, just some metadata parsing
* `OwnedRaw...` structs are auto-generated with the `borrowme` crate - they has the same structure as the `Raw...` structs, but own their bytes.
* `Decoded...` structs will be used to store `decoded` values - see geometry example
* enums containing `Raw(Raw...)` and `Decoded(Decoded...)` variants - this allows `in-place` decoding - e.g. it is possible to decode just one property column, while keeping others in their raw form. The enum also has a corresponding auto-generated `borrowme` owned copy that also starts with `Owned...`

When parsing, code very quickly runs over the input structs and only stores references to the streams and their metadata. At some point, I will implement all the decoders as well, just like the geometry one


rust/mlt-nom/Cargo.toml
```diff
@@ -10,14 +10,22 @@ categories.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-borrowme = "0.0.15"
-integer-encoding = "4.0.2"
-nom = "8.0.0"
-num_enum = "0.7.4"
+#mlt = { path = "../mlt" }
 #mvt-reader = "2.1.0"
```
CommanderStorm, line 16: Are these comments intentionall?

rust/mlt-nom/justfile
```diff
@@ -14,13 +14,14 @@ ci_mode := if env('CI', '') != '' {'1'} else {''}
...
 export RUSTDOCFLAGS := env('RUSTDOCFLAGS', if ci_mode == '1' {'-D warnings'} else {''})
-export RUST_BACKTRACE := env('RUST_BACKTRACE', if ci_mode == '1' {'1'} else {''})
+export RUST_BACKTRACE := env('RUST_BACKTRACE', if ci_mode == '1' {'1'} else {'0'})
 
 @_default:
     {{just_executable()}} --list
 
 # Run integration tests and save its output as the new expected output
 bless *args:  (cargo-install 'cargo-insta')
+    # rm -rf tests/snapshots-*
```
CommanderStorm, line 24: We want either of the below, or a comment explaining why it is commented out/when it should be added

```suggestion
```


```suggestion
    rm -rf tests/snapshots-*
```

rust/mlt-nom/src/_main.rs.old
```diff
@@ -1,45 +1,31 @@
 #![allow(unused_mut)]
 
+mod errors;
 mod structures;
```
CommanderStorm: What happened here? A .old file

rust/mlt-nom/src/utils.rs
```diff
@@ -55,3 +75,77 @@ pub fn encode_str(data: &mut Vec<u8>, value: &[u8]) {
...
+/// Decode RLE (Run-Length Encoding) data
+/// It serves the same purpose as the `decodeUnsignedRLE` and `decodeRLE` methods in the Java code.
+pub fn decode_rle<T: PrimInt + Debug>(
+    data: &[T],
+    runs: usize,
+    num_rle_values: usize,
+) -> Result<Vec<T>, MltError> {
+    let (run_lens, values) = data.split_at(runs);
+    let mut result = Vec::with_capacity(num_rle_values);
+    for (&run, &val) in run_lens.iter().zip(values.iter()) {
```
CommanderStorm, line 130: Do we need a check that `run_lens` and `values` are the same size?

This path currently cannot be triggered, as `StreamMetadata::decode` does not currently handle `Rle`

rust/mlt-nom/src/utils.rs
```diff
@@ -55,3 +75,77 @@ pub fn encode_str(data: &mut Vec<u8>, value: &[u8]) {
...
+    let mut last1 = T::zero();
+    let mut last2 = T::zero();
+
+    for i in (0..data.len()).step_by(2) {
+        last1 = T::decode(data[i]) + last1;
+        last2 = T::decode(data[i + 1]) + last2;
+        result.push(last1);
+        result.push(last2);
+    }
+
```
CommanderStorm, line 104: 🤔 I wonder if the compiler can still autovectorise this. Let's not check this at this point though ^^

rust/mlt-nom/src/v0x01/decodable.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// Borrow the decoded data if available
+    fn borrow_decoded(&self) -> Option<&Self::DecodedType>;
+
+    fn decode(&mut self) -> Result<&Self::DecodedType, MltError> {
+        if self.is_raw() {
+            // Temporarily replace self with a default value to take ownership of the raw data
+            let raw = self.take_raw().expect("Expected raw data");
+            *self = Self::new_decoded(Self::DecodedType::parse(raw)?);
+        }
+        Ok(self.borrow_decoded().expect("Expected decoded data"))
```
CommanderStorm, line 22: are the expects intentional?

rust/mlt-nom/src/v0x01/enums.rs
```diff
@@ -0,0 +1,166 @@
+use borrowme::borrowme;
+// use borrowme::borrowme;
```
CommanderStorm, line 2:
```suggestion
```

rust/mlt-nom/src/v0x01/enums.rs
```diff
@@ -0,0 +1,166 @@
...
+/// Column type enumeration
+#[derive(Debug, PartialEq, Clone, Copy, TryFromPrimitive)]
+#[repr(u8)]
+pub enum ColumnType {
+    Id = 0,
+    OptId = 1,
+    LongId = 2,
+    OptLongId = 3,
+    Geometry = 4,
+    Bool = 10,
```
CommanderStorm, line 91: We are skipping a few numbers here.
Intentional?

rust/mlt-nom/src/v0x01/enums.rs
```diff
@@ -0,0 +1,166 @@
...
+    Vec3,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum VectorType {
+    Flat,
+    Const,
+    Sequence,
+    // Dictionary,
+    // FsstDictionary,
```
CommanderStorm, line 159: Intentional?

rust/mlt-nom/src/v0x01/geometry.rs
```diff
@@ -0,0 +1,198 @@
...
+//                 VectorType::Const
+//             } else {
+//                 VectorType::Flat
+//             }
+//         }
+//     }
+// }
+
+/// Handle the parsing of the different topology length buffers separate not generic to reduce the
+/// branching and improve the performance
```
CommanderStorm, line 175: could you reword this?
I don't get what is the generic alternative

rust/mlt-nom/src/v0x01/root.rs
```diff
@@ -0,0 +1,251 @@
...
+        (input, typ) = Column::parse(input)?;
+        match typ.typ {
+            Geometry => geometries += 1,
+            Id | OptId | LongId | OptLongId => ids += 1,
+            _ => {}
+        }
+        col_info.push(typ);
+    }
+    if geometries > 1 {
+        return Err(MltError::MultipleGeometryColumn);
```
CommanderStorm: nit: to be consistent

```suggestion
        return Err(MltError::MultipleGeometryColumns);
```

rust/mlt-nom/src/v0x01/stream.rs
```diff
@@ -0,0 +1,328 @@
...
+        let logical1 = LT::parse(val >> 5)?;
+        let logical2 = LT::parse((val >> 2) & 0x7)?;
+        let physical = PT::parse(val & 0x3)?;
+
+        let (input, num_values) = utils::parse_varint::<u32>(input)?;
+        let (input, byte_length) = utils::parse_varint::<u32>(input)?;
+
+        let mut input = input;
+        let val1;
+        let val2;
```
CommanderStorm, line 61: Are num_bits, coordinate_shift better names or is this intentional?

rust/mlt-nom/src/errors.rs
```diff
@@ -0,0 +1,165 @@
...
+    // Other errors
+    #[error("not implemented: {0}")]
+    NotImplemented(&'static str),
+}
+
+impl From<Infallible> for MltError {
+    fn from(_: Infallible) -> Self {
+        unreachable!()
+    }
+}
```
CommanderStorm, line 165: What does this do? 🤔

nyurik: I had to do this for some edge case weird conversion where it was a required `try_from` impl, but the error was infallible, or something like that.  Weird edge case, i may remove it in the future

rust/mlt-nom/src/v0x01/decodable.rs
nyurik: yes, because it gets checked right before.  In other words, if it is "raw", the "take_raw" must succeed - but i need to check it so that I don't take the value unless it is of the right type... I may improve it a bit later, but for now this is good enough to try out the separation of common decoding approach and individual decoding for each struct

rust/mlt-nom/src/v0x01/enums.rs
nyurik: yes, we may have a few more specific geometry types - so skipped a few values on purpose

rust/mlt-nom/src/v0x01/enums.rs
nyurik: yeah, keeping things smaller at first - based on the available sample files

rust/mlt-nom/src/v0x01/geometry.rs
nyurik: i think this was copied from the mlt repo - not sure what it means here either - another thing for later rewording

rust/mlt-nom/src/v0x01/root.rs
nyurik: thx

rust/mlt-nom/src/v0x01/stream.rs
nyurik: intentional - this way i don't have to declare val1 and val2 for each branch, but instead just have two int vars on top

rust/mlt-nom/src/utils.rs
nyurik: i think this is for the future - i didn't want to get too deeply into the decoding part until i get the parsing part resolved

PR #580, opened 2025-10-08
Add release workflow for Java

louwers:
Setup publishing for the Java encoder.

Already published under my namespace. https://central.sonatype.com/artifact/nl.bartlouwers/mlt


.github/scripts/extract-version.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
nyurik: meh, this can be done so much nicer in just, with all the string manipulation functions, and it would work consistently in windows too.

PR #583, opened 2025-10-09
Trim Java dependencies

louwers:


java/build.gradle
```diff
@@ -13,23 +13,26 @@ repositories {
...
     implementation 'no.ecc.vectortile:java-vector-tile:1.3.23'
-    implementation 'org.apache.commons:commons-compress:1.28.0'
+    cliImplementation 'org.apache.commons:commons-compress:1.28.0'
     implementation 'org.apache.orc:orc-core:1.8.1'
-    implementation 'org.imintel:mbtiles4j:1.0.6'
+    cliImplementation 'org.imintel:mbtiles4j:1.0.6'
     implementation 'org.locationtech.jts:jts-core:1.20.0'
-    implementation 'org.slf4j:slf4j-simple:2.0.17'
-    implementation 'org.xerial:sqlite-jdbc:3.50.3.0'
+    cliImplementation 'org.xerial:sqlite-jdbc:3.50.3.0'
```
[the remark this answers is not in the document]
msbarry: I think copilot is partially right here.. I think we want slf4j-api in the library, and slf4j-simple in the cli so that an application that consumes the library can define it's own logging config.

java/build.gradle
```diff
@@ -13,23 +13,27 @@ repositories {
...
     implementation 'com.github.davidmoten:hilbert-curve:0.2.3'
     implementation 'com.google.code.gson:gson:2.13.2'
     implementation 'com.google.guava:guava:33.5.0-jre'
     implementation 'com.google.protobuf:protobuf-java:4.32.1'
-    implementation 'commons-cli:commons-cli:1.10.0'
+    cliImplementation 'commons-cli:commons-cli:1.10.0'
     implementation 'io.github.earcut4j:earcut4j:2.2.2'
     implementation 'io.github.sebasbaumh:mapbox-vector-tile-java:25.1.0'
     implementation 'me.lemire.integercompression:JavaFastPFOR:0.2.1'
     implementation 'no.ecc.vectortile:java-vector-tile:1.3.23'
```
msbarry, line 30: Just curious why do we need 2 vector tile libraries?

java/build.gradle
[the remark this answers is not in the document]
msbarry: Actually it doesn't look like we use slf4j at all in the codebase so it's just for transitive deps... this should be fine just in cli, or if it doesn't need to log anything from those deps than might not be needed at all.

java/build.gradle
louwers: Which one is Planetiler using? Maybe I can get rid of the other.

louwers: Ah the ecc one

msbarry: Planetiler is using a custom one derived from the ecc one.

louwers: I'll address this in a follow up, I think io.github.sebasbaumh:mapbox-vector-tile-java is only used in the benchmarks.

PR #588, opened 2025-10-10
chore(rust): more `decode_componentwise_delta_vec2s` tests

nyurik:
copied some tests from C++


nyurik:
eventually i think snapshots, especially the inline kind, might work

PR #596, opened 2025-10-14
fix: long-id and opt-id are swapped

nyurik:
Fixed incorrect bit setting of LongID vs OptionalID -- the last bit consistently used for optionality - whereas for IDs it was accidentally reversed.

Corrected values:

* ID = 0 (same as before)
* Optional ID = 1 (used to be 2)
* 64bit ID = 2 (used to be 1)
* Optional 64bit ID = 3 (same as before)

This will ensure decoders consistently decode both IDs and properties


java/src/main/java/org/maplibre/mlt/converter/encodings/MltTypeMap.java
```diff
@@ -33,7 +33,7 @@ public static Optional<Integer> encodeColumnType(
         }
       } else if (logicalScalarType != null) {
         if (logicalScalarType == MltTilesetMetadata.LogicalScalarType.ID) {
-          return Optional.of(isNullable ? (hasLongIDs ? 3 : 2) : (hasLongIDs ? 1 : 0));
+          return Optional.of(isNullable ? (hasLongIDs ? 3 : 1) : (hasLongIDs ? 2 : 0));
```
msbarry, line 36: It might make things a little more clear for C++ and java versions to define HAS_LONG_IDS=0b10 and IS_NULLABLE=0b01 then do Optional.of((isNullable ? IS_NULLABLE : 0) | (hasLongIds ? HAS_LONG_IDS : 0)) then below `.setNullable((typeCode & IS_NULLABLE) != 0)` and `.setLongId((typeCode & HAS_LONG_IDS) != 0)`

nyurik: agree, but lets do cleanups separately

PR #597, opened 2025-10-14
feat(rust): implement decode long stream

Weixing-Zhang:
- [ ] Migrate the required functions to new architecture
- [ ] Add le_bytes_to_u64s 
- [ ] Add decode_physical_u64
- [ ] Add decode_logical_u64
- [ ] Add decode_long_stream
- [ ] Add tests

### DIFF
--- a/rust/mlt/src/decoder/integer.rs
+++ b/rust/mlt/src/decoder/integer.rs
@@ -135,17 +135,30 @@ fn le_bytes_to_u32s(tile: &mut TrackedBytes, num_bytes: usize) -> MltResult<Vec<
         });
     }
 
-    let num_bytes = num_bytes / 4;
-    let encoded_u32s = (0..num_bytes)
-        .map(|_| {
-            let b1 = tile.get_u8();
-            let b2 = tile.get_u8();
-            let b3 = tile.get_u8();
-            let b4 = tile.get_u8();
-            u32::from_le_bytes([b1, b2, b3, b4])
-        })
-        .collect();
+    let num_ints = num_bytes / 4;
+    let encoded_u32s = (0..num_ints).map(|_| tile.get_u32_le()).collect();
     Ok(encoded_u32s)
+}
+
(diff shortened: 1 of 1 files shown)


nyurik:
@Weixing-Zhang hi, please make sure to look at the Slack message i sent to the `#maplibre-tile-format`

nyurik:
one big issue with these implementations is that unlike the first iteration (in the `/rust/mlt` dir), here we should focus on what is actually being used - i.e. any code we add should parse more MLT files than before

Weixing-Zhang:
> one big issue with these implementations is that unlike the first iteration (in the `/rust/mlt` dir), here we should focus on what is actually being used - i.e. any code we add should parse more MLT files than before

@nyurik Agreed. I will probably integrate the int decoder into the ID decoding in this PR as well.

nyurik:
@Weixing-Zhang please take a look at the https://github.com/maplibre/maplibre-tile-spec/pull/687 -- i think it will be easier to create PRs against that branch, so that we can see what changes with each PR. Think of it as a work-in-progress branch that we will later squash once we get all decoding done properly

Weixing-Zhang:
@nyurik I can pick the work in the `rust/mlt-nom/src/v01/stream.rs` based on https://github.com/maplibre/maplibre-tile-spec/pull/687.

Perhaps we can close this PR then.

nyurik:
@Weixing-Zhang ok, i just did some vibe coding with #687 - seems now all decoded snapshots are generated, but i suspect they are far from being in a "good state". I am beginning to think we should just implement the MLT -> MVT encoding, and do the comparison with the original data. This way we can let AI deal with the bugs, and just make it ensure generated MVTs are exactly as the original data...  In the mean time - please do a deep dive into the decoders - you have far better understanding of how they are being done than I

Weixing-Zhang:
The implementation is being done in this PR #692 now.

PR #603, opened 2025-10-15
Refactor stream capture into an observer interface

TimSylvester:
Also fix various warnings and implement IDE style suggestions


TimSylvester:
@nyurik This changes the naming of the raw stream files, hopefully an easy update.

Along with adding the layer names, I also restored the "id" in streams for ID columns.

There's a further issue that we end up writing the streams from multiple variations that are being tried, but then only use one of those in the output, so the stream files don't necessarily represent the MLT output.  But that's a separate issue.

nyurik:
Don't worry about test filenames, not a biggie. Why do you want to add the word 'id' to the mlt for the built in columns?

TimSylvester:
Not in the MLT, in the stream filenames, it was coming out, e.g., `prop__ints` instead of `prop_id_ints` since they don't have name strings anymore.

PR #605, opened 2025-10-15
Update npm dependencies

HarelM:
This breaks lint, but I'll fix it in another PR as it's getting too long to review.

It is needed as maplibre-gl uses a later version of typescript where this array view change is important, so this is needed in order to maplibre-gl typescript transpilation.


louwers:
You will fix the commented out code in a follow-up?

PR #607, opened 2025-10-15
Fix path Rust test metadata

louwers:


nyurik:
seems the parsing code is no longer matching the tile format. I disabled the test, but i think we should start transitioning all the good low-level decoding to the new mlt-nom crate which does the top level parsing a bit cleaner, but lacks the decoding aspects.

CC: @Weixing-Zhang @jjcfrancisco

PR #610, opened 2025-10-16
chore(rust): switch to validating mlt-nom variant

nyurik:
The `rust/mlt` branch is no longer capable of parsing current data format, and should be migrated to the newer `rust/mlt-nom` implementation. For now, I renamed `rust.yml` into `rust-legacy.yml`, which at some point will be deleted.


CommanderStorm (approved):
Fair, but we might just delete the old code instead of keeping it around, right?

I am somewhat failing to see the value otherwise

nyurik:
Correct, we should merge two efforts

PR #612, opened 2025-10-16
chore(js): Update lint, add to CI

HarelM:
This adds lint support for ts folder and adds build and lint to CI.


.github/workflows/test.yml
```diff
@@ -11,6 +11,17 @@ defaults:
     shell: bash
 
 jobs:
+  js-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: ./.github/actions/mlt-setup-node
+      - run: |
+          cd ts
```
louwers: Since we're using a justfile in this repo, I would add it there instead and call `just something`

ts/webpack.config.js
```diff
@@ -1,38 +0,0 @@
-const { CleanWebpackPlugin } = require("clean-webpack-plugin");
```
louwers, line 1: This package is still in `package.json`.

louwers (changes requested):
See comments.

Also why is there both a `ts` and `js` directory?

Personally I think it makes more sense to settle on `js`, but I don't hold that opinion strongly.

ts/tsconfig.json
```diff
@@ -14,8 +14,7 @@
     "forceConsistentCasingInFileNames": true
   },
   "include": [
-    "src/**/*.ts",
-    "test/**/*.ts"
+    "src/**/*.ts"
```
louwers, line 17: Might want to add

```
  "module": "ESNext",
  "strict": true,
  "noUnusedParameters": true,
  "erasableSyntaxOnly": true,
  "noFallthroughCasesInSwitch": true,
  "noUncheckedSideEffectImports": true,

```

.github/workflows/test.yml
HarelM: I looked at it and was puzzled as to why we are using it for all the libraries to be honest.
Is a single person planning to work on all the different languages?
I honestly don't see value, and needless to say that it requires some extra setup in the CI, which is also "a smell".
I'll add that there, but it will be unmaintained eventually, like it is now (even before my changes)...

ts/webpack.config.js
HarelM: Thanks! I missed that.

HarelM:
> Also why is there both a ts and js directory?

Should be a single one, I'm waiting for @Turtelll to bring back the tests and afterward we will remove the extra folder...

ts/tsconfig.json
HarelM: I want to have linting for all files and tests first and then I'll move to strict mode.
All in good time...

.github/workflows/test.yml
HarelM: Nevertheless, I've updated the just file...

HarelM:
I've removed the js folder for now, we could rename ts folder to js once things settle down a bit, but I don't want to do it right now.

PR #617, opened 2025-10-16
Apply prettier format

HarelM:
This is fully automated.
Is there a way to add this to pre-commit instead of the need to manually run this?
@Turtelll - note that this may cause conflicts with your changes, so make sure you update the code from version 0.0.1-alpha.7 and then apply formatting to ease the conflict resolution.


louwers:
Prettier is not compatible with pre-commit anymore unfortunately.

PR #618, opened 2025-10-16
Remove fixture files from test gitignore …

TimSylvester:
…so that unexpected changes are not hidden
Remove obsolete stuff from test gitignore
Remove unused tests and benchmark directory
Rename test class to resolve style issue

### DIFF
--- a/rust/mlt/src/decoder/mod.rs
+++ b/rust/mlt/src/decoder/mod.rs
@@ -44,7 +44,7 @@ mod tests {
             .collect::<Vec<_>>()
     }
 
-    #[ignore] // not updated for tag0x01 yet
+    #[ignore = "not updated for tag0x01 yet"]
     #[test]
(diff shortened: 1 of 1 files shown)


rust/mlt/src/decoder/mod.rs
```diff
@@ -44,6 +44,7 @@ mod tests {
             .collect::<Vec<_>>()
     }
 
+    #[ignore] // not updated for tag0x01 yet
```
nyurik:
```suggestion
    #[ignore = "not updated for tag0x01 yet"]
```

HarelM:
@TimSylvester I'm not following which rust is maintained?
Can I remove the rust-legacy.yml and have the rust CI trigger only on changes to rust folder?

TimSylvester:
I don't know

PR #619, opened 2025-10-16
Add Column Mapping to CLI Encoder

TimSylvester:
Allows the existing column mapping functionality to be used from the command line for single-tile or MBTiles conversion.

Multiple mappings can be specified, but they are all applied to all layers.

Also fixes an issue where the separator was ignored and ':' was always used.  Note that the specified separator is not stored, so it must be separately provided to the decoder to recover the same property names.

Refactors the argument parsing and try/catch block out of main.

### DIFF
--- a/java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/StringEncoder.java
+++ b/java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/StringEncoder.java
@@ -81,7 +81,7 @@ public class StringEncoder {
             ? encodedSharedFsstDictionary
             : encodedSharedDictionary;
 
-    final var width = (int) Math.ceil(Math.log10(dataStreams.size() - 1));
+    final var width = (int) Math.max(1, Math.ceil(Math.log10(dataStreams.size() - 1)));
     for (var i = 0; i < dataStreams.size(); i++) {
       var presentStream = presentStreams.get(i);
       var dataStream = dataStreams.get(i);


nyurik:
I think that a set of fields with a shared dictionary should not optimize field names, e.g. we have `name` and `name:en` share a dictionary, the layout should be like this example. The reason for this is that we must NOT break the existing concept of roundtripability - MVT->MLT->MVT should not loose exact field name.

```yaml
{
  type: struct,  # indicate we have a nested structure
  name:  NULL,   # skipped - names are pushed one level down
  data: Stream,  # one or more shared streams for the shared data
  values: [
    {
      type: string,
      name: "name",
      data: Stream,  # references into the shared data stream in the above level
    },
    {
      type: string,
      name: "name:en",
      data: Stream,  # references into the shared data stream in the above level
    },
}
```

TimSylvester:
If we store the original separator, the decoder can reconstruct whatever the source field names were.  The obvious place, without breaking anything, is to append the separator to the root name.
@mactrem Any reason not to do that?

nyurik:
sure, that would work too

nyurik:
note that the algorithm for computing root name should be "prefix of any size that is common to all names" -- so in my example above, the separator would be prepended to the 2nd value, not included in the root:

```yaml
{
  name: "name",
  values: [
    { name: "" } // empty string
    { name: ":en" } // english suffix
  ]
}
```

In case the names are totally different, the root name will be an empty `""`

nyurik:
also, note that there IS a downside to this -- splitting name into two places means that the decoder will have to allocate a string for each column, or at least allocate one for each struct case -- otherwise it won't be able to easily compare them. Without it, decoder can simply store a reference to the name in the original input without copying it

nyurik (approved):

mactrem:
> If we store the original separator, the decoder can reconstruct whatever the source field names were. The obvious place, without breaking anything, is to append the separator to the root name. @mactrem Any reason not to do that?

yes makes sense to store the seperator to reconstruct the original column names because for example the MVT styles can be reused and need no adaption. Back then i just took the shortcut and wanted to fix that later ...

mactrem:
Additionally, since we don’t yet have automatic detection of correlations between columns, and therefore rely on explicit configuration to decide when to apply shared dictionary encoding, we could consider a heuristic approach that detects correlated columns based on their names (for example, by identifying patterns like name:* or name_*). Otherwise, we’ll face the same issue in future tilesets that haven’t been explicitly fine-tuned

TimSylvester:
Assuming that arbitrary columns can be specified or selected, the root would be the longest common prefix of the values, right?

For `name`, `name:en`, the root would be `name` and the children blank and `:en`.

For `name:en` and `name:int`, the root would be `name:`.

For `name` and `label`, it would be blank.

In any case, the loop which establishes the types needs to be made more general, and I need to understand it better first, e.g., the use of "default".

PR #626, opened 2025-10-18
feat(rust): more decoding work

nyurik:
Simplify how decoding is done, focusing it just on the level that makes sense to be both in encoded and decoded state.

I added a new test to the `tests/snapshot.rs` file - uncomment the section at the end for decoding (i can't use `#[ignore]` with `test_each_path`


rust/mlt-nom/src/v01/id.rs
```diff
@@ -60,6 +60,16 @@ impl<'a> FromRaw<'a> for DecodedId {
     type Input = RawId<'a>;
 
     fn from_raw(RawId { optional, value }: RawId<'_>) -> Result<Self, MltError> {
-        todo!()
+        let decoded = match value {
```
CommanderStorm: Removing the panic when below code is still in the stubbed state does not seem to be a smart move, or is this necessary?

nyurik: thx, fixed

PR #628, opened 2025-10-19
chore(js): use vitest instead of jest

HarelM:
This moves the tests to use vitest instead of jest.
It moves the test files next to the code as is considered best practice for typescript/javascript.
It adds the tests to the CI.

Currently, since decode tests fail I added them with "skip" so that they can be enabled in a different PR.

### DIFF
--- a/ts/src/mltDecoder.spec.ts
+++ b/ts/src/mltDecoder.spec.ts
@@ -18,7 +18,7 @@ function testTiles(mltSearchDir: string, mvtSearchDir: string, isSorted = false,
     let mltFileNames = readdirSync(mltSearchDir)
         .filter((file) => parse(file).ext === ".mlt")
         .map((file) => parse(file).name);
-    mltFileNames = [mltFileNames[0]]; // TODO: remove this!
+    mltFileNames = [mltFileNames[0]]; // TODO: remove this once decoder is able to handle all tiles
     for (const fileName of mltFileNames) {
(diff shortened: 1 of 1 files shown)


HarelM:
@Turtelll please note that the decoder fails for the first tile I checked - 0_0_0.mlt inside the tag0x01 folder.
I'm not sure I understand why it fails and if there's a problem with the decoder or the tiles.
Once you get a chance to check it let me know.

louwers (commented):
This deletes a some existing tests such as the earcut tests. Is that intentional?

Node.js also has a built-in test runner now. Might be a good option to avoid adding any test dependencies. https://nodejs.org/api/test.html

But I understand if you want to standardize on Vitest.
But

HarelM:
I deleted the `earcut.ts` file since there's a library for it, and moved to use that.
I deleted most of the mltDecoder tests as the folders mentioned there are not part of this repo at this point, so there's no point in keeping these tests.

I've briefly looked at the node.js test docs and it seems that code coverage is still experimental there and I'm not sure how well this support typescript, so I'll let them stabilize things a bit (probably in like 10 versions) and then decide if it's worth the trouble... Thanks for the reference though!

birkskyum:
White nodejs just deletes the types, and effectively test the vanilla js code, Vitest can actually test types if we'd like - https://vitest.dev/guide/testing-types

I think for a Spec that can be quite handy

PR #629, opened 2025-10-19
chore(js): remove protobuf dependency

HarelM:
This removes the dependency for protobuf and simply uses a typescript definition of the proto file.
The definitions were generated using @louwers's pbf implementation, but since all the decoding code is written in this code, there's no need for an extra dependecy, so I removed the package and kept the typescript file.
It might be worth trying to use pbf decoding capabilities instead of coding it here, but I didn't want to create a large PR without proper testing to make sure that what is being parsed did not change.

The main goal of this PR is to reduce bundle size for maplibre-gl-js.
I want to see how big the code gets without (almost) any external dependecy.

### DIFF
--- a/ts/package-lock.json
+++ b/ts/package-lock.json
@@ -103,6 +103,74 @@
         "node": ">=18"
       }
     },
+    "node_modules/@esbuild/aix-ppc64": {
+      "version": "0.25.11",
+      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.25.11.tgz",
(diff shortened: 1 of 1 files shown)


birkskyum (approved):

HarelM:
I still don't have a good way to test this change, so I'll keep this PR open until I do...

HarelM:
Test were added in a separate PR and seems to continue passing, it doesn't mean too much as code coverage is at about 30%, but it does seem to allow parsing the tiles that are currently part of 0x01 tag, besides a few tests that are ignored until fixed (next PR hopefully).

PR #631, opened 2025-10-20
chore(js): use simple folder for js tests

HarelM:
Remove skip, remove using only the first test.
Add code coverage
Update node to version 22.13


HarelM:
cc: @Turtelll @mactrem 
Note that it might be related to the fact that there is only one feature in the tile and thus the geometry vector type is const and getting into the GpuVector code path, which might not be well implemented.
IDK, I don't fully understand the code.
Other tests are failing when trying to use something called `mortonSettings` which might not be fully implemented as well.
Please let me know what you find out.

HarelM:
Replaced by #632

PR #632, opened 2025-10-20
Use simple MLT files for testing

Turtelll:
- Add feature access for tessellated geometries 
- Simple and Amazon pass tests now
- OMT tiles still have an error with ID 
- Add code coverage
- Update node to version 22.13


HarelM:
I think the iterator implementation should still be added, but this can be done after this PR is merged.

PR #635, opened 2025-10-21
feat(rust): add decode int stream for id mlt-nom

Weixing-Zhang:
- [x] Implement decode_int_stream using the zero-copy design
- [x] Migrate and copied the required functions for only the existing testing cases

### DIFF
--- a/java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
+++ b/java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
@@ -35,6 +35,7 @@ import java.util.stream.Stream;
 import java.util.zip.Inflater;
 import java.util.zip.InflaterInputStream;
 import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.Converter;
 import org.apache.commons.cli.DefaultParser;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
@@ -63,6 +64,8 @@ import org.maplibre.mlt.converter.MLTStreamObserver;
 import org.maplibre.mlt.converter.MLTStreamObserverDefault;
 import org.maplibre.mlt.converter.MLTStreamObserverFile;
 import org.maplibre.mlt.converter.MltConverter;
+import org.maplibre.mlt.converter.encodings.fsst.FsstEncoder;
+import org.maplibre.mlt.converter.encodings.fsst.FsstJni;
 import org.maplibre.mlt.converter.mvt.ColumnMapping;
 import org.maplibre.mlt.converter.mvt.MapboxVectorTile;
 import org.maplibre.mlt.converter.mvt.MvtUtils;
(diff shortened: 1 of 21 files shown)


Weixing-Zhang:
@nyurik Do the placeholders look correct to you?

rust/mlt-nom/src/decoder/integer.rs
```diff
@@ -0,0 +1,18 @@
+use crate::v01::{Stream, StreamMeta};
+use crate::{MltError, MltRefResult};
+
+pub fn decode_int_stream(
+    input: &[u8],
+    metadata: &StreamMeta,
+    is_signed: bool,
+) -> MltRefResult<'_, Vec<i32>> {
```
nyurik: there is no point to use `MltRefResult` here -- you are not returning a reference to the original data - that's what Stream creation code does. Your code here converts `Stream` instances into owned `Vec<T>`.  So just use `Result<Vec<i32>, MltError>`. Same elsewhere

rust/mlt-nom/src/decoder/integer.rs
```diff
@@ -0,0 +1,18 @@
+use crate::v01::{Stream, StreamMeta};
+use crate::{MltError, MltRefResult};
+
+pub fn decode_int_stream(
+    input: &[u8],
```
nyurik: i think you should just accept a reference to a stream instance as input here

nyurik:
Before implementing all this, please take a look at how https://github.com/maplibre/maplibre-tile-spec/blob/9aab54864c5dc10eb6af88c6bf313c6850574890/rust/mlt-nom/src/v01/geometry.rs#L133 is done

Weixing-Zhang:
@nyurik Could you please review?

rust/mlt-nom/src/utils.rs
```diff
@@ -150,6 +150,30 @@ impl<T> SetOptionOnce<T> for Option<T> {
     }
 }
 
+/// Decode a slice of bytes into a vector of u32 values assuming little-endian encoding
+pub fn bytes_to_u32s(mut input: &[u8], num_values: u32) -> MltRefResult<'_, Vec<u32>> {
```
nyurik, line 154: @Weixing-Zhang i am not sure if you saw my feedback - why do you use `MltRefResult` here?  You don't need to return `input` because you are creating a new Vec. I can clean it up i guess

Weixing-Zhang: Returning the ref allows us to (1) verify if all the data is consumed (we can do it internally),  (2) match the nom stype, and more importantly (3) allow us to return the leftover ref if needed.

nyurik (approved):

HarelM:
I think the name Test for the rust CI is a bit too generic, can you update it to match the other names?

nyurik:
@HarelM i'm not sure what you were referring to. CI was not updated in this PR

HarelM:
It wasn't, it was changed in a previous PR by me, and I'm not sure I did the right thing in regards to Rust, so I'm looking for feedback for the people who maintain rust in this repo...

PR #637, opened 2025-10-21
Fix faulty ID decoding 

Turtelll:
This PR fixes faulty ID decoding in some tiles from the "omt" folder are decoded incorrectly and dont pass the ID comparisson with MVT.

### DIFF
--- a/ts/src/encodings/integerStreamDecoder.ts
+++ b/ts/src/encodings/integerStreamDecoder.ts
@@ -181,10 +181,7 @@ export default class IntegerStreamDecoder {
             case LogicalLevelTechnique.DELTA:
                 if (streamMetadata.logicalLevelTechnique2 === LogicalLevelTechnique.RLE) {
                     const rleMetadata = streamMetadata as RleEncodedStreamMetadata;
-                    if (rleMetadata.runs > 1) {
-                        return decodeDeltaRle(values, rleMetadata.runs, rleMetadata.numRleValues);
-                    }
-                    values = decodeUnsignedRle(values, rleMetadata.runs, rleMetadata.numRleValues);
+                    return decodeDeltaRle(values, rleMetadata.runs, rleMetadata.numRleValues);
                 }
                 decodeZigZagDelta(values);
                 return values;
@@ -222,10 +219,7 @@ export default class IntegerStreamDecoder {
             case LogicalLevelTechnique.DELTA:
                 if (streamMetadata.logicalLevelTechnique2 === LogicalLevelTechnique.RLE) {
                     const rleMetadata = streamMetadata as RleEncodedStreamMetadata;
-                    if (rleMetadata.runs > 1) {
-                        return decodeDeltaRleInt64(values, rleMetadata.runs, rleMetadata.numRleValues);
-                    }
-                    values = decodeUnsignedRleInt64(values, rleMetadata.runs, rleMetadata.numRleValues);
+                    return decodeDeltaRleInt64(values, rleMetadata.runs, rleMetadata.numRleValues);
                 }
                 decodeZigZagDeltaInt64(values);
                 return values;


Turtelll:
The TypeScript MLT decoder was incorrectly decoding feature IDs when Delta-RLE encoding had multiple runs.

HarelM:
Thanks! Make sure the tests (with the "different" tiles) are really covering the code you changed, or add some other tests.

Turtelll:
When comparing the IDs from mlt and mvt i still have issues in some tiles. For example, when i decode omt/10_530_683.mlt i get these IDs in the waterway layer:
- mlt ID = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12
- mvt ID = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11

This stream is DeltaRLE encoded. So the stream prior to decoding should be: 
- [11, 2]
- 2 after ZigZag = 1
- 11 runs of delta value 1

What i actually get for the DeltaRle stream is:
- [10, 1, 2, 4]
- 2, 4 are the deltas -> after ZigZag 1, 2
- 10 runs of delta value 1
- 1 run of delta value 2 
- resulting in ID = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12 ]

Is this a bug in the encoder or am I understanding something wrong? Please help me out

HarelM:
@TimSylvester can you take a look? 
I saw some skipped tests in the cpp omt test suite, are the tiles there valid?

TimSylvester:
I'm not sure, I'm having trouble even building the cpp targets from main now.

I have been able to compare all the fixture files to the original after Java encode/decode, so if there's a bug there it must be matched by the decoder and preserve the data.  Or, I suppose, the compare could be broken.

There have been some issues with incorrect test files, in part due to incorrect git ignores.  You may want to reset the test directory to make sure. 

I see that the `generate-expected-mlt` is using `--tessellate`, unfortunately that precludes the use of `--compare-all` for now, as the Java decoder doesn't handle the triangles.  You may want to tweak `justfile` to replace the former with the latter as a sanity check.

HarelM:
I meant these skipped tiles:
https://github.com/maplibre/maplibre-tile-spec/blob/bf4f4e736b6f9465b0d2704ebb652a1ecec36edc/cpp/test/test_decode.cpp#L165-L173

TimSylvester:
As I suspected, those excludes are obsolete, and I'm able to decode everything after removing it. (#639)

However, if I `just generate-expected-mlt`, it no longer passes, so something is wrong there.

HarelM:
Ok, thanks for the info.

TimSylvester:
Aha, `justfile` wasn't updated when the Java code was split, and is running the Java encoder from its previous location, if at all.

nyurik:
note that currently we have more than one justfile (they should eventually be merged i think). Also, just searches for the first available justfile going up the tree - so it can be ran from subdirs

Turtelll:
I messed up with the MVT ID it is actually [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12 ]. But there was a small bug in the decoder nevertheless. 

The encoding implementation creates two deltaRLE runs when starting ID value is not 1
e.g. [ 2, 3, 4, 5, 6 ] gets encoded with two runs to [ 1, 4, 4, 2 ]. 
- 1 run of ZigZag = 4 -> 2 (starting point)
- 2 runns of ZigZag =2 -> 1 (delta value)

That is why run == 1 and run == 2 were detected as a sequence in the ts decoder. But in my case the Tile had two runs but wasn't a sequence
ToDo: check condition if two runs are still a sequence or not

HarelM:
Can you maybe add a unit test for the 64 part to cover this case?
Or if this is not needed currently and we don't have any use cases maybe we can introduce this part of the code when it's relevant in a later commit?

ts/src/mltDecoder.spec.ts
```diff
@@ -21,14 +21,8 @@ describe("MLT Decoder - MVT comparison for Amazon tiles", () => {
...
-    // TODO: remove ignore list when the issues with the ID of these tiles are resolved
-    testTiles(omtMltTileDir, omtMvtTileDir, [
-        "10_530_683",
-        "10_532_683",
-        "11_1062_1367",
-        "11_1064_1366",
-        "13_4267_5467",
-    ]);
+    // TODO: handle case where ID is Delta-RLE encoded and sequential
+    testTiles(omtMltTileDir, omtMvtTileDir);
```
HarelM, line 25: I think it's worth removing the ignore list as a parameter if it is not in use.
We can always introduce it in the future if this is needed.
Although I'm against skipped tests, so only in very rare cases we'll add this back.

ts/src/encodings/integerStreamDecoder.ts
```diff
@@ -179,6 +181,9 @@ export default class IntegerStreamDecoder {
             case LogicalLevelTechnique.DELTA:
                 if (streamMetadata.logicalLevelTechnique2 === LogicalLevelTechnique.RLE) {
                     const rleMetadata = streamMetadata as RleEncodedStreamMetadata;
+                    if (rleMetadata.runs > 1) {
```
HarelM: From a code perspective - I would expect this part to be something like:
`return rleMetadata.runs > 1 ? decodeDeltaRle(values, rleMetadata.runs, rleMetadata.numRleValues) : decodeUnsignedRle(values, rleMetadata.runs, rleMetadata.numRleValues);

Meaning, `decodeDeltaRle` has `zigzag` "inside" while `decodeUnsignedRle` has `zigzag` "outside".
This isn't a good practice as both methods should behave the same, I believe.

ts/src/mltDecoder.spec.ts
```diff
@@ -21,14 +21,8 @@ describe("MLT Decoder - MVT comparison for Amazon tiles", () => {
...
     const omtMvtTileDir = "../test/fixtures/omt";
-    // TODO: remove ignore list when the issues with the ID of these tiles are resolved
-    testTiles(omtMltTileDir, omtMvtTileDir, [
-        "10_530_683",
-        "10_532_683",
-        "11_1062_1367",
-        "11_1064_1366",
-        "13_4267_5467",
-    ]);
+    // TODO: handle case where ID is Delta-RLE encoded and sequential
```
HarelM, line 24: Can we add tests for this now? Or maybe in a next PR?
If not, I would suggest to create an issue and add all the information there to support this in the future instead of keeping this TODO.

ts/src/mltDecoder.spec.ts
Turtelll: i will remove it for now

ts/src/mltDecoder.spec.ts
Turtelll: Will be my next PR

HarelM: Ok, great.

ts/src/encodings/integerStreamDecoder.ts
Turtelll: Cleaned it up. We don't need to check runs anymore as it's handled in getVectorType and decodeDeltaRle handles the ZigZag internally.

HarelM:
Do we have a use case for the usage of: `decodeDeltaRleInt64`?
I see that it is not covered by the current test according to the code review report?
I think this can be merged as is for now, but we need to improve coverage...

PR #643, opened 2025-10-23
Get correct vectortype

Turtelll:
With this PR getVectorType returns the correct Vectortype when the stream is Delta-Rle encoded

### DIFF
--- a/ts/src/encodings/integerStreamDecoder.spec.ts
+++ b/ts/src/encodings/integerStreamDecoder.spec.ts
@@ -1,0 +1,60 @@
+import {describe, it, expect} from "vitest";
+import IntegerStreamDecoder from "./integerStreamDecoder";
+import {RleEncodedStreamMetadata} from "../metadata/tile/rleEncodedStreamMetadata";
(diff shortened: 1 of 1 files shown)


Turtelll:
The Delta-Rle encoding is implemented in a way that allows for a sequence to exist with more than two delta runs. 
> - [ 1, 2, 3, 4 ] is an obvious sequence and encoded as [ 4, 2 ]           // [ repetition, delta (zigZag encoded 2 = 1, 4 = 2) ]

But when the sequence starts with an ID other than '1' the first delta run is used recreate the first ID value.
> - [2, 3, 4, 5, 6] will be encoded as [1, 4, 4, 2] one run of delta 2 and then 4 runs of delta 1

HarelM:
Before and after this fix all the tests pass, do we need to add a unit test here or something similar for the cases that are not covered with the tiles?
I see some parts in the integerStreamDecoder that are not covered by the tiles tests, right?

Turtelll:
getVectorType is fully covered. I will make sure integerStreamDecoder  is covered in a seperate PR

HarelM (approved):
Good work!

PR #644, opened 2025-10-23
Enforce layer feature-count parity and iterate comparison by MVT

### LINKED ISSUE #640: Typescript decoder fails to decode geometries for some features
I was testing planetiler MLT output from https://github.com/onthegomap/planetiler/pull/1355 against @HarelM's  PR https://github.com/maplibre/maplibre-gl-js/pull/6570 and noticed some weird ocean artifacts:

Image

Image

It seems that the ts decoder is failing to decode some geometries.  Here are 2 tiles: 110.mlt looks right 100.mlt does not look right:

[Archive.zip](https://github.com/user-attachments/files/23051427/Archive.zip)

Salkin975:
- Fixes #640

Enforce strict parity between MVT and MLT layers by asserting equal feature counts, iterate over MVT features to ensure every MVT feature is compared, and remove unused imports.

### DIFF
--- a/ts/package-lock.json
+++ b/ts/package-lock.json
@@ -1516,14 +1516,14 @@
       }
     },
     "node_modules/@vitest/coverage-v8": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/@vitest/coverage-v8/-/coverage-v8-4.0.1.tgz",
-      "integrity": "sha512-nmB+UVryiWQLC0pfPQ6KmJacew1ecpuKeUyiGbXtp1+KoYtCTAAlLI++8X/wJfzlULil+l/1jiWPreFnB1U5Mg==",
+      "version": "4.0.2",
+      "resolved": "https://registry.npmjs.org/@vitest/coverage-v8/-/coverage-v8-4.0.2.tgz",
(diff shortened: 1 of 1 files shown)


HarelM:
Nice! Now all that's left is to solve the problem in order to make the tests pass :-)

ts/src/mltDecoder.spec.ts
```diff
@@ -64,9 +64,11 @@ function comparePlainGeometryEncodedTile(
         // Use getFeatures() instead of iterator (like C++ and Java implementations)
         const mltFeatures = featureTable.getFeatures();
 
-        for (let j = 0; j < mltFeatures.length; j++) {
-            const mltFeature = mltFeatures[j];
+        expect(layer.length).toEqual(mltFeatures.length);
```
HarelM:
```suggestion
        expect(layer.length).toBe(mltFeatures.length);
```

Turtelll:
StreamMetadata can be RLE encoded. The numValues function of StreamMetadataDecoder returned the values of the physical stream of StreamMetadata straight without decoding. So if the StreamMetadata is encoded we got the wrong amount of Features and apparently more than halve of the StreamMetadata was RLE encoded.

ts/package-lock.json
HarelM, line 1: I think you should revert this so the CI could run...

Turtelll: yep was by accident

PR #646, opened 2025-10-23
More work on column mapping in Java encoder

TimSylvester:
- Rework struct columns to store separators
- Add support for explicit column maps
- Enable some disabled tests, add more


nyurik (approved):
seems ok (although @msbarry might have a more in-depth ideas?).  Also, would it make sense to simply use a prefix?  E.g. `--group-strings-with-prefix name`  would combine all columns that begin with `name` AND have a string type into a single group?  Or same with a regex?  And multiple params would create multiple groups?  Regex seems a bit more versatile i guess: `--group-strings-by '^name[:_]'`

TimSylvester:
Yes, that last works currently, though the start anchor is implied.

PR #653, opened 2025-10-27
chore(js): add unit tests to vector classes

Turtelll:
This PR introduces tests for vector types that have no coverage from tile decoding tests yet.

### DIFF
--- a/java/mlt-core/src/main/java/org/maplibre/mlt/converter/MltConverter.java
+++ b/java/mlt-core/src/main/java/org/maplibre/mlt/converter/MltConverter.java
@@ -144,7 +144,17 @@ public class MltConverter {
         if (previousScalarType.hasPhysicalType()) {
           final var previousPhysicalType = previousScalarType.getPhysicalType();
           if (previousPhysicalType != scalarType) {
-            if (enableCoerceOnMismatch) {
+            if (previousPhysicalType == MltTilesetMetadata.ScalarType.INT_32
+                && scalarType == MltTilesetMetadata.ScalarType.INT_64) {
+              // Allow implicit upgrade from INT_32 to INT_64
+              columnSchemas.put(
+                  mvtPropertyName,
+                  createScalarColumnScheme(
+                      mvtPropertyName, true, MltTilesetMetadata.ScalarType.INT_64));
+            } else if (previousPhysicalType == MltTilesetMetadata.ScalarType.INT_64
+                && scalarType == MltTilesetMetadata.ScalarType.INT_32) {
+              // keep INT_64
+            } else if (enableCoerceOnMismatch) {
               if (previousPhysicalType != MltTilesetMetadata.ScalarType.STRING) {
                 columnSchemas.put(
                     mvtPropertyName,
@@ -648,8 +658,12 @@ public class MltConverter {
     }
     // TODO: also handle unsigned int to avoid zigZag coding
     // TODO: quick and dirty fix for wrong data types -> make proper solution
-    else if (propertyValue instanceof Integer || propertyValue instanceof Long) {
+    else if (propertyValue instanceof Integer) {
       return MltTilesetMetadata.ScalarType.INT_32;
+    } else if (propertyValue instanceof Long) {
+      return ((long) propertyValue > Integer.MAX_VALUE || (long) propertyValue < Integer.MIN_VALUE)
+          ? MltTilesetMetadata.ScalarType.INT_64
+          : MltTilesetMetadata.ScalarType.INT_32;
     }
     // TODO: also handle unsigned long to avoid zigZag coding
     /*else if (propertyValue instanceof Long) {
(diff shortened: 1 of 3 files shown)


ts/src/vector/constant/intConstVectors.spec.ts
HarelM, line 1: Please split this file into two files, one for each type of vector, you can add test utils if you need a method the exists in both files.

ts/src/vector/constant/constVectors.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        expect(vec.match([10, 99]).limit).toBe(0);
+    });
+
+    it("should filter selected", () => {
+        const vec = new IntConstVector("test", 42, 10);
+        const selection = new FlatSelectionVector([2, 3, 4, 5]);
+        vec.filterSelected(42, selection);
+        expect(getSelectionIndices(selection)).toEqual([2, 3, 4, 5]);
+
+        const selection2 = new FlatSelectionVector([2, 3, 4]);
```
HarelM: Is this a different test? I would consider creating a different test for it if this tests something a bit different.
A good practice is a single `expect` in a unit test, I break this rule if I test different parts on the same object, but otherwise it's better to have a different test.

ts/src/vector/flat/intFlatVector.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+function getSelectionIndices(selection: { selectionValues: () => number[]; limit: number }): number[] {
+    return selection.selectionValues().slice(0, selection.limit);
+}
+
+const simpleVector = createVector([10, 20, 30, 40, 50]);
+const withDuplicates = createVector([10, 20, 30, 20, 50, 10]);
+const withNulls = createNullableVector([10, 20, 30, 40, 50], 0b00010111);
+
+describe("IntFlatVector", () => {
+    describe("constructor and getValue", () => {
```
HarelM: The constructor is executed outside this `describe`, so I'm not sure this describes well the tests in this group...

ts/src/vector/flat/intFlatVector.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+        it("should create vector and get values", () => {
```
HarelM: This test doesn't really create the vector as it is done once for all the tests...

ts/src/vector/flat/numericFlatVectors.spec.ts
HarelM, line 1: It is better to split this file as well for float and double...

ts/src/vector/sequence/intSequenceVectors.spec.ts
HarelM, line 1: Same here - split the file.

HarelM: Or simply rename, as I don't see tests for long sequence vector.

ts/src/vector/flat/numericFlatVectors.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+        expect(getSelectionIndices(sel2)).toEqual([0, 1]);
+    });
+
+    it("should handle nullability", () => {
+        const nullVec = createNullableVector([1.5, 2.5, 3.5, 4.5], 0b0111);
+        expect(getSelectionIndices(nullVec.filter(4.5))).toEqual([]);
+        expect(getSelectionIndices(nullVec.greaterThanOrEqualTo(3.5))).toEqual([2]);
+    });
+
+    it("should throw for not implemented methods", () => {
```
HarelM: Do we want to keep those non implemented methods?

HarelM:
Check out the recent codecov coverage report, this will tell you which tests are still missing.
I would be hesitate to use GenAI here since it might generate too many tests, and some might not be needed.
So I would recommend using it with caution and making sure you only add tests that improve coverage.
What I usually do is run the coverage report locally.
Add a test to cover some functionality.
And check that the coverage went up.

image

The tests are really good, thanks!

ts/src/vector/filter/selectionVectorUtils.ts
```diff
@@ -13,7 +13,7 @@ export function createSelectionVector(size: number) {
 export function createNullableSelectionVector(size: number, nullabilityBuffer: BitVector) {
     const selectionVector = [];
     for (let i = 0; i < size; i++) {
-        if (nullabilityBuffer || nullabilityBuffer.get(i)) {
+        if (nullabilityBuffer && nullabilityBuffer.get(i)) {
```
HarelM: Use "?." operator?

ts/src/vector/sequence/intSequenceVectors.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    });
+
+    it("should throw for not implemented methods", () => {
+        const vec = new IntSequenceVector("test", 10, 5, 5);
+        expect(() => vec.match([10, 15])).toThrow("Not implemented yet.");
+        expect(() => vec.greaterThanOrEqualTo(15)).toThrow("Not implemented yet.");
+    });
+});
+
+describe("LongSequenceVector", () => {
```
HarelM: This should be a part of a different file...

HarelM:
The coverage percentage improvement is great, but we should aim for more, especially for the last files in the following list (where we have less than 80%):
image

If you prefer to split it into separate PRs that's fine too, whatever works for you, just let me know.
Besides the last two comments, this can be merged.

PR #659, opened 2025-10-29
chore(js): Improve code coverage

Turtelll:
Improve code coverage by:
- removing unused features/code like filter/query functions from vector classes and fastpfor decoding wich was not included in used in the decoder. Fastpfor will defenitifly be re-introduced. 
- unit tests

### DIFF
--- a/ts/package-lock.json
+++ b/ts/package-lock.json
@@ -13,7 +13,7 @@
       },
       "devDependencies": {
         "@mapbox/vector-tile": "^2.0.4",
-        "@maplibre/maplibre-gl-style-spec": "^24.3.0",
+        "@maplibre/maplibre-gl-style-spec": "^24.3.1",
         "@types/bytebuffer": "^5.0.49",
         "@types/earcut": "^3.0.0",
         "@types/node": "^24.9.2",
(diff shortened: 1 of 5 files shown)


ts/package-lock.json
HarelM, line 1: Make sure to revert this file change...

HarelM:
Nice! this increased coverage to 67%.
image

ts/src/decoding/integerDecodingUtils.spec.ts
HarelM, line 1: Should this file be in a folder called `encodings` when it does decoding?

ts/src/vector/featureTable.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { describe, it, expect } from "vitest";
+import { decodeTile } from "../index";
+import path from "node:path";
+import fs from "node:fs";
+
+const TILE = path.resolve(__dirname, "../../../test/expected/tag0x01/simple/multiline-boolean.mlt");
```
HarelM: This is basically tested in the mltDecoder code, isn't it? I think it would be better to focus here on unit tests, or if these are too complicated completely remove this test.
Generally speaking, I would like to have end to end tests like those done in the mltDecoder and the rest of the tests be unit tests that don't load files and don't perform network calls.
This is mainly from an ordering perspective.
If this test is important, move it to the mltDecoder.spec.ts file.

ts/src/decoding/integerDecodingUtils.spec.ts
Turtelll: i don't know why this file was called encoding in the first place. It only includes decoding. So, i will rename it

ts/src/vector/featureTable.spec.ts
Turtelll: I moved it.

PR #662, opened 2025-10-30
Add workflow to bump version & create tag MLT JS

louwers:
If this works I will update the release workflow to be triggered by a tag push.

I have used a deploy key here. GitHub recommends using a GitHub App for this scenario, but since we don't have the right permissions in the MapLibre organization that would be a bit harder to manage.

Uses `npm version`: https://docs.npmjs.com/cli/v9/commands/npm-version

### DIFF
--- a/ts/package.json
+++ b/ts/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@maplibre/mlt",
-  "version": "0.0.1",
+  "version": "0.0.1-alpha.12",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
--- a/ts/package-lock.json
+++ b/ts/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@maplibre/mlt",
-  "version": "0.0.1",
+  "version": "0.0.1-alpha.12",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@maplibre/mlt",
-      "version": "0.0.1",
+      "version": "0.0.1-alpha.12",
       "license": "(MIT OR Apache-2.0)",
       "dependencies": {
         "@mapbox/point-geometry": "^1.1.0"


ts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@maplibre/mlt",
-  "version": "0.0.1-alpha.12",
+  "version": "0.0.1",
```
HarelM: I would revet this for now.

.github/workflows/release-js-create-tag.yml
```diff
@@ -0,0 +1,46 @@
...
+          git remote set-url origin git@github.com:${{ github.repository }}.git
+
+      - name: Update version, commit, and create tag
+        working-directory: ts
+        run: |
+          npm version ${{ github.event.inputs.release_type }} --no-git-tag-version
+          VERSION=$(node -p "require('./package.json').version")
+          git add package.json package-lock.json
+          git commit -m "chore: bump MLT JS version to ${VERSION}"
+          git tag -a "js-v${VERSION}" -m "Release js-v${VERSION}"
```
HarelM: Will this create a github release?

ts/package.json
louwers: I just want to try out bumping the version.

I don't think `npm version` can handle this non-standard version string.

.github/workflows/release-js-create-tag.yml
louwers: No, it will just push a tag.

I will work on making the release next.

.github/workflows/release-js-create-tag.yml
```diff
@@ -0,0 +1,46 @@
...
+  release-js-create-tag:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+
+      - uses: ./.github/actions/mlt-setup-node
+
+      - uses: webfactory/ssh-agent@a6f90b1f127823b31d4d4a8d96047790581349bd
```
HarelM: In maplibre-gl-js we have a "Create bump version PR" workflow to do this, it doesn't need special permissions I think as it simply creates a PR.

There's another CI that run every commit to main and check if the version changed and does the rest of the publishing, like we have here, but what we have here doesn't create a github release and doesn't add a tag.
So if we are keeping the other workflow I would guess it might be better to update it.

I generally wanted to align with other workflows in this repo, I can copy the workflow from gl-js if needed.

HarelM:
Looks like the release docs are also not up-to-date, I think it was copied from gl-js if I needed to guess.
https://github.com/maplibre/maplibre-tile-spec/blob/main/RELEASE.md

louwers:
@HarelM Please take another look.

.github/workflows/release-js.yml
```diff
@@ -36,50 +32,48 @@ jobs:
...
 
       - name: Get version
-        id: package-version
-        uses: martinbeentjes/npm-get-version-action@v1.3.1
-        with:
-          path: ts
+        run: |
+          version="$(node -p "require('./package.json').version")"
+          echo version="$version" >> "$GITHUB_ENV"
+          if [[ "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
HarelM, line 47: Why was this changed? I think semver is a better tool than this regexp...

.github/workflows/release-js.yml
```diff
@@ -36,50 +32,48 @@ jobs:
...
-
       - name: Build
-        working-directory: ts
         run: |
           npm run build
 
-      - name: Publish NPM package (regular)
-        working-directory: ts
-        if: ${{ steps.prepare_release.outputs.release_type == 'regular' }}
+      - name: Release (GitHub) - Create Draft
```
HarelM, line 60: Consider doing this only after npm succeeds. I had issues in the past and moving this down a bit reduces the need to delete a draft GitHub release...

ts/package.json
HarelM: Can you try this locally maybe? I wouldn't want that merging this PR will create a new npm version when it's still not ready (although it should be in the next few days)...

.github/workflows/release-js.yml
HarelM: I think this can be done in a single step like we are doing in gl-js...?
https://github.com/maplibre/maplibre-gl-js/blob/e626b5d528f96ed416bc3488ec4b263f488d2ed5/.github/workflows/release.yml#L112

HarelM:
I've added minor comments, this looks great, thanks!!

.github/workflows/release-js.yml
louwers: Which tool?

Simple stuff like this I like to just do inline, since external tools can be compromised.

louwers: Ah the npm package. Sure!

.github/workflows/release-js.yml
louwers: I prefer doing it this way. If something goes wrong when a GitHub release is created, it's possible that there is a npm release without a corresponding GitHub release.

I am also used to this pattern, because without it it is not possible to create an immutable release with release assets. Not really a problem here, but still.

HarelM: Fair enough.

.github/workflows/release-js.yml
louwers: I think this approach is more readable.

And relying on one less external package is preferable in my opinion... Not dying on that hill though.

ts/package.json
louwers: Reverted.

.github/workflows/release-js.yml
HarelM: As long as it works I don't have objection...

PR #666, opened 2025-10-31
chore(java): update Java FSST tests

TimSylvester:
- Re-enable FSST unit tests
- Allow FSST tests to run without the native JNI module
- ~~Fix the JNI module build~~


louwers:
Java 25 is installed on the runners (not Java 22).

https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#macos-14

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

A Windows build is not yet included in `nl.bartlouwers.fsst4j`...

TimSylvester:
Should I update it to 25 then?  Seems like it should b as low as it can be for widest compatibility.  But I would have thought 25 could build a project set to 22.

The code only uses the native implementation if it's present, so it should be fine that it's not available on all platforms.

louwers:
Cool, looks like the FSST tests passed for macOS.

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/fsst/FsstEncoder.java
```diff
@@ -1,5 +1,7 @@
 package org.maplibre.mlt.converter.encodings.fsst;
 
+import nl.bartlouwers.fsst.SymbolTable;
```
louwers, line 3: Need to publish this under the MapLibre org before merging.

PR #668, opened 2025-11-02
Improve CI to only run on relevant folders changes

HarelM:
This hopefully sorts out the CI to be similar for all languages.
The relevant tests will only run for changes in the relevant folder + when changing the test folder as this where assets maybe that are shared between the different languages.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
         args: ["-style=file:.clang-format"]
 
   - repo: https://github.com/Mateusz-Grzelinski/actionlint-py
-    rev: v1.7.7.24
+    rev: v1.7.8.24
     hooks:
       - id: actionlint
         additional_dependencies: [ shellcheck-py ]


.github/workflows/rust.yml
```diff
@@ -2,9 +2,13 @@ name: Rust CI
 
 on:
   push:
-    branches: [ main ]
+    paths:
+      - 'rust/**'
+      - 'test/**'
   pull_request:
-    branches: [ main ]
```
nyurik, line 7: Why are you deleting main here? Usually prs should run on a PR against main, or on merge to main, but not on all branches

HarelM: I've aligned this with the cpp.yml. Also why not run them on a PR from one branch to another?

birkskyum (approved):

HarelM:
@birkskyum can you check how to remove the requirement of "CI Finished"? I went to the settings -> branches but I don't see any branch rules, is there a different way to configure this?

birkskyum:
@HarelM it appear to be the "Rules" settings (don't know when it has been rolled out). If you go to this repo's settings, under Rules, and then "main" there's a section like this ([link](https://github.com/maplibre/maplibre-tile-spec/settings/rules/4545508)):

Screenshot 2025-11-04 at 13 48 04

HarelM:
Thanks! I couldn't find it, it was hidden under "show more" apparently...
This screen is full of bugs... :-/
But I managed to get it updated...

PR #671, opened 2025-11-03
Fix plain string columns in the C++ decoder

TimSylvester:
- Fix plain string decoding and add support for optional string columns
- ~~Update expected files with `just generate-expected-mlt`~~


TimSylvester:
#586

I'm not exactly sure what's involved, but I can try to copy it from MLN

PR #680, opened 2025-11-05
Add maxzoom option to encoder

Salkin975:
Add an optional maxzoom parameter to the Java CLI MBTiles converter, enabling users to limit the generated zoom levels when the source data contains more zoom layers than needed for the output

### DIFF
--- a/java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
+++ b/java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
@@ -112,7 +112,7 @@ public class Encode {
 
     int maxzoom = Integer.MAX_VALUE;
     if (cmd.hasOption("maxzoom")) {
-        maxzoom = Integer.parseInt(cmd.getOptionValue("maxzoom"));
+      maxzoom = Integer.parseInt(cmd.getOptionValue("maxzoom"));
     }
 
     if (verbose > 0 && !columnMappings.isEmpty()) {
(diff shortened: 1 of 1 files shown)


java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
```diff
@@ -488,6 +495,13 @@ private static void encodeMBTiles(
         try {
           while (tiles.hasNext()) {
             final var tile = tiles.next();
+            if (tile.getZoom() > maxzoom) {
```
HarelM, line 498: Are tiles ordered by zoom level so this won't miss tiles in lower zoom level accidentally?

Salkin975: I have tried with multiple mbtiles datasets and it seems that some of them are in the right order and some aren't

Salkin975: I have tried multiple datasets and it seems that the order differs, some are and some aren't

HarelM: The question around this comment is if the fact that the order is different cause the code to stop before processing all the tile?

Salkin975: Yes, the conversion will be prematurely terminated if the tiles are not arranged in the correct order, so information will be lost.

TimSylvester:
I'm currently working on splitting the advanced option into more granular options, and that will conflict badly, so I'm going to wrap these changes in with that work.

TimSylvester:
See #684

PR #682, opened 2025-11-06
Add Plain String Encoding Support to TypeScript Decoder

Turtelll:
This PR adds support for decoding plain (non-dictionary) string encoding to the TypeScript decoder and adapts changes to stream count.


HarelM:
I think it would be better to only have the ts changes as part of this PR...
Without the cpp and the tile charges...
This will allow to release a new version and avoid keeping the PR hostage until charges are made to the other decoders...

TimSylvester:
Why do I see diffs in C++ code that already appear to be part of main?

https://github.com/maplibre/maplibre-tile-spec/blob/main/cpp/src/mlt/decode/string.hpp#L39

HarelM:
This branch is probably not up to date with main if I needed to guess...

TimSylvester:
Still doesn't make sense to me, but I suppose it doesn't matter.  Looks like this also adds FastPFOR support?

Turtelll:
No unfortunately not but i will look to add support for FastPFOR next.

PR #683, opened 2025-11-06
Add Plain String Encoding Support to TypeScript Decoder

Turtelll:
This PR adds support for decoding plain (non-dictionary) string encoding to the TypeScript decoder and adapts changes to stream count.

### DIFF
--- a/ts/src/decoding/stringDecoder.spec.ts
+++ b/ts/src/decoding/stringDecoder.spec.ts
@@ -91,6 +91,65 @@ function setupVarintDecodeMock(value: number | number[] = 0): void {
     });
 }
 
+describe('decodePlainStringVector', () => {
+    it('should return null when plainLengthStream is null', () => {
+        const result = (StringDecoder as any).decodePlainStringVector('test', null, new Uint8Array([1, 2, 3]), null, null);
+        expect(result).toBeNull();
+    });
+
+    it('should return null when plainDataStream is null', () => {
+        const result = (StringDecoder as any).decodePlainStringVector('test', new Int32Array([0, 3]), null, null, null);
+        expect(result).toBeNull();
+    });
+
+    it('should return StringDictionaryVector when offsetStream exists (non-nullable)', () => {
+        const plainLengthStream = new Int32Array([0, 3, 7]);
+        const plainDataStream = new Uint8Array([97, 98, 99, 100, 101, 102, 103]);
+        const offsetStream = new Int32Array([0, 1]);
+
+        const result = (StringDecoder as any).decodePlainStringVector('test', plainLengthStream, plainDataStream, offsetStream, null);
+
+        expect(result).toBeDefined();
+        expect(result.name).toBe('test');
+    });
+
+    it('should return StringDictionaryVector when offsetStream exists (nullable)', () => {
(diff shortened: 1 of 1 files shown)


HarelM:
Thanks!
Note that there are failing unit tests that needs to be fixed and I would also recommend adding a few more for the case you added.
I agree that using the tiles is a good way to cover this code, but unit tests are also important.

ts/src/decoding/stringDecoder.ts
```diff
@@ -77,17 +95,45 @@ export class StringDecoder {
...
+            const nullabilityBuffer = bitVector ?? presentStream;
+            return nullabilityBuffer
+                ? new StringDictionaryVector(name, offsetStream, dictionaryLengthStream, dictionaryStream, nullabilityBuffer)
                 : new StringDictionaryVector(name, offsetStream, dictionaryLengthStream, dictionaryStream);
         }
 
-        return bitVector
-            ? new StringFlatVector(name, offsetStream, dictionaryStream, bitVector)
-            : new StringFlatVector(name, offsetStream, dictionaryStream);
+        // Plain string encoding: LENGTH + DATA streams
```
HarelM: This comment it exactly the type of comments that require to export part of the code to a method and call that method instead of adding this comment :-)

Turtelll: I refactored the decode methode and fixed tests

ts/src/decoding/stringDecoder.ts
```diff
@@ -40,54 +44,141 @@ export class StringDecoder {
...
-                dictionaryStream,
-                symbolLengthStream,
-                symbolTableStream,
-                bitVector,
-            );
-        } else if (dictionaryStream) {
-            return bitVector
-                ? new StringDictionaryVector(name, offsetStream, dictionaryLengthStream, dictionaryStream, bitVector)
-                : new StringDictionaryVector(name, offsetStream, dictionaryLengthStream, dictionaryStream);
+        return this.decodeFsstDictionaryVector(
```
HarelM, line 88: I think there are too many ?? in this area, it makes it hard to understand what's going on here...

HarelM:
Did you add unit tests for the new code you added?

ts/src/decoding/stringDecoder.ts
Turtelll: It decodes with Fsst Dictionary or Plain depending on what streams are available I think it's ok.

Turtelll:
I will add a unit test for decodePlainStringVector

ts/src/decoding/stringDecoder.spec.ts
```diff
@@ -91,6 +91,65 @@ function setupVarintDecodeMock(value: number | number[] = 0): void {
     });
 }
 
+describe('decodePlainStringVector', () => {
+    it('should return null when plainLengthStream is null', () => {
+        const result = (StringDecoder as any).decodePlainStringVector('test', null, new Uint8Array([1, 2, 3]), null, null);
```
HarelM, line 96: Why is "as any" needed?

HarelM (approved):
I'll merge this to get things moving, but I think the tests should cover more of the code.

PR #686, opened 2025-11-07
Update expected files with plain and shared string encodings

TimSylvester:
Re-encode the test data now that all three decoders support plain and shared string encodings.

### DIFF
--- a/rust/mlt-nom/src/errors.rs
+++ b/rust/mlt-nom/src/errors.rs
@@ -110,6 +110,15 @@ pub enum MltError {
         multiple_of: usize,
         got: usize,
     },
+    #[error("byte length expected multiple of {expected_multiple_of}, got {got}")]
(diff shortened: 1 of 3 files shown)


nyurik (approved):

HarelM:
Nice work!
BTW, is the rust code supplicated?
I remember that the CI was targeting two folders and I'm not sure which one was in use...

HarelM:
I don't see the rust CI as part of the list in the PR, did I cause it?

TimSylvester:
I don't know.  When I ran the tests locally the Rust tests failed initially, until I updated the snapshots, so it's definitely checking them.

The files in this PR should match both `rust/**` and  `test/**` in `rust.yml`, so I don't know why it isn't in the list of checks.

HarelM:
They ran, they just have a bad name, I'll fix that.

PR #690, opened 2025-11-10
chore(ci): minor CI and justfile cleanup

nyurik:
* run all CI when `.github/**` changes
* do not install protoc in CI - it is no longer needed by mlt-nom, and mlt is not being tested

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,4 +1,4 @@
-name: gh-pages
+name: Publish docs
 
 on:
   push:
--- a/.github/workflows/java-release.yml
+++ b/.github/workflows/java-release.yml
@@ -1,4 +1,4 @@
-name: Release to Maven Central
+name: Java - Release
 
 on:
   push:
--- a/.github/workflows/ts-bump-version.yml
+++ b/.github/workflows/ts-bump-version.yml
@@ -1,3 +1,5 @@
+name: Manual TypeScript Version Bump
+
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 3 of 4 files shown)


.github/workflows/rust.yml
HarelM, line 1: Can you add Rust to the job names? I wanted to do but haven't got to it.
When all the tests are running, it's not clear what "Test" and "Test MSRV" are testing, so there's a need to add the word Rust there.
If you want to have better names to other CI runs by adding the `name: ...` to the others that can be nice as well.

HarelM:
We can probably remove the bench-js stuff as it doesn't do anything at the moment.

.github/workflows/rust.yml
nyurik: could you clarify where we only see `Test`?  The review below and actions tabs look like these - so pretty clear what is what

image

image

HarelM: I can't find it right now, but when I looked at the following PR while it was running (I think) there was not prefix of which CI it is and all I saw was "Test" (which is the name of the `test` job in the rust CI).
https://github.com/maplibre/maplibre-tile-spec/actions/runs/19174437302/job/54815374220

In any case, it is not aligned with other CIs, so I think we should align it.

nyurik: sure, if you look at a specific workflow, it only shows job names. The workflow itself is in the top left corner:

image

---

let's maybe remove the duplication completely?  This way this screen will be much cleaner too:

image

HarelM: I don't have strong feelings either way, just keep the CI files similar in terms of job "naming".

nyurik: sounds good. Most of the files already have a `name` field. The jobs on the other hand some do and some don't, and we can clean them up in a subsequent PR if desired

Issue #694, opened 2025-11-11
LICENSE missing in npm package


thurnbauermatthi:
There are three licenses included at the top level directory, the `package.json` in the ts module also specifies `(MIT OR Apache-2.0)`, but the npm package does not contain a LICENSE file.

Which license should be used for the mlt package?

I am happy to provide a PR for that, I just need to know which license should be used for the npm package.
The license file which should be copied should also be renamed to `LICENSE` or `LICENCE`, so that npm pack can collect it.

HarelM:
@mactrem, @nyurik do you know which license should be applied to the mlt npm package?

thurnbauermatthi:
I have already talked to Markus today in our office, he has no preference regarding the license, he told me to open an issue and ask

HarelM:
I'm good with MIT, but I'm not sure I can decide :-)

nyurik:
See root Readme, w already specify MIT or apache 2, at user choice

thurnbauermatthi:
Then how should the licenses be included?

We can either add both files to the `files` property in the package.json, so that npm pack collects them, or merge both license files and create a standard LICENSE.txt or LICENSE.md.

I tried putting both license files (LICENSE-MIT and LICENSE-APACHE) into the mlt folder under node modules, but our license checker failed picking them up, because of the suffix.

HarelM:
Is it possible that this is a limitation of the checker then?

HarelM:
According to google's AI, there's a need to write "x OR y" license in package.json file and include both LICENSE files in the npm package (`files` property).

thurnbauermatthi:
Yes that could possibly be an issue in the checker.

Yes the `x OR y` is fine 👍 
So both licenses should be put into the package then, got it.

PR #698, opened 2025-11-12
Add step to copy licenses before publish

thurnbauermatthi:
This change copies the `LICENSE-MIT` and `LICENSE-APACHE` files into the `ts/` directory BEFORE publishing the npm package. The package.json specifies both files to be included in the final package.
I have also added the missing copyright in the Apache license file.

Fixes:
https://github.com/maplibre/maplibre-tile-spec/issues/694

### DIFF
--- a/ts/package.json
+++ b/ts/package.json
@@ -4,9 +4,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
-    "/dist",
-    "LICENSE-APACHE",
-    "LICENSE-MIT"
+    "/dist"
   ],
   "license": "(MIT OR Apache-2.0)",
   "homepage": "https://github.com/maplibre/maplibre-tile-spec/#readme",


mactrem:
LGTM. @nyurik Was it intentional that you did not include the copyright notice in the apache license file?

thurnbauermatthi:
@nyurik Is it really a good idea to include two separate files in the npm package? This is the only package in our project which has two license files.

LICENSE-APACHE
```diff
@@ -186,7 +186,7 @@ APPENDIX: How to apply the Apache License to your work.
    same "printed page" as the copyright notice for easier
    identification within third-party archives.
 
-Copyright [yyyy] [name of copyright owner]
+Copyright 2024 MapLibre contributors
```
HarelM, line 189: 2025?

thurnbauermatthi: I have just copied the copyright content from the other license file, but I am happy to change it to 2025

HarelM: They probably need an update as well...

thurnbauermatthi:
Btw. @HarelM I have tested both license files with the [license-checker](https://www.npmjs.com/package/license-checker) and it seems that only the first found file gets returned (-APACHE in our case due to alphabetical order)

HarelM:
Both are permissive licenses, but yeah, two licenses is odd for me as well...

thurnbauermatthi:
@HarelM Year is updated to 2025 now

HarelM:
Can you check why the pre-commit is complaining?

thurnbauermatthi:
Is fixed, there was a trailing whitespace

nyurik:
> Both are permissive licenses, but yeah, two licenses is odd for me as well...

Most Rust libraries/core has MIT+Apache2 licencing.  IANAL, but my understanding that this has to do with GPL, which is not very compatible with Apache2, but at the same time most FOSS experts prefer Apache2 as it gives much more protection to the users of the code. To my knowledge, it is possible to license things under MIT, but later sue for patent infringement.  So, for wider adaption, it is recommended to actually license under both to cover all cases, such as embedding into GPL work.

@mactrem
> Was it intentional that you did not include the copyright notice in the apache license file?

My mistake, i keep forgetting that Apache2 license has a notice down below.  MIT has it at the top. I should update all my apache2 files too, thx.

@thurnbauermatthi
> Is it really a good idea to include two separate files in the npm package? This is the only package in our project which has two license files.

It doesn't have to be two files if it causes issues. That said, I have been reading package.json [documentation](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#license) and [this blog](https://blog.inedo.com/npm/making-sense-of-npm-package-licenses) -- it seems license file is NOT required if it is a standard SPDX one, so including both of the files without referencing them in package.json should be acceptable per docs.  Note that even authors can be listed in the package.json as `contributors` field, so technically the license files are more of a help than a requirement. If the verifier tool is causing issues, we should ask the author of the tool to fix it. BTW, package.json essentially follows the same approach as Rust's cargo.toml with licensing - either a file or SPDX expression.

thurnbauermatthi:
@nyurik yes, looks like only having the SPDX expression in the package.json is sufficient for npm. I just looked into the license checker code and it seems like they do not support fetching the license with solely the SPDX expression in the package.json.

So not putting the license files inside the package should be enough, but apparently the most used license checker on npm cannot handle that...

nyurik:
i guess just put LICENSE.md with apache2 text, and LICENSE-MIT.md with mit, and that should be ok?

louwers (changes requested):
We tend not update license years in license years in MapLibre repos. We only include the first copyrightable year. If that's 2025 then that's OK.

thurnbauermatthi:
@louwers I changed both copyright texts to 2024

thurnbauermatthi:
@nyurik I am unsure now how to proceed, I think having both files as they are in the package is enough, regardless of the suffix or file ending.

Just for reference, this is the current output of the license-checker:
```bash
├─ @maplibre/mlt
│  ├─ licenses: (MIT OR Apache-2.0)
│  ├─ repository: https://github.com/maplibre/maplibre-tile-spec
│  ├─ path: C:\...\node_modules\@maplibre\mlt
│  └─ licenseFile: C:\...\node_modules\@maplibre\mlt\README.md
```

This is the output with the license file in the package:
```bash
├─ @maplibre/mlt
│  ├─ licenses: (MIT OR Apache-2.0)
│  ├─ repository: https://github.com/maplibre/maplibre-tile-spec
│  ├─ path: C:\...\node_modules\@maplibre\mlt
│  └─ licenseFile: C:\...\node_modules\@maplibre\mlt\LICENSE-APACHE
```

HarelM:
Since this is a copy step your can concat both licenses into the same file I believe.

thurnbauermatthi:
@HarelM Done

ts/package.json
```diff
@@ -4,7 +4,9 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
-    "/dist"
+    "/dist",
+    "LICENSE-APACHE",
```
HarelM: This needs an update I believe, right?

thurnbauermatthi: Correct, is changed

louwers (approved):

HarelM:
@thurnbauermatthi let me know if this is urgent enough to get a new version released or if this could wait a week or two for the next release.

thurnbauermatthi:
@HarelM This can wait until the next release 👍

PR #695, opened 2025-11-11
Support decoding tessellation in Java

TimSylvester:
Decodes the triangles, but doesn't do anything with them.

Issue #660, opened 2025-10-30
Error occur when converting or decoding of mvt-render-test tiles


Salkin975:
When converting MVT tiles and subsequently testing them with the decoder tests, the following errors occur:
```bash
TypeError: Cannot read properties of null (reading '0')
 - /src/vector/flat/stringFlatVector.ts:16:28

TypeError: Cannot read properties of null (reading 'length')
 - /src/vector/fsst-dictionary/stringFsstDictionaryVector.ts:102:59

RangeError: byte length of Float64Array should be a multiple of 8
 - /src/encodings/decodingUtils.ts:115:16

Error: Only delta encoding is supported for transforming length to offset streams yet.
 - /src/encodings/integerStreamDecoder.ts:308:15

Error: FastPFor is not implemented yet.
 - /src/encodings/integerDecodingUtils.ts:130:11
```

I am unable to determine whether this is an issue with the conversion or the decoding process.

The java decoder also seems to fail:
```bash
Decoding failed:                                                                                         
org.apache.commons.lang3.NotImplementedException: Not implemented yet.
        at org.maplibre.mlt.decoder.GeometryDecoder.decodeGeometryColumn(GeometryDecoder.java:55)
        at org.maplibre.mlt.decoder.MltDecoder.decodeMltLayer(MltDecoder.java:88)
        at org.maplibre.mlt.decoder.MltDecoder.parseBasicMVTEquivalent(MltDecoder.java:29)
        at org.maplibre.mlt.decoder.MltDecoder.decodeMlTile(MltDecoder.java:42)
        at org.maplibre.mlt.cli.Decode.main(Decode.java:73)
```

These are the base mvt-tiles as well as the converted mlt-tiles with wich the errors occur

[FaultyTiles.zip](https://github.com/user-attachments/files/23226501/FaultyTiles.zip)

@TimSylvester

TimSylvester:
Assuming you included tessellation in the encoding, the Java decoder exception is expected.

I haven't been keeping up with the state of the TypeScript decoder.  It looks like some null checks are needed, but I can't tell if this is due to unsupported cases or invalid data.  @HarelM?

HarelM:
@Salkin975 keep me honest here:
The source of the faulty tiles are from the maplibre-gl-js repo render tests, converted to MVT in order to try running the render tests with MLT instead of MVT (to flush out bugs early on).
As part of this process, the above tiles failed decoding with the js version.
[...]

PR #699, opened 2025-11-12
Fixes incorrect treatment of triangle indices in the Java decoder

TimSylvester:
- Fixes incorrect treatment of tessellated triangle indices in the Java decoder
- Apply layer filters to MVT while comparing to avoid spurious errors.
- Enable the one working tessellation unit test
- Also includes some troubleshooting enhancements used in debugging the issue, and some cleanup and refactoring

Follow-up to #695 

With this, I'm able to encode-decode-compare all the tiles in #660 with or without tessellation, FSST, and FastPFOR.

### DIFF
--- a/java/mlt-core/build.gradle
+++ b/java/mlt-core/build.gradle
@@ -38,6 +38,7 @@ dependencies {
 }
 
 test {
+    jvmArgs "-Dcom.google.protobuf.use_unsafe_pre22_gencode"
     useJUnitPlatform()
     testLogging {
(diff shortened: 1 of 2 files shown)


HarelM:
I see some changes in the encoder as well, does it has to do anything with the fix or is this just an improvement you added since you worked on the code for debugging?

TimSylvester:
Just refactoring.  There are no changes to the encoded data, which I verified by re-generating the expected files.

PR #700, opened 2025-11-12
Use fixed Earcut4j

TimSylvester:


louwers:
If it works fine I will work towards onboarding it or getting my fixes merged. On that last point, did not get any reply to my [PR](https://github.com/earcut4j/earcut4j/pull/10) so probably I will try to onboard it).

TimSylvester:
It seems to work as well as before, based on the OMT tiles in the test directory, with [maplibre-native/#3945](https://github.com/maplibre/maplibre-native/pull/3945).

I was holding off in hopes of it being merged, but it seems like time to use it anyway.

louwers:
@TimSylvester `org.maplibre.earcut4j` is publishing...

image

PR #704, opened 2025-11-14
Add back methods and properties for columnar buckets

Salkin975:
- [x] add back vector files with full implementation
- [x] add back missing methods
- [x] create unit-tests for every re-added file to keep test coverage
- [ ] resolve all TODOs
- [ ] eliminate 'not implemented yet' Errors


HarelM:
I would consider finding a way to split this PR into smaller pieces for easier review.
At it's current state it is very large which increases the risk of not addressing code review comments and not reviewing important parts of the code.
A possible solution would be to create a branch here as part of this repo and open small PRs against it, or simply open small PRs against main branch if this is possible.

Salkin975:
I will split it up

PR #705, opened 2025-11-14
refactor(cpp): Herrnb f/fsst

FloHerrnb:
During implementing and testing my own fsst implementation for maplibre-tile-spec, other commits were merged which added fsst support.

With this refactor I try to improve readability and keep performance.

Also wanted to add some edge cases in testing.

happy for feedback and discussions. I´m curious about your suggestions

### DIFF
--- a/justfile
+++ b/justfile
@@ -158,13 +158,13 @@ mkdocs-build:
     cd mkdocs && docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build --strict
 
 # Build Java encoder and generate .mlt files for all .pbf files in test/fixtures
-[working-directory: 'java']
+#[working-directory: 'java']
 generate-expected-mlt:  (cargo-install 'fd' 'fd-find')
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -158,13 +158,13 @@ mkdocs-build:
     cd mkdocs && docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build --strict
 
 # Build Java encoder and generate .mlt files for all .pbf files in test/fixtures
-[working-directory: 'java']
+#[working-directory: 'java']
```
TimSylvester: Aren't these necessary?

FloHerrnb:
@TimSylvester any thoughts regarding that?

CommanderStorm:
@TimSylvester if you want to add this to the cpp part, we can reopen it, otherwise I am going to close the PR because a bit old/stale and has not gotten any reviews.
Sorry that the refactoring did not went in, despite looking good to me

PR #708, opened 2025-11-17
Add selection vectors

Salkin975:
adds base methods for re integration of filter methods

- [x] reach test coverage of over 90% on added files

### DIFF
--- a/ts/src/vector/filter/selectionVectorUtil.spec.ts
+++ b/ts/src/vector/filter/selectionVectorUtil.spec.ts
@@ -24,11 +24,30 @@ describe("selectionVectorUtils", () => {
     });
 
     describe("createNullableSelectionVector", () => {
-        it("Should create FlatSelectionVector when filtering by BitVector", () => {
+        it("Should return all indices when nullabilityBuffer is not provided", () => {
+            const sv = createNullableSelectionVector(5);
+            expect(sv).toBeInstanceOf(FlatSelectionVector);
+            expect(sv.limit).toBe(5);
+        });
+
+        it("Should return empty vector when size is 0 and nullabilityBuffer is not provided", () => {
+            const sv = createNullableSelectionVector(0);
+            expect(sv).toBeInstanceOf(FlatSelectionVector);
+            expect(sv.limit).toBe(0);
+        });
+
+        it("Should return empty vector when size is 0 with BitVector provided", () => {
+            const buffer = new Uint8Array([0b11111111]);
+            const bitVector = new BitVector(buffer, 8);
+            const sv = createNullableSelectionVector(0, bitVector);
+            expect(sv).toBeInstanceOf(FlatSelectionVector);
+            expect(sv.limit).toBe(0);
+        });
+
+        it("Should create FlatSelectionVector with only set bits", () => {
             const buffer = new Uint8Array([0b00001011]); // bits 0, 1, 3 are set
             const bitVector = new BitVector(buffer, 8);
             const sv = createNullableSelectionVector(8, bitVector);
-
             expect(sv).toBeInstanceOf(FlatSelectionVector);
             expect(sv.limit).toBe(3);
         });
@@ -37,7 +56,6 @@ describe("selectionVectorUtils", () => {
             const buffer = new Uint8Array([0b00000000]);
             const bitVector = new BitVector(buffer, 8);
             const sv = createNullableSelectionVector(8, bitVector);
-
             expect(sv).toBeInstanceOf(FlatSelectionVector);
             expect(sv.limit).toBe(0);
         });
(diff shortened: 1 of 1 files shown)


ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
+import {beforeEach, describe, it, expect} from "vitest";
+import {SequenceSelectionVector} from "./sequenceSelectionVector";
+
+let seqVector: SequenceSelectionVector;
```
HarelM: It is best to move this and the `beforeEach` into the first (main) `describe` method. This way when reading this it easy to understand that they are "linked".

ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+
+beforeEach(() => {
+    seqVector = new SequenceSelectionVector(0, 1, 5);
```
HarelM: I also see that this is sometimes different from test to test, so I would consider completely removing this `beforeEach` and simply initialize the same values in every test, I know this is somewhat repeating yourself but it help understand what each test it testing.

ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+            seqVector.setIndex(0, 25);
+            seqVector.setIndex(2, -48);
+            seqVector.setIndex(3, 1000000000000001);
+
+            expect(seqVector.getIndex(0)).toBe(25);
+            expect(seqVector.getIndex(2)).toBe(-48);
+            expect(seqVector.getIndex(3)).toBe(1000000000000001);
+        });
+
+        it("Should materialize array on first set", () => {
```
HarelM: I'm not sure the title of this test is reflected in the test itself.
This is probably because materialize in an internal implementation details that is hard to reflect in the tests.
I would consider just making sure things are working as expected without "knowing" how they work internally.

ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+            // After setting, array should be materialized
+            customVector.setIndex(2, 999);
+            expect(customVector.getIndex(2)).toBe(999);
+
+            // Other values should still be correct
+            expect(customVector.getIndex(0)).toBe(10);
+            expect(customVector.getIndex(1)).toBe(15);
+        });
+
+        it("Should return Index out of bounds", () => {
```
HarelM: How is this different from the previous test?

ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+        });
+    });
+
+    describe("setLimit Test", () => {
+        it("Should set limit", () => {
+            seqVector.setLimit(250);
+            expect(seqVector.limit).toBe(250);
+            seqVector.setLimit(0);
+            expect(seqVector.limit).toBe(0);
+            seqVector.setLimit(-125);
```
HarelM: How is this a valid limit?

ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+            expect(seqVector.limit).toBe(-125);
+        });
+
+        it("Should change limit independently of capacity", () => {
```
HarelM: I'm not sure I fully understand the difference between the two, it might be worth adding some comment in the code to explain this.

ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+            const values = seqVector.selectionValues();
+            expect(values).toStrictEqual([0, 1, 2, 3, 4]);
+        });
+
+        it("Should return empty array for empty vector", () => {
+            const emptyVector = new SequenceSelectionVector(0, 1, 0);
+            expect(emptyVector.selectionValues()).toStrictEqual([]);
+        });
+
+        it("Should return correct values with custom base and delta", () => {
```
HarelM: How is this different from the `Should return selectionVector` test?

ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+            expect(negVector.selectionValues()).toStrictEqual([100, 90, 80, 70, 60]);
+        });
+
+        it("Should cache materialized array", () => {
+            const values1 = seqVector.selectionValues();
+            const values2 = seqVector.selectionValues();
+            expect(values1).toBe(values2); // Same reference
+        });
+
+        it("Should return materialized array after setIndex", () => {
```
HarelM: I think there's no need to mention the `materialized` part of it, the test simply makes sure that after setIndex the selectionValues still works as expected.

ts/src/vector/filter/sequenceSelectionVector.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+        });
+
+        it("Should initialize limit to size", () => {
+            const customVector = new SequenceSelectionVector(10, 5, 50);
+            expect(customVector.limit).toBe(50);
+            expect(customVector.capacity).toBe(50);
+        });
+    });
+
+    describe("Memory efficiency Test", () => {
```
HarelM: I think you'll have a hard time actually testing this.
Or to put it in other words, if the internal implementation would change these tests would still pass, so I don't think they are adding valuable information in terms of testing.
Consider removing them.

ts/src/vector/filter/sequenceSelectionVector.spec.ts
HarelM: I now see the explanation, in order to "copy" the comment from the interface use `/** @inheritdoc */` (do check me) to say that this is the same as the interface it implements.

ts/src/vector/filter/selectionVectorUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+    return new FlatSelectionVector(selectionVector);
+}
+
+export function updateSelectionVector(selectionVector: SelectionVector, nullabilityBuffer: BitVector) {
+    const buffer = selectionVector.selectionValues();
+    let limit = 0;
+    for (let i = 0; i < selectionVector.limit; i++) {
+        if (!nullabilityBuffer || nullabilityBuffer.get(i)) {
+            buffer[limit++] = buffer[i];
```
HarelM: Does this update the internal "memory" of the selectionVector? If so I think it's a bad idea...

ts/src/vector/filter/sequenceSelectionVector.ts
```diff
@@ -0,0 +1,59 @@
+import { type SelectionVector } from "./selectionVector";
+
+export class SequenceSelectionVector implements SelectionVector {
+    private _baseValue: number;
```
HarelM: You can use the same initialization as the `FlatVector` by using:
`constructor(private baseValue: number ...)`
I think it should be one or the other, but avoid using both, from a coding standard perspective.

ts/src/vector/filter/flatSelectionVector.ts
```diff
@@ -0,0 +1,49 @@
+import { type SelectionVector } from "./selectionVector";
+
+export class FlatSelectionVector implements SelectionVector {
```
HarelM, line 7: How is this different from a regular array? Because it implements `SelectionVector`? Do we really need this wrapper?

ts/src/vector/filter/flatSelectionVector.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (index >= this._limit || index < 0) {
+            throw new Error("Index out of bounds");
+        }
+
+        return this._selectionVector[index];
+    }
+
+    setIndex(index: number, value: number): void {
+        if (index >= this._limit || index < 0) {
+            throw new Error("Index out of bounds");
```
HarelM: The other implementation uses `RangeError`, maybe it's worth using the same type of error?

ts/src/vector/filter/flatSelectionVector.spec.ts
```diff
@@ -0,0 +1,76 @@
+import {beforeEach, describe, it, expect} from "vitest";
+import {FlatSelectionVector} from "./flatSelectionVector";
+
+let vector: number[];
+let FSVector: FlatSelectionVector;
```
HarelM: Use a different casing for this variable name (for example `fsVector`). This looks like a class name...

ts/src/vector/filter/selectionVectorUtil.spec.ts
HarelM, line 1: There might be a bit too many tests in this file.
The code in the utils should be tested without testing the code of `FlatSelectionVector` or `SequenceSelectionVector` as much as possible.
This test files feels like it repeats some of the other tests, which I'm not sure it necessary.

ts/src/vector/filter/sequenceSelectionVector.spec.ts
Salkin975: see next comment

ts/src/vector/filter/sequenceSelectionVector.spec.ts
Salkin975: Each test now initializes its own vector, making it clear what values each test validates.

ts/src/vector/filter/sequenceSelectionVector.spec.ts
Salkin975: Changed test names in general to reflect the expected outcome better

ts/src/vector/filter/sequenceSelectionVector.spec.ts
Salkin975: I added a check in the setLimit method

ts/src/vector/filter/sequenceSelectionVector.spec.ts
Salkin975: I changed the index test setup to check get and set Limit together

ts/src/vector/filter/sequenceSelectionVector.spec.ts
Salkin975: I have added the /** @inheritdoc */

ts/src/vector/filter/flatSelectionVector.spec.ts
Salkin975: I changed the name

ts/src/vector/filter/flatSelectionVector.ts
Salkin975: Changed the error to also be of type RangeError

ts/src/vector/filter/selectionVectorUtils.ts
Salkin975: It did, but I changed it to not use it anymore

ts/src/vector/filter/sequenceSelectionVector.ts
Salkin975: I created the constructor

ts/src/vector/filter/selectionVectorUtil.spec.ts
Salkin975: I have removed unnecessary tests

ts/src/vector/filter/selectionVectorUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+    for (let i = 0; i < selectionVector.limit; i++) {
+        if (!nullabilityBuffer || nullabilityBuffer.get(i)) {
+            filteredIndices.push(selectionVector.getIndex(i));
+        }
+    }
+    return new FlatSelectionVector(filteredIndices);
+}
+
+/** Returns new selection vector with only indices where nullability is true. */
+export function updateNullableSelectionVector(selectionVector: SelectionVector, nullabilityBuffer: BitVector): SelectionVector {
```
HarelM: Isn't this and the method above identical?

ts/src/vector/filter/selectionVectorUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+    const filteredIndices = [];
+    for (let i = 0; i < selectionVector.limit; i++) {
+        if (!nullabilityBuffer || nullabilityBuffer.get(i)) {
```
HarelM: Are you sure about this condition? It looks reversed...

ts/src/vector/filter/flatSelectionVector.ts
Salkin975: It was originally the only SelectionVector implementation. I added the sequenceSelectionVector as part of a todo (https://github.com/Turtelll/maplibre-tile-spec/blob/a8edfa36480d32fd5fd17770bb3974b5fda8c7fc/ts/src/vector/constant/intConstVector.ts#L17), so I don't think it should be removed
https://github.com/Turtelll/maplibre-tile-spec/tree/a8edfa36480d32fd5fd17770bb3974b5fda8c7fc/ts/src/vector/filter

ts/src/vector/filter/selectionVectorUtils.ts
Salkin975: yes they are, I will remove one

ts/src/vector/filter/selectionVectorUtils.ts
Salkin975: it used the wrong index, I updated it to use the actual Index of the selectionVector

ts/src/vector/filter/flatSelectionVector.ts
HarelM: Fair enough.

ts/src/vector/filter/selectionVectorUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+    return new FlatSelectionVector(selectionVector);
+}
+
+/** Returns new selection vector with only indices where nullability is true. */
+export function updateNullableSelectionVector(selectionVector: SelectionVector, nullabilityBuffer: BitVector): SelectionVector {
+    const filteredIndices = [];
+    for (let i = 0; i < selectionVector.limit; i++) {
+        const vectorIndex = selectionVector.getIndex(i);
+        if (!nullabilityBuffer || nullabilityBuffer.get(vectorIndex)) {
```
HarelM, line 36: Doesn't `!nullabilityBuffer` means that if `nullabilityBuffer` is defined (not null) it will add all the values of the original selection vector? I'm still confused.
I did look at the tests, and they seems to test this, but I'm still finding this code confusing, what am I missing?

ts/src/vector/filter/selectionVectorUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+import { SequenceSelectionVector } from "./sequenceSelectionVector";
+
+export function createSelectionVector(size: number) {
+    return new SequenceSelectionVector(0, 1, size);
+}
+
+export function createNullableSelectionVector(size: number, nullabilityBuffer: BitVector) {
+    const selectionVector = [];
+    for (let i = 0; i < size; i++) {
+        if (nullabilityBuffer?.get(i)) {
```
HarelM: If `nullabilityBuffer` is not passed it will return an empty array, wouldn't it?
It is the expected results? I haven't seen a test to cover this input.

ts/src/vector/filter/flatSelectionVector.ts
```diff
@@ -0,0 +1,59 @@
...
+    setIndex(index: number, value: number): void {
+        if (index >= this._limit || index < 0) {
+            throw new RangeError("Index out of bounds");
+        }
+
+        this._selectionVector[index] = value;
+    }
+
+    /** @inheritdoc */
+    setLimit(limit: number): void {
```
HarelM, line 41: I think this method is missing coverage...

ts/src/vector/filter/selectionVectorUtils.ts
Salkin975: I didn't implement the method myself, but I understand it as:
No nullability buffer = all values are considered NOT null
With nullability buffer = only values where the bit is 1 are valid

HarelM: Yeah, you are right, `!nullabilityBuffer` is equal to `nullabilityBuffer == null`, so all values will be added.

HarelM:
Thanks for all the work on this!
> ✅ All modified and coverable lines are covered by tests.

Is a great indication of good coverage!
There's only one last comment remaining before merging this, unless you would like to wait with it for other parts to be added?

ts/src/vector/filter/selectionVectorUtils.ts
Salkin975: I have added comments to both methods and changed them to use the same logic. I also added tests for these edge cases

ts/src/vector/filter/flatSelectionVector.ts
Salkin975: is done

Salkin975:
No, these changes are required for the other ones

PR #712, opened 2025-11-18
Add basic vector utility methods

Salkin975:
- [x] Add basic methods in vector.ts
- [x] Create a general test suite for the methods
- [x] Mark incompatible methods in stringFlatVector
- [x] Mark stringFlatVector as not compatible with base implementation
- [x] Mark incompatible methods in boolFlatVector
- [x] All new and existing tests pass
- [x] The question about the .slice method is resolved

### Related changes planned for further pull requests
- Edge case tests for float, int, double, long, and bool
- Implementation and tests for stringFlatVector methods


Salkin975:
The Issue with the .selectionValues() has been solved the following way:
- directly updating the selection vector's internal _selectionVector array through its public API in the vector.ts methods and not just editing a copy of it
- changing the .selectionVector method in the flat/sequenceSelectionVector to slice the current array to the limit 
- removing the .slice() methods from the vector.spec.ts tests

ts/src/vector/filter/flatSelectionVector.ts
```diff
@@ -47,7 +47,7 @@ export class FlatSelectionVector implements SelectionVector {
 
     /** @inheritdoc */
     selectionValues(): number[] {
-        return this._selectionVector;
+        return this._selectionVector.slice(0, this._limit);
```
mactrem: The proposed implementation returns a slice of the selection vector, which creates an unnecessary copy of the data. To avoid this copy overhead, I directly exposed the internal array and delegated limit handling to callers, which breaks encapsulation and should be improved for production use.

**Proposed Solution**
Switch from JavaScript arrays to TypedArrays (Uint32Array) to enable zero-copy access via ``.subarray()``
  
**Future Optimization**
Consider implementing pooled/reusable buffers for selection vectors to avoid repeated allocations

ts/src/vector/flat/booleanFlatVector.ts
```diff
@@ -12,4 +13,16 @@ export class BooleanFlatVector extends Vector<Uint8Array, boolean> {
     protected getValueFromBuffer(index: number): boolean {
         return this.dataVector.get(index);
     }
+    override greaterThanOrEqualTo(value: boolean): SelectionVector {
```
HarelM: A way to avoid these unimplemented exceptions is by splitting the interface into multiple interfaces and each class will implement just the relevant interfaces.

ts/src/vector/vector.ts
```diff
@@ -33,5 +41,190 @@ export default abstract class Vector<T extends ArrayBufferView = ArrayBufferView
         return this._size;
     }
 
+    presentValues(): SelectionVector {
```
HarelM, line 44: I'm not sure this is an intuitive name for what this method does... Can we try and find a better name?

ts/src/vector/vector.ts
```diff
@@ -33,5 +41,190 @@ export default abstract class Vector<T extends ArrayBufferView = ArrayBufferView
...
+        const selectionVector = [];
+        for (let i = 0; i < this._size; i++) {
+            if (this.has(i) && this.getValue(i) === value) {
+                selectionVector.push(i);
+            }
+        }
+        return new FlatSelectionVector(selectionVector);
+    }
+
+    filterNotEqual(value: K): SelectionVector {
```
HarelM: Instead of making an awkward inheritance tree, maybe all these function should be exported as utility methods for the vector class.
i.e. in a different file and receive a vector and return a selection vector.
It might make more sense, IDK...

HarelM:
Due to the "throw not implemented" that is in both string and bool vectors I think the design in regards to inheritance is probably not a great one. 
I would consider moving the filtering, matching etc to a different file and only receive classes that implements "filterable" vector or something like that.

Salkin975:
When I applied the changes, I noticed the following. The StringFlatVector only inherits the baseVector natively. Should the string implementation also inherit filterableVector and override the methods in it, or should the methods in it be implemented without inheritance?

ts/src/vector/filter/flatSelectionVector.ts
Salkin975: Doesn't the use of Uint32Array introduce bugs when the values are negative, or does this case never occur?

HarelM:
I think you have two types here: 
1. FilterableVector
2. ComparableVector

It might be that all vectors are filterable, if so, this type might not be "intersting" and we can ignore it.
Regarding comparable, this definitely does not apply to all vector types.
And in general, while in the past there was a significant focus on inheritance, base classes etc, I believe this is no longer the case. 
The idea of code sharing does not automatically imply inheritace, and in may cases I would recommend composition over inheritance. 
Instead of adding all this code to the vector class you could easily create "external" methods to do the same work (as the method added here are all using public APIs of the vector class).
This both allows for tree shaking if the methods are not in use by the code and also create a separation of creating selection vectors in the sense that a vector does not know what a selection vector is, which is a good decoupling.
In typescript you can even define a type that is a union of other types.
`type ComparableVector = IntVecotor | FloatVector | ...` and write the greaterThan, or smallerThan to only work on those types. 
You could even reduce some verbosity and define a general method that receives a predecate (function that receives two values and return a boolean) in order to make this a lot more generic. 
I hope this answers your question, if not, do let me know.

ts/src/vector/filter/flatSelectionVector.ts
mactrem: indices, can't be negative and evene if we want't to handle specific cases e.g. with -1 we can use a Int32Array. The point is we should also use for this case typed arrays

mactrem (changes requested):
I recommend adopting the vectors from https://github.com/mactrem/mlt-evaluation/tree/main/ts/src/vector. That implementation was built with performance in mind, because of that in several places it intentionally chooses small amounts of duplication over a generic abstraction to avoid overhead. 

We only need to add the missing stubs (that throw for unsupported errors) which were skipped during the implementation of the POC because the filters/styles used didn’t require them. For example, StringFlatVector is currently unimplemented. The only class that doesn’t fully satisfy its contract is BooleanFlatVector, so if we want strict OO compliance (Liskov Substitution Principle), we can consider a dedicated handling strategy only for this vector. @HarelM  I’m not generally a fan of inheritance, but in this case i think it’s justified.

Issues with this current generic base‑class approach:
- Using a single set of generic methods for all types tends to create polymorphic or megamorphic call sites, which V8’s inline caches handle poorly and can trigger deoptimizations. In hot paths, generic JS is often a performance liability.
- A key design principle is to operate directly on compressed/encoded data without decoding, because it’s faster (no decoding, faster processing). That puts filter logic in the subclasses that understand the encoding, so a generic “getValueFromBuffer” abstraction doesn’t fit. This is for example the case for the stringDictionaryVector and the fsstStringDictionaryVector (which is not fully implemented yet). The goal is also to directly operate on the UTF-8 Buffer when filtering and avoid costly string materialization. 

@HarelM One open question is whether we want to validate the UTF-8 buffer for security reasons. I discussed this with @FloHerrnb (cybersecurity specialist), and their view is that this additional validation is likely unnecessary for our use case, but we should confirm as a team.

Salkin975:
I will revert to the POC approach / state. Since the code can not be used currently and is not fully tested I would appreciate it to get a seperate branch where these can be fully implemented and tested before merging it into the production branch. This would also make it possible to create pull requests that are incomplete, since the reverting and changing affects 22 files without adding any appitional methods / tests from me

ts/src/vector/vector.ts
Salkin975: what is your opinion to validValues(Selected)

HarelM:
I don't see a reason to validate UTF-8 buffer for security reasons.
It's worth mentioning that type/javascript supports also functional programming well. So not everything has to be a class or a class method.
It's also a lot easier to tree shake it, so I would advise to see how this fits in the concept of the design here.
From the code I see here I would even go a step further and advise to use it instead of bloating the base class.
For example, if we add some of the methods as "stand alone" methods and end up not using them we improve the bundle size of maplibre-gl-js (or even introduce them as an initial step and only later on use them in gl-js), which is not the case for inheritance (not to mention these "throw if not implemented");

We've seen how the bundle size was reduced when all these method were removed before the release of the previous version.
We can keep the functionality in the code while not bloating the package if maplibre-gl-js is not using it, while it might be required by other users.

HarelM:
@Salkin975 can you clarify what you need? If you need to work on a branch but still have CI running, you can do that on your fork, can't you?

Salkin975:
yes, I could implement the whole thing on my fork. The only issue is that the pull request would have around 6000 lines that get added wich is very impractical to review.
My idea was to make the changes in smaller chucks and create Pull requests on a seperate branch in the root repository, so it can be reviewed in said chuncks and then merged when everything works as intended and is tested properly.

HarelM:
There are some strict rules on breaches here (which I don't like, but that's besides the point).
We can do the same process on your fork, just tag me and Marcus on the relevant PRs there.

PR #714, opened 2025-11-19
feat(ts): add fastPfor decoder to Typescript 

Turtelll:
This PR introduces fastPfor decoding to typesctipt
- testcoverage is reached with unit 
- i tested the decoder with a large fastPfor encoded tileset and re-generated the expected testdata and tests pass


ts/src/decoding/fastPforDecoder.spec.ts
```diff
@@ -0,0 +1,136 @@
+import {describe, expect, it} from "vitest";
+import {decodeFastPfor} from "./fastPforDecoder";
+import IntWrapper from "./intWrapper";
+
+// encoded data was generated using the java fastPfor encoder and expected values are re-generated in the tests
+const ENCODED_NON_ALINGED_358_ENCODED = new Uint8Array([0, 0, 1, 0, 0, 0, 0, 65, 3, 2, 1, 0, 7, 6, 5, 4, 11, 10, 9, 8, 15, 14, 13, 12, 19, 18, 17, 16, 23, 22, 21, 20, 27, 26, 25, 24, 31, 30, 29, 28, 35, 34, 33, 32, 39, 38, 37, 36, 43, 42, 41, 40, 47, 46, 45, 44, 51, 50, 49, 48, 55, 54, 53, 52,
```
mactrem: Add the FastPfor test data as binary files to keep the tests more clean.
Also let's add more test cases to ensure that the implementation is also valid for edge cases.

ts/src/decoding/fastPforDecoder.ts
```diff
@@ -0,0 +1,317 @@
...
+    const dataView = new DataView(
+        encodedValuesSlice.buffer,
+        encodedValuesSlice.byteOffset,
+        encodedValuesSlice.byteLength,
+    );
+    const numCompleteInts = Math.floor(byteLength / 4);
+    const hasTrailingBytes = byteLength % 4 !== 0;
+    const numInts = hasTrailingBytes ? numCompleteInts + 1 : numCompleteInts;
+
+    const intValues = new Int32Array(numInts);
```
mactrem, line 31: there are two copies generated (data. slice and intValues) before decoding the actual data. Let's reduce the number of data transformations, because at least we do not need slice. Additionaly, if the offset into the data buffer is aligned (to 4 bytes) no copy is needd at all. One approach to complety avoid the copies is to directly work on the DataView but which leads to a slower data access performance see https://v8.dev/blog/dataview. But first let's add benchmarks (against varint as baseline) and then let's optimize

ts/src/decoding/fastPforDecoder.spec.ts
Turtelll: i switched to using binary files and will add more tests

Turtelll:
I will start benchmarking the decoder with different tilesets i encoded accordingly. But i'm not sure if we want to have those benchmarks in the repo itself.

mactrem:
> I will start benchmarking the decoder with different tilesets i encoded accordingly. But i'm not sure if we want to have those benchmarks in the repo itself.

We can think about publishing a seperate package for the fastpfor decoder

DoFabien:
Thanks for the PR, awesome to see FastPFOR progress in TS, and the test fixtures are really helpful.

A couple of thoughts that could make this easier to merge:

## 1. Scope

From what I see, this PR currently combines a few distinct concerns:
- FastPFOR decoder + tests/fixtures
- RLE `byteLength`/offset handling in the decoders
- `FeatureTable.getFeatures()` removal (API change)

Would you be open to splitting these? The RLE fix seems valuable and mergeable on its own, while the `FeatureTable` change is an API behavior change that likely deserves its own discussion/review.

## 2. FastPFOR implementation

I’ve been working on a TS port of the reference `CompositeCodec(FastPFOR + VariableByte)`, with:
- **Symmetry** (encode + decode) for easier testing
- **Reference-vector tests** based on `cpp/test/test_fastpfor.cpp`
- Extra edge-case handling (trailing bytes packing, page-size alignment, safer bit masks)

If you’re open to it, I’ve opened a draft PR we can compare/rebase onto: https://github.com/maplibre/maplibre-tile-spec/pull/758

## 3. Benchmarks

I also added an optional Benchmark.js harness to compare `decodeFastPfor` vs `decodeVarintInt32`.

**High-level results:**
- **Compression:** Consistent 22–61% reduction vs Varint.
- **Performance:** Faster on higher-entropy data and competitive on sequential data (full numbers in the draft PR).

---

Happy to help split the PRs or coordinate the merge strategy. Let me know what you prefer!

Turtelll:
Thanks for the feedback!

My FastPFOR implementation is just a draft and very slow, so your implementation in #758 is definitely preferred.

I agree on splitting the scope. I will move the RLE byteLength/offset fixes to a separate PR. The FeatureTable.getFeatures() removal is intended for a future step once a iterator is implemented, so I will put that on hold for now.

Just a heads-up: I won't be able to contribute further until after New Year's, but I will handle the split and cleanup when I return.

PR #718, opened 2025-11-23
Add sources content to source map files

HarelM:
Source mapping is using the `.ts` files (not inline source mapping).
So in order to get proper source mapping, there's a need to include the `.ts` files as well in the npm package.
I've discovered this while running vitest with coverage on maplibre-gl-js as vitest complained that it can't find the source files linked in the source mapping...
This PR inlines the ts files into the source map files.

### DIFF
--- a/ts/package.json
+++ b/ts/package.json
@@ -4,8 +4,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
-    "/dist",
-    "/src"
+    "/dist"
   ],
   "license": "(MIT OR Apache-2.0)",
   "homepage": "https://github.com/maplibre/maplibre-tile-spec/#readme",
--- a/ts/tsconfig.json
+++ b/ts/tsconfig.json
@@ -6,6 +6,7 @@
     "module": "es2022",
     "declaration": true,
     "sourceMap": true,
+    "inlineSources": true,
     "moduleResolution": "bundler",
     "esModuleInterop": true,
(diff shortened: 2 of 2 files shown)


louwers (changes requested):
I don't think npm packages are supposed to include `.ts` files.

This is a dependency for maplibre-gl-js so it should not be included in coverage.

HarelM:
Some packages do include ts files. 
Moreover, the other option is to include a complicated build tool, which I'm trying to avoid if possible, this is the easiest solution I believe.

louwers:
What packages? I have read in various places that it is (still) discouraged. The biggest problem (I believe) is that it would be quite expensive to compile TypeScript for all dependencies. Although when also distributing `.js` and `.d.ts` files that is not really a problem.

If you want to avoid complicated build tooling, another option to consider is JavaScript with JSDoc comments. This is the approach [svelte](https://github.com/sveltejs/svelte) is now using. You can then still use the TypeScript compiler for type checking, but don't have to use it for transpilation for distribution.

My question remains though: why is coverage of this package generated on the maplibre-gl-js side?

HarelM:
I'm not 100% sure, if I need to guess, vitest is travesing the code and looking for source maps before starting to run the tests, and report these warnings. 
I don't have strong feelings towards this solution, but it's a lot simpler than any thing else I can think of....

HarelM:
@louwers I've updated the solution according to our slack conversation.

louwers (approved):

PR #726, opened 2025-11-29
support geometry collections in Java encoder

breunigs:
This patch recursively flattens GeometryCollections to allow their encoding.

My current pipeline sometimes generates tiles with an (empty) GeometryCollection. While potentially worth fixing in the pipeline/upstream projects, it would still be nice if it didn't prevent experimenting with MLT. 

Here's one such tile that doesn't encode without the patch: [example.zip](https://github.com/user-attachments/files/23833208/example.zip) Zipped because of Github. ODbL licensed because it contains OpenStreetMap data.

### DIFF
--- a/java/mlt-core/src/test/java/org/maplibre/mlt/converter/encodings/GeometryEncoderTest.java
+++ b/java/mlt-core/src/test/java/org/maplibre/mlt/converter/encodings/GeometryEncoderTest.java
@@ -1,0 +1,122 @@
+package org.maplibre.mlt.converter.encodings;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.locationtech.jts.geom.*;
+import org.maplibre.mlt.converter.MLTStreamObserverDefault;
+import org.maplibre.mlt.metadata.stream.PhysicalLevelTechnique;
+
+public class GeometryEncoderTest {
+
+  private static final GeometryFactory GEOM = new GeometryFactory();
+  private static final Polygon polygon1 =
+      GEOM.createPolygon(
+          GEOM.createLinearRing(
+              new Coordinate[] {
+                new Coordinate(1, 2),
+                new Coordinate(3, 4),
+                new Coordinate(5, 6),
+                new Coordinate(1, 2),
+              }));
+  private static final Polygon polygon2 =
+      GEOM.createPolygon(
+          GEOM.createLinearRing(
+              new Coordinate[] {
+                new Coordinate(9, 8),
+                new Coordinate(7, 6),
+                new Coordinate(5, 4),
+                new Coordinate(9, 8),
+              }));
+
+  private static final GeometryCollection nestedPolygon2 =
+      GEOM.createGeometryCollection(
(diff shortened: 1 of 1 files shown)


HarelM:
Thanks for taking the time to open this PR!
This obviously can't go in without a test. 
Preferably also with decoder and not just encoder support...

breunigs:
Can you clarify what you mean by decoder support? Is the format desired to be round-trippable, i.e. `source == decode(encode(source))`? If so, then the approach in this PR cannot work as it doesn't retain the GeometryCollections.

HarelM:
I don't know what the solution should be, but if there's support for encoding there should be a support for decoding. I don't know if a round trip should be identical, but at least it should be lossless as much as possible. Is this the case here?

breunigs:
No, since it converts something like `obj1, obj2, [obj3, obj4]` into `obj1, obj2, obj3, obj4`, losing the grouping. The grouping is the only property that GeoCollections offer.

Curious, it's not supported (according to spec) on the classic tile format… let me check more closely where the GeometryCollection gets introduced. Unless the format was extended (and this extension is supported by Maplibre clients!), it maybe hints at a classic tile format decoder bug in the Java parts?

breunigs:
For the example at hand it's the vector-tile-java library introducing an empty collection whenever it fails to parse the geometry from the MVT, e.g. https://github.com/ElectronicChartCentre/java-vector-tile/blob/master/src/main/java/no/ecc/vectortile/VectorTileDecoder.java#L216-L218 
It fails because the source in the MVT is a degenerate polygon (one coordinate is fixed, so it becomes a linestring).

Since according to MVT spec polygons must have an area greater than 0, this is a bug in the tile generator. I'll follow up with them.

So this PR doesn't make sense in its current form. It's still an adoption problem, because working but spec-violating MVTs cannot be converted. The tile renders fine in Maplibre JS+Native, and Mapbox' `vt2geojson` tool reproduces the degenerate polygon. Depending on how many people run into this, making the MLT encoder skip empty GeometryCollections might be an acceptable compromise. Making vector-tile-java pass-through degenerate polygons like that is probably possible, but the MLT encoder would probably still need to ignore it. Even if it doesn't break anything, it makes no sense to include them since they won't be rendered.

breunigs:
Correction: the [MVT tile spec](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/README.md#4344-polygon-geometry-type) actually does allow degenerate polygons:

> A linear ring SHOULD NOT have an area calculated by the surveyor's formula equal to zero, as this would signify a ring with anomalous geometric points.

It's a bit surprising given that it defines exterior/interior rings by this area as well:

> An exterior ring is DEFINED as a linear ring having a positive area as calculated by applying the [surveyor's formula](https://en.wikipedia.org/wiki/Shoelace_formula) to the vertices of the polygon in tile coordinates.

> An interior ring is DEFINED as a linear ring having a negative area as calculated by applying the [surveyor's formula](https://en.wikipedia.org/wiki/Shoelace_formula) to the vertices of the polygon in tile coordinates.

In any case, it seems that these polygons need to be at least handled gracefully. Would you accept a PR that just ignores GeometryCollections of size 0?

HarelM:
I suggest to open an issue, describe the problem and the proposed solution and then continue with implementation. 
This way everyone will be aligned on the direction of the solution, also we can make sure this is addressed in all the encoders.

breunigs:
done: https://github.com/maplibre/maplibre-tile-spec/issues/736

PR #738, opened 2025-12-03
test(js): refactor decoding tests

Turtelll:
This PR refactors the decoder test suite to improve test quality by introducing shared test utilities and rewriting tests to use proper data encoding.
Before tests used raw byte streams or mocked metadata objects. Now tests properly encode data using encoding utilities.

### DIFF
--- a/ts/src/decoding/integerStreamDecoder.spec.ts
+++ b/ts/src/decoding/integerStreamDecoder.spec.ts
@@ -50,9 +50,10 @@ describe("decodeIntStream", () => {
     it("should decode with PhysicalLevelTechnique.NONE", () => {
         const metadata = createStreamMetadata(LogicalLevelTechnique.NONE);
         const expectedValues = new Int32Array([10, 20, 30]);
+        const data = encodeVarintInt32Array(expectedValues);
         const offset = new IntWrapper(0);
 
-        const result = decodeIntStream(expectedValues, offset, metadata, false);
+        const result = decodeIntStream(data, offset, metadata, false);
 
         expect(result).toEqual(expectedValues);
     });


ts/src/decoding/decodingTestUtils.ts
HarelM, line 1: I might be reading this wrong, but I see some encoding methods that are not "hard coded".
If this is the case, and there might be in the far future a javascript encoder, we might as well place these methods in files that are symmetrical to the decoding files (in terms of naming).
Later on, we could use them in the test like you did.
Let me know if this makes sense.

HarelM: It seems there's a lot of logic in this file, I'm not sure if it is related to the previous comment or not, but it's worth taking it into account.

HarelM:
Please checkout the coverage of the decoding files:
- propertyDecoder.ts
- integerStreamDecoder.ts
- integerDecodingUtils.ts

I would expect those to only have missing coverage for some edge cases when throwing, or code that we don't have a good way to test (morton?).
Although, if we can't reach that code, it might be dead code that needs to be removed.
But I'm guessing a method like `decodeDoubleColumn` should be covered in tests, for example.

ts/src/decoding/decodingTestUtils.ts
Turtelll: I mean i can place them in a encoder file structure, but the encodings I included for the tests are really only the most basic encodings and half of the logic in this file is needed for creating stream(metadata). If I were to set those up manually we would have at least 10000 of extra code. The only thing that is "hard coded" are the FSST Symbol Table Streams.

TLDR: half of decodingTestUtils.ts is only for test setup and encodings like ZigZag can be moved to a symmetrical encoder directory. But I am not sure if this makes sense right now

HarelM: Let's start with something, move as much code as possible to encoding folder/files. 
If you can take a bit more time to create basic encoding methods that can facilitate for "full cycle encode-decode" tests that would be the best approach, but I don't know how complicated that is, I would assume that for the basic stuff this should be trivial.
This will both make the tests easier to read and maintain and help us in the future if we decide to create a js encoder.

Turtelll: Ok i will move them to encoding folders and add tests where coverage is still lacking.

Turtelll: "full cycle encode-decode" tests are already happening

ts/src/decoding/decodingTestUtils.ts
```diff
@@ -0,0 +1,461 @@
...
+}
+
+/**
+ * Encodes stream metadata into binary format.
+ * - Byte 1: Stream type (physical type in upper 4 bits, logical subtype in lower 4 bits)
+ * - Byte 2: Encodings (llt1[5-7], llt2[2-4], plt[0-1])
+ * - Varints: numValues, byteLength
+ * - If RLE: Varints: runs, numRleValues
+ */
+export function encodeStreamMetadata(metadata: StreamMetadata | RleEncodedStreamMetadata): Uint8Array {
```
HarelM, line 506: Shouldn't this be part of the encoding?

Turtelll: It is really specific to the testdata streams so i would keep it here

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -71,115 +33,413 @@ describe("getVectorType", () => {
...
+
+        const result = decodeFloat64Buffer(values, metadata, false);
+
+        expect(result).toEqual(new Float64Array([1.5, 2.5, 3.5]));
+    });
+
+    it("should decode NONE signed", () => {
+        const metadata = createStreamMetadata(LogicalLevelTechnique.NONE);
+        // ZigZag encoded: positive values are multiplied by 2
+        const values = new Float64Array([4, 10, 6]); // zigzag encoded [2, 5, 3]
```
HarelM: is it possible to create something like `encodeFloat64Buffer([2,5,3])` to improve full-circle readability?

ts/src/encoding/encodingUtils.ts
HarelM, line 1: I would expect this file to have similar folder structure to encoding, this way there's a way to see which method is the counter part of the decoding method.
I would expect to heve integerStremEncoder file, propertyEncoder file etc.

Turtelll: I understand what you mean but that would require a whole encoder for all the different streams and at that point we would have half of a fully functioning MLT encoder. So basically it would take a lot more effort then just creating streams like I do in the test utils. But the tests would be verry easy to adapt to using a fully implemented encoder once there is one.

HarelM: I think there might be a middle ground here, improve test readability by adding some encoders without implementing everything...

Turtelll: Yes but I think the tests are testing the decoder well enough and full cycle testing could wait until we introduce a encoder. There are still other gaps like fastPfor decoding we are missing that should be fixed before we spend a lot of time on this.

HarelM: The tests here seems very low level decoding - int, double, float etc, I'm failing to see how this would be a big effort...

Turtelll: I tried to implement your suggestions

ts/src/encoding/stringEncoder.ts
```diff
@@ -0,0 +1,365 @@
...
+    return concatenateBuffers(...streams);
+}
+
+/**
+ * Encodes FSST-compressed strings into a complete stream.
+ * This uses hardcoded test data: ["cat", "dog", "cat"]
+ * @returns Encoded Uint8Array that can be passed to decodeString
+ */
+export function encodeFsstStrings(): Uint8Array {
+    const symbolTable = new Uint8Array([99, 97, 116, 100, 111, 103]); // "catdog"
```
HarelM: This is hard coded values, right?
Can this method receive the string instead of having it hard coded?
If not, this probably needs to move to the relevant test helper.

ts/src/encoding/stringEncoder.ts
```diff
@@ -0,0 +1,365 @@
...
+        },
+    );
+
+    const dataStream = createStream(PhysicalStreamType.DATA, encodedDictionary, {
+        logical: new LogicalStreamType(dictionaryType),
+        count: encodedDictionary.length,
+    });
+
+    if (useFsst) {
+        const symbolTable = new Uint8Array([99, 97, 116, 100, 111, 103]); // "catdog"
```
HarelM: Same here...

ts/src/encoding/stringEncoder.ts
```diff
@@ -0,0 +1,365 @@
...
+}
+
+/**
+ * Encodes streams for a struct field.
+ * @param offsetIndices - Indices into the shared dictionary
+ * @param presentValues - Boolean array indicating which values are present
+ * @param isPresent - Whether the field itself is present
+ * @returns Encoded streams for the field
+ */
+export function encodeStructField(
```
HarelM: I can't find the "counter-decode" function for this, in which file is it?
It might be worth a file of its own as I'm not sure how it is related to string...

ts/src/encoding/propertyEncoder.ts
HarelM, line 1: The names in the decode are `decodeXXXColumn`, should the methods here called `encodeXXXColumn` to make it symmetrical?

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -71,115 +44,447 @@ describe("getVectorType", () => {
...
+        const result = decodeIntStream(data, offset, metadata, true);
+
+        expect(result).toEqual(expectedValues);
+    });
+
+    it("should decode RLE signed with Int32", () => {
+        const numRleValues = 5;
+        const runs = 2;
+        const metadata = createRleMetadata(LogicalLevelTechnique.RLE, LogicalLevelTechnique.NONE, runs, numRleValues);
+        const expectedValues = new Int32Array([100, 100, 100, -50, -50]);
```
HarelM, line 122: Is it possible to avoid these kind of tests where the expected values and the encoded values are different?
It would be much more clear if the test was similar to other tests where you:
```
const expectedValues = new Int32Array(...);
const data = encodeIntStream(..., expectedValues, ...);
const result = decodeIntStream(data, ...);
```

Is it a lot of work?

ts/src/encoding/propertyEncoder.ts
Turtelll: yea you are right

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -71,115 +44,447 @@ describe("getVectorType", () => {
...
+});
+
+describe("decodeFloat64Buffer", () => {
+    it("should decode NONE unsigned", () => {
+        const metadata = createStreamMetadata(LogicalLevelTechnique.NONE);
+        const values = new Float64Array([1.5, 2.5, 3.5]);
+
+        const result = decodeFloat64Buffer(values, metadata, false);
+
+        expect(result).toEqual(new Float64Array([1.5, 2.5, 3.5]));
```
HarelM: This can use the `values`, can't it?

HarelM:
I've added some more comments.
I think this version is going in the right direction in terms of file splitting, thanks!
I think it still needs some more work to reach the symmetry I'm looking for.

ts/src/encoding/stringEncoder.ts
Turtelll: FSST algorithms scans the data to find frequently occurring substrings so is not that easy to adapt a real FSST encoder. So i will move it out of the encoder for now.

ts/src/encoding/stringEncoder.ts
Turtelll: i will move this as well

ts/src/encoding/stringEncoder.ts
Turtelll: It is part of a shared decodeSharedDictionary case with childFields and its a string ScalarType so it should ultimately be a part of a encodeSharedDictionary wich is not implemented. So i will move it back to utils for now i guess.

ts/src/decoding/integerStreamDecoder.spec.ts
Turtelll: yes

ts/src/decoding/integerStreamDecoder.spec.ts
Turtelll: i fixed all the integerStreamTests to compare with expectedValues

HarelM:
I'll merge this and rework the things I find, there's still a lot of work here unfortunately...

HarelM (approved):

PR #739, opened 2025-12-05
feat(java): add optional int encoding config java

Weixing-Zhang:
Motivation
- Currently, integer encoding is automatically selected based on size optimization. This makes it difficult to generate predictable test outputs or compare different encoding strategies. This PR makes integer encoding explicitly configurable.

Changes
- `IntegerEncodingOption` enum: Adds explicit control over integer encoding with options
- Builder pattern for `ConversionConfig`: Replaces the implicit factory pattern of multiple constructor overloads with an explicit Builder.

Discussion
This is a proposal to make encoding methods explicitly configurable. If this approach works well, I plan to extend the same pattern to:
- String encoding configuration
- Geometry encoding configuration

This would provide full control over encoding strategies for testing and comparison purposes.

### DIFF
--- a/java/mlt-core/src/main/java/org/maplibre/mlt/converter/ConversionConfig.java
+++ b/java/mlt-core/src/main/java/org/maplibre/mlt/converter/ConversionConfig.java
@@ -8,6 +8,23 @@ import java.util.regex.Pattern;
 import org.jetbrains.annotations.NotNull;
 
 public class ConversionConfig {
+  public enum IntegerEncodingOption {
+    AUTO, // Automatically select best encoding (default)
+    PLAIN, // Force plain encoding
+    DELTA, // Force delta encoding
+    RLE, // Force RLE encoding (only for const streams)
+    DELTA_RLE // Force delta-RLE encoding
+  }
+
+  public static final boolean DEFAULT_INCLUDE_IDS = true;
+  public static final boolean DEFAULT_USE_FAST_PFOR = false;
+  public static final boolean DEFAULT_USE_FSST = false;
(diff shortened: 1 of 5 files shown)


TimSylvester:
Seems like a good start.  The other types will be more complex, of course.

I don't think is useful outside of test cases until it can be specified per layer/property, right?

nyurik:
@TimSylvester that's my understanding too - this is mostly to get us off the ground for creating unit tests, but in the future we could further expand this to allow complex configurations

Weixing-Zhang:
@TimSylvester I can extend this to per-layer/property configuration following the existing pattern like how optimizations uses `Map<String, FeatureTableOptimizations>`. Would you prefer I add that in this PR, or peep this PR focused on the foundation and add per-layer/property in a follow-up?

java/mlt-core/src/main/java/org/maplibre/mlt/converter/ConversionConfig.java
```diff
@@ -204,4 +245,101 @@ public Pattern getLayerFilterPattern() {
...
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /**
+   * Short builder for ConversionConfig with sensible defaults. Example: var config =
+   * ConversionConfig.builder() .includeIds(true) .useFastPFOR(true)
+   * .integerEncoding(IntegerEncodingOption.DELTA) .build();
+   */
+  public static class Builder {
```
msbarry, line 262: Thanks for adding this! The telescoping constructors are getting pretty unwieldy

TimSylvester (approved):

PR #745, opened 2025-12-10
update package-lock.json for encoding-server

louwers:
Should fix CI


louwers:
Good call, this should be better.

PR #749, opened 2025-12-11
Refactor TypeScript version bump workflow

HarelM:
Let's see if that can be used here as well, like in other repos.


HarelM:
I see there's some changelog stuff I need to fix before I merge this...

HarelM:
I think the reusable workflows are not well built for mono repos like this one, so I don't think this has enough merit to actually worth the trouble in changing them to support this.
While I did have this repo in mind when I built them, it felt too complicated to incorporate this logic.
We can always reopen this in the future if we think it's worth the effort.

PR #755, opened 2025-12-18
chore(java): update apache orc dependency

jsorel:
Updating depdencies is always a good thing.
And in this case it solves protobuf versions conflicts (mixed version 3 and 4 in the transitive dependencies)


louwers:
We removed it completely. 👍

PR #759, opened 2025-12-22
Fix docs creation CI

HarelM:
This is to test out what's causing the docs to fail in CI.


HarelM:
I've opened the following issue, so I hope this can be fixed shortly:
- https://github.com/PLAYG0N/mkdocs-panzoom/issues/44

In case the above is fixed this PR can be closed, otherwise we will need to comment out panzoom.

HarelM:
Looks like the issue is fixed, this is no longer needed, closing.

PR #760, opened 2025-12-22
Remove outdated test data

louwers:
These tests expectations use the outdated metadata format.


louwers:
@HarelM I think they contain the same data as the MLT/MVT. Sounds useful to keep around.

PR #771, opened 2026-01-04
Fix property coercion

msbarry:
While upgrading planetiler from 0.4 to 0.5 I had to disable this test: https://github.com/onthegomap/planetiler/blob/5c363906d408c82a0767d84aa11ad53e3bb64ec9/planetiler-core/src/test/java/com/onthegomap/planetiler/PlanetilerTests.java#L355-L404

It looks like there is/was a bug in property type coercion, it was updating the previous value with the wider type but then replacing that type immediately afterwards.


louwers (approved):
Yep, looks good.

PR #775, opened 2026-01-07
fix(ts): RLE decoding byte consumption

Turtelll:
This PR fixes issues with byte consumption in the RLE (Run-Length Encoding) decoder for boolean
  streams. The decoder was not correctly consuming all compressed bytes, causing offset misalignment when
  reading subsequent streams.
 - Added byteLength parameter to both functions to track compressed stream size
  - Implemented proper stream boundary checking
  - Ensures position is set to exact end of compressed data

### DIFF
--- a/ts/src/decoding/decodingUtils.spec.ts
+++ b/ts/src/decoding/decodingUtils.spec.ts
@@ -116,6 +116,24 @@ describe("decodingUtils", () => {
             expect(result[2]).toBe(3);
             expect(offset.get()).toBe(4);
         });
+
+        it("should handle truncated stream when byteLength runs out before numBytes", () => {
+            // Request 10 bytes but byteLength only allows 2 bytes (header + value)
+            // header=0 means numRuns=3, but stream ends after value byte
+            const data = new Uint8Array([0, 42]);
+            const offset = new IntWrapper(0);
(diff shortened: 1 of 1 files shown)


ts/src/mltDecoder.ts
```diff
@@ -89,7 +89,7 @@ export default function decodeTile(
                 if (columnMetadata.nullable) {
                     const presentStreamMetadata = decodeStreamMetadata(tile, offset);
                     const streamDataStart = offset.get();
-                    const values = decodeBooleanRle(tile, presentStreamMetadata.numValues, offset);
+                    const values = decodeBooleanRle(tile, presentStreamMetadata.numValues, presentStreamMetadata.byteLength, offset);
```
HarelM, line 92: remove comment below here too.

Turtelll: done.

HarelM:
Looks like there's an if around end of stream which is not covered in tests.
If this if is never reachable please remove it from the code, but if it is please add a test to cover it.

Turtelll:
This checks if we've consumed all bytes in the compressed stream before decoding all requested values. So i will add a test for it.

HarelM (approved):

PR #778, opened 2026-01-11
Java encoding performance improvements

msbarry:
Fixed some low hanging fruit on Java encoding performance ahead of first planetiler release with mlt support:

- Migrate hilbert encoding to this optimized algorithm: https://github.com/rawrunprotected/hilbert_curves (public domain)
- Reduce boxing/unboxing round trips for int/long lists by converting lists to int[] or long[] early on and changed some utilities to just take primitive arrays.
- Make decodeVarints that runs during encodeVarints an assertion so it only happens during tests or when run with `-ea`
- Switch approach for getting dictionary and morton encoded offsets to build a reverse index from value to index instead of doing a binary search each time

Overall this makes encoding 2-3x faster. I ran planetiler over the planet with openmaptiles profile on a 32 core machine with 64gb ram:

  | Tile encoding time | Total planetiler time
-- | -- | --
MVT | 14m | 1h28m
MLT from main | 55m | 2h13m
MLT from this branch | 29m | 1h44m

Fixes https://github.com/onthegomap/planetiler/issues/1349

### DIFF
--- a/java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/EncodingUtils.java
+++ b/java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/EncodingUtils.java
@@ -141,9 +141,8 @@ public class EncodingUtils {
     } while (v != 0);
 
     // ensure that the result decodes back into the input
-    if (DecodingUtils.decodeVarints(sink, new IntWrapper(offset), 1)[0] != checkValue) {
-      throw new IOException("Varint Overflow");
-    }
+    assert DecodingUtils.decodeVarints(sink, new IntWrapper(offset), 1)[0] == checkValue
+        : "Varint Overflow";
 
     return offset + sinkUsed;
   }
@@ -163,9 +162,8 @@ public class EncodingUtils {
       sink[offset + sinkUsed++] = b;
     } while (v != 0);
 
-    if (DecodingUtils.decodeLongVarint(sink, new IntWrapper(offset)) != checkValue) {
-      throw new IOException("Varint Overflow");
-    }
+    assert DecodingUtils.decodeLongVarint(sink, new IntWrapper(offset)) == checkValue
+        : "Varint Overflow";
     return offset + sinkUsed;
   }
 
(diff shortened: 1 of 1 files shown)


java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/EncodingUtils.java
```diff
@@ -142,9 +141,8 @@ static int putVarInt(int v, byte[] sink, @SuppressWarnings("SameParameterValue")
     } while (v != 0);
 
     // ensure that the result decodes back into the input
-    if (DecodingUtils.decodeVarints(sink, new IntWrapper(offset), 1)[0] != checkValue) {
-      throw new IOException("Varint Overflow");
-    }
+    assert DecodingUtils.decodeVarints(sink, new IntWrapper(offset), 1)[0] == checkValue
+        : "Varint Overflow";
```
msbarry, line 145: Is this necessary at runtime? A decent chunk of time is spent in it so I changed it to just run when `-ea` is enabled (ie. when running tests)

Image

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/GeometryEncoder.java
```diff
@@ -55,103 +55,84 @@ public static EncodedGeometryColumn encodePretessellatedGeometryColumn(
...
+            var y = (int) point.getY();
+            vertexBuffer.add(new Vertex(x, y));
           }
-        default:
-          throw new IllegalArgumentException(
-              "Specified geometry type is not (yet) supported: " + geometryType);
+        }
+        default ->
+            throw new IllegalArgumentException(
+                "Specified geometry type is not (yet) supported: " + geometry.getGeometryType());
```
msbarry, line 135: The switch is all the same, I just changed each

```java
case Geometry.TYPENAME_MULTIPOINT:
  var multiPoint = (MultiPoint) geometry;
  break;
```

to

```java
case MultiPoint multiPoint -> {
}
```

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/GeometryEncoder.java
```diff
@@ -536,130 +501,103 @@ public static EncodedGeometryColumn encodeGeometryColumn(
...
-            final var point = (Point) geometry;
-            vertexBuffer.add(new Vertex((int) point.getX(), (int) point.getY()));
-            break;
-          }
-        case Geometry.TYPENAME_LINESTRING:
-          {
-            // TODO: verify if part of a MultiPolygon or Polygon geometry add then to numRings?
-            geometryTypes.add(GeometryType.LINESTRING.ordinal());
-            var lineString = (LineString) geometry;
+      switch (geometry) {
```
msbarry, line 504: Same with this switch

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/EncodingUtils.java
TimSylvester: Good idea.  It's certainly not necessary, no, that was added to help track down a specific bug.  I don't think we have assertions enabled in the CLI tests run from `just`, but we should.

PR #780, opened 2026-01-12
feat: configurable encoding tests

Weixing-Zhang:
This PR introduces the ability to programmatically configure encoding parameters and input data within the Java test suite. It shifts our testing capability from static MVT fixtures to a configurable reference generator model.

The breakdowns are
- Hard-code deterministic tile geometry + attributes in Java
- Encode that into MLT using a specific ConversionConfig
- Decode the resulting MLT back into an in-memory representation
- Dump that decoded representation to JSON for inspection and comparison


test/synthetic/mixed-fastpfor-rle.json
```diff
@@ -0,0 +1,46 @@
+{
+  "layers": [
+    {
+      "extent": 4096,
+      "features": [
+        {
+          "geometry": "POINT (2993 411)",
```
michaelkirk: This is almost-but-not-quite a geojson FeatureCollection. Is there a reason to diverge? 

Using something standard like geojson could be nice, because you can plop it into qgis or whatever visualization tool and explore. 

Though it's not shown here, you may want multiple layers per file, so a single feature collection wouldn't suffice.

In that case you could do adapt a schema like:
```
{ layers: Array<FeatureCollection> }
```

At that point you wouldn't be able to just "drop it in" to a standard visualization tool, but it would be a very simple transformation to do so. (`jq .layers[0] > first_layer_feature_collection.geojson`)

...not sure if it's worth it. 🤷

Weixing-Zhang: Thanks for the questions — this is helpful, and you basically got the right idea, with one important clarification.

The JSON files in this PR are not inputs to the encoder. The encoder input is fully defined in code (hard-coded geometries, attributes, IDs, layer structure, etc.). The JSON is derived output, not a source format.

What this PR is doing is:
- Hard-code deterministic tile geometry + attributes in Java
- Encode that into MLT using a specific ConversionConfig
- Decode the resulting MLT back into an in-memory representation
- Dump that decoded representation to JSON for inspection and comparison

So `points-fastpfor.mlt` is the primary artifact. `points-fastpfor.json` is a human-readable, language-agnostic reference of the decoded result. The intent of the JSON is not to describe what goes into the encoder, but rather what a correct decoder should produce when given the MLT.

java/mlt-core/src/test/java/org/maplibre/mlt/MltFixtureGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+    tileGeometry.put("polygons", buildPolygonsTile(gf));
+    tileGeometry.put("mixed", buildMixedGeometriesTile(gf));
+    return tileGeometry;
+  }
+
+  // Simple fixture case: just the output name, tile to use, and config
+  private record FixtureCase(String outputName, String tileGeometryKey, ConversionConfig config) {}
+
+  // Define all the fixture cases - just pick short names that make sense
+  private static List<FixtureCase> defineFixtureCases() {
```
nyurik: I feel this could be made a bit simpler -- there is no point to create a list, keep it all in memory, etc.  We can have multiple functions like this, e.g. `writePoints() { ... }` and `writeLines() { ... }`, etc.

```java
// just a simple function that makes individual function calls for each fixture.
generateMltFixtures() {
    wrtiteMlt("points-plain", "points", ConversionConfig.builder()
                .includeIds(true)
                .useFastPFOR(false)
                .useFSST(false)
                .coercePropertyValues(false)
                .useMortonEncoding(false)
                .preTessellatePolygons(false)
                .outlineFeatureTableNames(outlineNames)
                .integerEncoding(plain)
                .build()));
    writeMlt("points-plain", "points", ConversionConfig.builder()
                .includeIds(true)
                .useFastPFOR(false)
                .useFSST(false)
                .coercePropertyValues(false)
                .useMortonEncoding(false)
                .preTessellatePolygons(false)
                .outlineFeatureTableNames(outlineNames)
                .integerEncoding(plain)
                .build()));
    // ...
}
```

Lastly, if you want to keep record of which files were generated, just have some internal state to keep a list of fixtures that were generated - this way you can have some `Generator` class - which has different modes if you really want to. I just feel returning a list is a bit strange.  Note that I haven't written any java for some years, so this might be a bad idea for Java-native developer.

test/synthetic/mixed-fastpfor-rle.json
Weixing-Zhang: The JSON here is intentionally not full GeoJSON. For this PR I am just reusing the existing dump utility `CliUtil.printMLT` to keep the scope small and focused on decoder validation, not visualization. I agree that the dump format could be improved. Using GeoJSON (or a { layers: [...] } wrapper for multi-layer tiles) would make inspection easier and seems like a reasonable follow-up once the fixture generation is settled.

nyurik: I think proper geojson but without changing the projection might be a good future improvement:

```json
{
  "type": "FeatureCollection",
  // "properties" on a collection is not valid GeoJSON per spec, but might be ok
  "properties": {
    "layer": "points",
    "extent": 4096
  },
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [2993, 411]
      },
      "properties": {
        "prop0": "value0"
      }
    }
  ]
}
```

java/mlt-core/src/test/java/org/maplibre/mlt/MltFixtureGenerator.java
nyurik: P.S. Perhaps there is not even a need for the `ConversionConfig` builder, but rather a more direct system for clarity:

```java
class FixtureGenerator {
    public static void main(String[] args) {
        // generator will have some state,
        // so the same calls could write fixtures or validate them, etc
        var gen = new FixtureGenerator();
        // generate fixtures, each function can create multiple fixtures
        gen.points();
        gen.lines();
    }

    void points() {
        // create json blob
        fixture("point-bool")
            .includeIds(true)
            .useFastPFOR(false)
            .useFSST(false)
            .coercePropertyValues(false)
            .useMortonEncoding(false)
            .preTessellatePolygons(false)
            .integerEncoding(plain)
            .add_point(2993, 411)  // call this function once per feature
            .add_props("""{"a": true}""")  // add properties as a dict, once per feature
            .run();
    }

    FixtureBuilder fixture(String layerName) {
        return new FixtureBuilder(layerName);
    }
}
```

Weixing-Zhang: @nyurik Thank you for the suggestions.

java/mlt-core/src/test/java/org/maplibre/mlt/MltFixtureGenerator.java
```diff
@@ -0,0 +1,159 @@
...
+import org.locationtech.jts.geom.GeometryFactory;
+import org.maplibre.mlt.cli.CliUtil;
+import org.maplibre.mlt.converter.ConversionConfig;
+import org.maplibre.mlt.converter.MltConverter;
+import org.maplibre.mlt.converter.mvt.MapboxVectorTile;
+import org.maplibre.mlt.data.Feature;
+import org.maplibre.mlt.data.Layer;
+import org.maplibre.mlt.decoder.MltDecoder;
+
+public class MltFixtureGenerator {
```
louwers: I don't think this belongs in the core MLT Java library.

Probably also not in the CLI, although that would be a better place.

Maybe it deserves its own Gradle subproject?

Weixing-Zhang: Good point. I separated it into the mlt-tools.

Weixing-Zhang:
Could you please review the PR one more time, just in case I miss any of your suggestions @nyurik @louwers?

test/synthetic/mixed-fastpfor-rle.json
HarelM: I think it's worth changing this as part of this PR instead of redoing it later. 
Unless the json and mlt files will not be committed, but this is currently not the date of this PR.

nyurik: I spoke with @TimSylvester about this - it seems "geojson-style" dumper is part of the core code already, plus a similar one might exist in other implementations - so not worth changing it as part of this PR.

HarelM: I'm sorry but I don't understand this answer and it doesn't address my concern about recommiting the files. 
If this is not made as part of this PR I would recommend opening a different PR, address that there and incorporate ot here before merging this.

nyurik: changing the format of the dump geojson files is needed, but this works blocks progress on stabilizing all implementations, so i'm ok to have a bit of churn for the text representation of MLTs.  That said, we need to progress on parser implementation, and for that we need a clear set of synthetic MLT + some text representation checked into this repo

HarelM: This PR is open for some days now, adding this change will take less time to actually write it than to debate why we should not wait for it. 
I'll do it tomorrow.

HarelM: Here it is: 
- #819

Weixing-Zhang: #819 Switching from WKT to GeoJSON-style geometry improves readability, but it does not address the fundamental document structure. The output is still not valid GeoJSON.

If the goal is proper GeoJSON export, we should implement a dedicated serializer that produces valid FeatureCollection output per layer.

However, this PR is only meant to be a debug JSON dump (not GeoJSON), then we can merge this PR as-is but clarify that it does not produce valid GeoJSON. What do you think, @HarelM ?

HarelM: The fact that you would be able to copy-paste the features property from the layers into geojson.io for example and see the features is a very important debug tool.
It doesn't have to be a valid geojson, but making it a bit closer to geojson helps. 
I don't see a reason not to do it.
The other PR is like 3 lines of code changes, so I would even argue that it can be incorporated as part of this PR, have the json updated and merge this. 
Or the other PR can be merged and this one updated. 
Easy debug is an important aspect of development...

Weixing-Zhang: @HarelM I added the GeoJSON writer instead. I hope this makes sense. 


cc: @nyurik

test/synthetic/mixed-fastpfor-rle.json
```diff
@@ -0,0 +1,131 @@
...
+      "geometry": {
+        "type": "Point",
+        "coordinates": [
+          20.0,
+          40.0
+        ],
+        "crs": {
+          "type": "name",
+          "properties": {
+            "name": "EPSG:0"
```
HarelM: I'm not sure this is correct, I would try and see if the crs can be committed.
Otherwise this looks good.

HarelM: Oh, wait, the coordinates are relative to the tile, right? 
So converting them to wgs:84 is a hassle. 😕

java/mlt-tools/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+dependencies {
+    implementation project(':mlt-core')
+    implementation project(':mlt-cli')
+    implementation 'org.locationtech.jts:jts-core:1.20.0'
+}
+
+java {
```
louwers, line 19: I think this file can be slimmed by sharing some code with the other subprojects.

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -0,0 +1,155 @@
...
+import org.maplibre.mlt.converter.MltConverter;
+import org.maplibre.mlt.converter.mvt.MapboxVectorTile;
+import org.maplibre.mlt.data.Feature;
+import org.maplibre.mlt.data.Layer;
+import org.maplibre.mlt.decoder.MltDecoder;
+
+public class SyntheticMltGenerator {
+
+  private static final int TILE_EXTENT = 4096;
+  private static final String OUTPUT_DIR = "../test/synthetic";
```
louwers: I'd pass this as an argument to the script instead of hardcoding it.

justfile
```diff
@@ -179,6 +179,10 @@ generate-one-expected-mlt file:
         --coerce-mismatch \
         --verbose
 
+[working-directory: 'java']
+generate-synthetic-mlts:
```
louwers, line 193: We still need some check on CI to ensure that the checked in data is up to date.

test/synthetic/mixed-fastpfor-rle.json
HarelM: In any case, I guess it would do for now...

test/synthetic/mixed-fastpfor-rle.json
HarelM: This is good enough from my point of view.

justfile
nyurik: good point, I added it to the CI as another just job.  Process is very simple: in CI, check that git is clean, regenerate MLTs, and assert git is still clean.

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
nyurik: lets not overcomplicate this - simpler is better, synthetics should be as hardcoded as possible.  I.e. convention over configuration.

java/mlt-tools/build.gradle
nyurik: what specifically do you want to share with other files?

nyurik:
I think this is done - the test files are being re-generated by CI automatically on all platforms, and each test file is guaranteed to be unique despite similar data being placed into the .mlt files (json files may be identical).  A lot more tests should be added, and that will be driven by coverage reports and other subsequent PRs.

CommanderStorm (commented):
I am happy to call this progress.

One thing that I would like to have added to the geojson is the how of it was generated -> `ConversionConfig` as `properties`

Otherwise, I would need to hardcode this on the rust side as well, which would be suboptimal.
Suboptimal also works though 🤷🏻

nyurik:
@CommanderStorm that would be somewhat problematic because:
* configuration pattern will be very different in other encoders. The current config-based system is a hack on top of a hack simply because java has no way of allowing low-level tile construction - instead you throw some data at it, and disable some optimizations via a global config system
* if multiple configurations produce identical MLTs, its a bug in the synthetic generator, but the moment we include that data into the MLT itself, all files will be different. Writing it into JSON though might work, but...
* we actually do not know what the internal structure of the MLT is - the global config makes some magical changes to the encoder, but which specifics of each item were used are ... a big secret :)

nyurik:
P.S. Lastly - the goal of the first round of synthetics is NOT to test other language `encoders`, but rather just the `decoders` -- be able to validate that every possible MLT 0x01 data can be verifiably decoded (certified).  Encoders is a bit further away.

CommanderStorm (approved):

CommanderStorm:
> java has no way of allowing low-level tile construction - instead you throw some data at it, and disable some optimizations via a global config system

Exactly what would make including this data a good idea..
I think this makes a good testcase if it is reproducible and simple like that.

Zhese being dumped somewhere would make sense.

PR #781, opened 2026-01-12
 fix(ts): Use Uint32Array for offset buffers

Turtelll:
Changes offset buffer types from `Int32Array` to `Uint32Array` throughout the TypeScript implementation. Since offsets are always non-negative values, using unsigned integers is more semantically correct and prevents potential issues with negative values.

### DIFF
--- a/ts/src/decoding/integerDecodingUtils.spec.ts
+++ b/ts/src/decoding/integerDecodingUtils.spec.ts
@@ -303,26 +303,4 @@ describe("IntegerDecodingUtils", () => {
         // The decoder is adding a 0 at the start
         expect(Array.from(decoded)).toEqual([0, 1, 2, 3, 4]);
     });
-
-    describe("Edge cases for offset buffer decoding (Uint32Array result)", () => {
-        it("should throw error for corrupted data producing negative offsets (zigzag RLE delta)", () => {
-            const corruptedData = new Int32Array([1, encodeZigZagInt32Value(-100)]);
-            const numRuns = 1;
-            const numTotalValues = 1;
-
-            expect(() => {
-                decodeZigZagRleDeltaInt32(corruptedData, numRuns, numTotalValues);
-            }).toThrow("Invalid offset buffer: negative value -100 at index 1");
-        });
-
-        it("should throw error for corrupted data producing negative offsets (RLE delta)", () => {
-            const corruptedData = new Int32Array([1, 1, 10, -20]);
-            const numRuns = 2;
-            const numTotalValues = 2;
-
-            expect(() => {
-                decodeRleDeltaInt32(corruptedData, numRuns, numTotalValues);
-            }).toThrow("Invalid offset buffer: negative value -10 at index 2");
-        });
-    });
 });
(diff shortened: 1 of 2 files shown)


ts/src/decoding/geometryDecoder.ts
```diff
@@ -39,15 +39,18 @@ export function decodeGeometryColumn(
     const geometryTypeMetadata = decodeStreamMetadata(tile, offset);
     const geometryTypesVectorType = getVectorType(geometryTypeMetadata, numFeatures, tile, offset);
 
-    let geometryOffsets: Int32Array = null;
-    let partOffsets: Int32Array = null;
-    let ringOffsets: Int32Array = null;
+    let geometryOffsets: Uint32Array = null;
```
HarelM: Can we move the definition of the variables closer to where they are assigned please?

Turtelll: Sure

ts/src/decoding/integerDecodingUtils.ts
```diff
@@ -489,15 +502,22 @@ export function decodeZigZagRleDeltaInt32(data: Int32Array, numRuns: number, num
         value = decodeZigZagInt32Value(value);
         for (let j = offset; j < offset + runLength; j++) {
             decodedValues[j] = value + previousValue;
+
+            if (decodedValues[j] < 0) {
```
HarelM: Can this really happen?
I would consider moving this to a different PR...?

ts/src/decoding/integerDecodingUtils.spec.ts
```diff
@@ -303,4 +303,28 @@ describe("IntegerDecodingUtils", () => {
         // The decoder is adding a 0 at the start
         expect(Array.from(decoded)).toEqual([0, 1, 2, 3, 4]);
     });
+
+    describe("Edge cases for offset buffer decoding (Uint32Array result)", () => {
+        it("should throw error for corrupted data producing negative offsets (zigzag RLE delta)", () => {
+            // Manually create corrupted encoded data that would produce negative offset
```
HarelM: These comments are redundant.

HarelM:
Added some minor comment, overall this looks good.

ts/src/decoding/integerDecodingUtils.ts
Turtelll: It can't happen with our encoder. The validation also adds runtime. But if tiles are encoded by something else it would be nice to have a proper error message. So I am not sure if we should keep this. What do you think ?

HarelM:
The only comment left unanswered is the following one:
https://github.com/maplibre/maplibre-tile-spec/pull/781#discussion_r2681441654

Thanks.

Turtelll:
@HarelM I did address it in my comment: "It can't happen with our encoder. The validation also adds runtime. But if tiles are encoded by something else it would be nice to have a proper error message. So I am not sure if we should keep this. What do you think ?"

HarelM:
If this is a result of a "corrupted" there might be other places that won't work as well, and since I don't think we should address all of them I prefer to remove this.

PR #783, opened 2026-01-13
feat(ts): Add fsstEncoder and refactor tests to encode-decode-compare pattern

Turtelll:
- Create fsstEncoder.ts with encodeFsst() and createSymbolTable() helper
  - Refactor fsstDecoder.spec.ts to use encode-decode-compare pattern
  - All tests now use string inputs via TextEncoder/TextDecoder

  Note: Encoder requires pre-defined symbol table. Real FSST learns optimal symbols from data. This
  implementation is for testing decoder only.


HarelM:
This is great! Thanks!
You can take a look at the coverage report in the above comment to see which decoders still needs some coverage in terms of tests.
But this is progressing nicely overall.

PR #786, opened 2026-01-18
fix(ts): decode double as float to match the behavior of the Java encoder

keiya01:
I noticed that the TS decoder decodes `double` values as `double`.

This causes a binary mismatch because the [Java encoder encodes `double` values as `float`](https://github.com/maplibre/maplibre-tile-spec/blob/b30a33c7f31899b6e0a64aa312a2df4fde9916fb/java/mlt-core/src/main/java/org/maplibre/mlt/decoder/PropertyDecoder.java#L93), while[ the C++ decoder also decodes them as `float`](https://github.com/maplibre/maplibre-tile-spec/blob/b30a33c7f31899b6e0a64aa312a2df4fde9916fb/cpp/src/mlt/decode/property.hpp#L181-L190).

To avoid this mismatch, this PR changes the handling of double values to treat them as float.

Please let me know if there is any intention.
Thank you.


HarelM:
It sounds like we should fix the encoder and not break the decoder. 
@TimSylvester, @Weixing-Zhang @louwers can you comment here, I'm not familiar with the encoder code enough...

TimSylvester:
Full double support is a version 2 item, I think, or it breaks compatibility.

HarelM:
Is this decision documented somewhere?

TimSylvester:
No, I don't think so.  There was a brief discussion at one point concluding that double support was not a priority and it was left as-is.   I assume that state was the result of @mactrem previously finding that double values are not widely used.

HarelM:
I suggest to document this somewhere. I'm also not sure what are the plans to support different version of the spec if we encode double as float.
I would expect to get an exception stating that double is not supported instead of creating weird tiles.
If not an exception, at least a warning.
This will surely bite us in the future... :-/

HarelM:
The following was recently merged, which might solve this same this:
https://github.com/maplibre/maplibre-tile-spec/pull/940

I'm not sure which version is better, I find this one simpler to read... @keiya01 what do you think? Should we revert the other one and include this?

keiya01:
@HarelM 
I think my approach is simpler if MLT plans to support double encoding in the future, since the condition can easily be reverted.

HarelM:
Yeah, I find this PR to be simpler as well.

HarelM:
This was replaced by:
- #965

Closing, sorry for the delay in merging this...

keiya01:
Thank you so much 🙏

PR #788, opened 2026-01-20
fix(ts): nullable MORTON & COMPONENTWISE_DELTA integer decoder bugs

Turtelll:
Two missing unpackNullable() calls in nullable int32 decoder caused wrong array sizes when decoding
MORTON/COMPONENTWISE_DELTA streams. A compact stream with only non-null values was returned. 
  - Before: Returns compact array [10, 20, 30] (length = 3, only non-null values)
  - After: Returns full array [10, 0, 20, 0, 30] (length = 5, with 0 at null positions from BitVector)


Turtelll:
There was a nullable test case missing so the bug was not caught. One nullable test case for DELTA was missing as well, so I added it too.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -194,10 +194,10 @@ function decodeInt32(
                 return unpackNullable(compactRle32, nullabilityBuffer, 0);
             case LogicalLevelTechnique.MORTON:
                 fastInverseDelta(values);
-                return values;
+                return unpackNullable(values, nullabilityBuffer, 0);
```
HarelM: Can you check if this whole switch case can be removed?
The reason I thought there might be a bug here is because I compared it to the switch case below.
Since both switch cases are basically doing the same and only if `nullabilityBuffer` is defined there's a call to `unpackNullable` I suggest to move the `unpackNullable` to after the switch case and remove a lot of duplicated code here...

HarelM: Please check the same for `decodeInt64` as well.

HarelM:
Can you check if more tests are needed to cover this file properly?
https://app.codecov.io/gh/maplibre/maplibre-tile-spec/pull/788/blob/ts/src/decoding/integerStreamDecoder.ts?dropdown=coverage

I think that `COMPONENTWISE_DELTA` encoder is missing as I was not sure I understood what the decoder does in order to write an encoder, so tests to cover this functionality are also missing.

ts/src/decoding/integerStreamDecoder.ts
Turtelll: Good idea.

Turtelll:
I would prefer to  make implementing encodeComponentwiseDeltaVec2  and tests for it as a separate PR.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -177,70 +177,50 @@ function decodeInt32(
...
-                    decodeZigZagInt32(values);
-                }
-                return unpackNullable(values, nullabilityBuffer, 0);
-            default:
-                throw new Error(
-                    `The specified Logical level technique is not supported: ${streamMetadata.logicalLevelTechnique1}`,
-                );
-        }
-    }
+    let useRepeatUnpack = false;
```
HarelM: Are you sure abount this repeat unpack? I havent seen it in the java code...

HarelM:
Sure, component can be a separate PR.

ts/src/decoding/integerStreamDecoder.ts
Turtelll: Yes you are right. Since nulls are removed before encoding they are not part of the delta chain. Repeating values after decoding made no sense. I fixed it now.

HarelM (approved):
Looks good now, thanks for the fixes!

PR #790, opened 2026-01-22
fix(ts): Stack overflow in Field metadata decoder for column-mapped tiles

Turtelll:
Bug
- The decoder crashed with a stack overflow error when decoding tiles with nesting/column mapping. 

Cause
- The decodeField() function in embeddedTilesetMetadataDecoder.ts was using an obsolete format that doesn't match how the encoding is done now. 
- When the Java encoder writes Field metadata for struct children, it follows this format: first a typeCode
  varint, then the field name string (if the typeCode indicates a name should be present), then the child count
   and children for complex types.
- TS still expected a old bitfield-based format

Fix
- alinges typescript decoder with java encoder format
- remove FieldOptions enum as it's not used anymore.

### DIFF
--- a/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
+++ b/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
@@ -190,7 +190,7 @@ describe("embeddedTilesetMetadataDecoder", () => {
 
                 expect(() => {
                     decodeField(buffer, new IntWrapper(0));
-                }).toThrow("Unsupported field type code: 999");
+                }).toThrow("Unsupported field type code 999. Supported: 10-29(scalars), 30(STRUCT)");
             });
         });
     });
--- a/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
+++ b/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
@@ -43,11 +43,12 @@ function columnToField(column: Column): Field {
  */
 export function decodeField(src: Uint8Array, offset: IntWrapper): Field {
     const typeCode = decodeVarintInt32(src, offset, 1)[0] >>> 0;
-    const column = decodeColumnType(typeCode);
 
-    if (typeCode <= 4) {
+    if (typeCode < 10 || typeCode > 30) {
         throw new Error(`Unsupported field type code ${typeCode}. Supported: ${SUPPORTED_FIELD_TYPES}`);
     }
+
+    const column = decodeColumnType(typeCode);
 
     if (columnTypeHasName(typeCode)) {
         column.name = decodeString(src, offset);


HarelM:
Can you add a test to cover this change somehow? Either a tile with the relevant data or a unit test that simulates what happened.

Turtelll:
I could hack together a test with a tile to test this but i would rather look to create proper integration tests that cover all encoding possibilities once fastPfor is added. The test fixtures we have aren't really good enough to do that. We should have a round trip decode-encode integration test that encoded a small tileset with all available options we have each time we run the full test suit. But I don't know if that is possible.

HarelM:
For every bug we find it is best to write a test, this way the bug will not return as the test will fail if we change the code to reintroduce the bug...

Turtelll:
@HarelM i added a tile with column mappings like name with multiple language variants (name:af, name:ar, name:az, name:be, etc.) this is exactly the nested struct/column mapping scenario that would trigger the stack overflow bug.

The test verifies:
  1. ✓ Tile decodes without throwing (no stack overflow)
  2. ✓ Metadata structure is valid (tables exist, have names, extent > 0)
  
I would still remove these kinds of tests once a sulution for general integration tests is implemented.

HarelM:
Looks like the code you have changed is not fully covered.
Can you achieve the same coverage with a simple unit test instead of a full tile?
I would keep the tiles as a last resort, and as you said, a better tile covering strategy is underway, but I don't think it will solve all of our problems, I test to think that in a lot of cases unit tests are the key to really cover some edge cases...

Turtelll:
@HarelM i removed the tile based test and added unit-tests

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+import { encodeVarintInt32Value } from "../../encoding/integerEncodingUtils";
+import { concatenateBuffers } from "../../decoding/decodingTestUtils";
+import { ComplexType, ScalarType } from "./tilesetMetadata";
+
+const STRUCT_TYPE_CODE = 30;
+
+/**
+ * Encodes a single typeCode as a varint.
+ */
+function encodeTypeCode(typeCode: number): Uint8Array {
```
HarelM: Can you move the encode functions to an encode file, much like the other encode methods?

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+
+function scalarTypeCode(scalarType: number, nullable: boolean): number {
+    return 10 + scalarType * 2 + (nullable ? 1 : 0);
+}
+
+describe("embeddedTilesetMetadataDecoder", () => {
+    describe("decodeField", () => {
+        describe("scalar fields", () => {
+            it.each([
+                {
```
HarelM: I think this is worth splitting to different tests, it doesn't save that much in terms of lines of code and it highlights the tests and improves readability.

HarelM:
I've added a few minor comments.
Thanks for looking into this!

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
Turtelll: moved them to a new file in encoding folder

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
Turtelll: I split them up again

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+
+                const field = decodeField(buffer, new IntWrapper(0));
+
+                expect(field.name).toBe("address");
+                expect(field.nullable).toBe(false);
+                expect(field.type).toBe("complexField");
+                expect(field.complexField?.physicalType).toBe(ComplexType.STRUCT);
+                expect(field.complexField?.children).toHaveLength(children.length);
+
+                children.forEach((child, i) => {
```
HarelM: Sorry for the nit-picking, can you use regular "for" loop?

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+
+                const address = field.complexField?.children[0];
+                expect(address?.name).toBe("address");
+                expect(address?.type).toBe("complexField");
+
+                const coordinates = address?.complexField?.children[0];
+                expect(coordinates?.name).toBe("coordinates");
+                expect(coordinates?.complexField?.children).toHaveLength(leafChildren.length);
+
+                leafChildren.forEach((child, i) => {
```
HarelM: Same here - avoid using forEach.

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
```diff
@@ -36,54 +22,43 @@ function decodeString(src: Uint8Array, offset: IntWrapper): string {
...
+ */
+export function decodeField(src: Uint8Array, offset: IntWrapper): Field {
+    const typeCode = decodeVarintInt32(src, offset, 1)[0] >>> 0;
+    const column = decodeColumnType(typeCode);
 
-    const field = {} as Field;
-    if ((fieldOptions & FieldOptions.nullable) !== 0) {
-        field.nullable = true;
+    if (!column) {
+        throw new Error(`Unsupported field type code: ${typeCode}`);
```
HarelM: I think this error message should be elaborated.
For example, unsupported typecode x when trying to decode field 'something' , supported types are: a, b, c.
I'm not sure what part of this is possible, but the more descriptive the error is, the more chance a user will be able to really understand what went wrong.

These error types from MVT parsing are a good example of error code that don't inform the user what went wrong or how to fix the error they encountered.

HarelM:
Almost there, added a few minor comments.

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
Turtelll: sure

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
Turtelll: Changed the typeCode check to only scalar and struct, which decodable as fields. And tried to improve the error message.

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
```diff
@@ -36,54 +25,43 @@ function decodeString(src: Uint8Array, offset: IntWrapper): string {
...
+ * Decodes a Field used as part of complex types (STRUCT children).
+ */
+export function decodeField(src: Uint8Array, offset: IntWrapper): Field {
+    const typeCode = decodeVarintInt32(src, offset, 1)[0] >>> 0;
+    const column = decodeColumnType(typeCode);
 
-    const field = {} as Field;
-    if ((fieldOptions & FieldOptions.nullable) !== 0) {
-        field.nullable = true;
+    if (typeCode <= 4) {
```
HarelM: Shouldn't this be 9 instead of 4? According to the message it only starts from 10...?

Turtelll: yes i was not finished. I need to adapt the test as well

HarelM:
I've just noticed that the coverage on this file is not full, can you add coverage for the children code path?
https://app.codecov.io/gh/maplibre/maplibre-tile-spec/pull/790/blob/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts?dropdown=coverage

HarelM:
image

PR #798, opened 2026-01-26
Add PMTiles support to the Java Encoder CLI

TimSylvester:
Adds a CLI option to convert an MVT/PBF pmtiles file (via local path or URL) to a local MLT pmtiles file.  Also adds an option to run the conversion on multiple threads.

Also removes some dead code.


java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
```diff
@@ -914,12 +1143,14 @@ private static byte[] convertTile(
...
+            }
             tileData = outputStream.toByteArray();
           } else {
             if (verboseLevel > 1) {
               System.err.printf(
-                  "Compression of %d:%d,%d not effective, saving uncompressed (%d vs %d bytes)%n",
+                  "  Compression of %d:%d,%d not effective, saving uncompressed (%d vs %d bytes)%n",
                   z, x, y, tileData.length, outputStream.size());
             }
             didCompress.setFalse();
```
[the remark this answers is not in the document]
TimSylvester: Incorrect.  This code is only reachable when `didCompress` is non-null, as it's in the else block of an `if` with that condition.

TimSylvester:
@louwers The Gradle changes are the only thing here that really needs much review.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
```diff
@@ -1649,6 +2061,17 @@ Automatic column mapping for the specified layers (Not implemented)
...
               .required(false)
               .get());
+      options.addOption(
+          Option.builder()
+              .option("j")
+              .longOpt(PARALLEL_OPTION)
+              .hasArg(true)
+              .optionalArg(true)
+              .argName("threads")
+              .desc("Enable parallel encoding of tiles")
```
louwers: This is only used for PMTiles right? Should probably be in the name.

TimSylvester: #818 adds it for the others

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
```diff
@@ -808,6 +861,304 @@ private static void encodeOfflineDB(
     }
   }
 
+  /// Encode the MVT tiles in a PMTiles file
+  private static boolean encodePMTiles(
```
louwers, line 869: This function is quite large and could probably be broken up in some smaller functions for readability.

I would suggest at least moving it to its own file because this file now has 89 lines of imports.

java/mlt-cli/build.gradle
```diff
@@ -83,13 +96,22 @@ tasks.withType(Test).configureEach {
...
-    task "$name"(type: Jar, dependsOn: [compileJava, ':mlt-core:jar']) {
+    task "$name"(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar, dependsOn: [compileJava, ':mlt-core:jar']) {
         archiveFileName = name + ".jar"
         archiveClassifier = "cli"
         duplicatesStrategy = DuplicatesStrategy.EXCLUDE
-        from {
-            configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }
+        configurations = project.configurations.named('runtimeClasspath').map { [it] }
+
+        // Filter out dependency JAR signatures to prevent `SecurityException: Invalid signature file digest for Manifest main attributes`
```
louwers, line 105: I have no idea what this is but I believe you.

TimSylvester: I don't really understand what's going on there, but the previous didn't work once I switched to ShadowJar.

That's just what I could get to work, based on copying examples from  https://gradleup.com/shadow/configuration/dependencies/

louwers: > I don't really understand what's going on there

Average Gradle experience. If it works it works.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
TimSylvester: Cleanup in #818 ok?

PR #801, opened 2026-01-27
Upload Java coverage

louwers:
Creating a branch to test it.


.github/workflows/java.yml
```diff
@@ -44,3 +50,10 @@ jobs:
         working-directory: java/encoding-server
       - run: npm run checkstyle
         working-directory: java/encoding-server
+
+      - name: Upload coverage to Codecov
+        if: github.repository_owner == 'maplibre'
```
HarelM, line 55: Is this required? Worst case it will upload for a different repo, right? what's the problem with that?

louwers: Yes it is required because otherwise it will run from PRs on forks and fail the workflow.

HarelM: OK, I tend to think this is solvable for people who would like to see coverage on their forks, and probably other CI will fails as I don't think it's configured this way for all the CIs, but I don't have strong feelings either way...
Feel free to keep it.

louwers: It will fail even for PRs that target this repository I think.

PR #810, opened 2026-02-03
fix(ts): shared-dictionary STRUCT columns silently dropped during decoding

Turtelll:
The decoder silently dropped all struct columns due to a wrong numStreams check in both mlt and property decoder. 

This PR fixes the issues and adds tests to cover the bug.

Note: 
Encoder strips longest common prefix from sibling names (name, name:en, name:de → parent name, children "", ":en", ":de"). Empty child name or "default" maps to parent name; colon-prefixed names are concatenated directly; otherwise separator is inserted.


TimSylvester:
> ROOT_COLUMN_NAME ... NESTED_COLUMN_SEPARATOR

Is this mandating a specific separator?  The intent is that the column names are recovered through simple concatenation. Any separator, or no separator at all, is equally supported, and the "default" naming is obsolete.

PR #813, opened 2026-02-04
feat(cpp): implement MLT encoder

dannote:
# C++ MLT encoder

Inverse of the C++ decoder from #441. Encodes `Encoder::Layer` → MLT binary, wire-compatible with the Java and TypeScript encoders.

## 🤖 AI slop disclaimer

This was written with Claude Opus 4.5, with me steering the whole way. As an open-source maintainer myself, I'm painfully aware of the current wave of low-effort AI PRs on GitHub — mass-filed, poorly tested, and dumped on maintainers to review.

Here's what I did to make sure this isn't that:

- **Not a Java port.** The encoder matches the existing C++ decoder idioms: C++20 templates with `requires`, `span<const T>`, PIMPL, anonymous namespaces. I read the Java encoder as a reference for _what_ to encode, not _how_ to write it.
- **378 new tests.** 60 unit/integration tests + 6 parameterized corpus suites across 128 real-world fixtures (OMT, Bing Maps, Amazon). Every test was written to catch a real bug or validate a specific wire format contract, not to pad coverage.
- **Cross-validated against Java.** 6 fixtures encoded by the Java encoder are decoded, re-encoded by C++, decoded again, and compared field-by-field. 6 byte-comparison tests verify C++ output ≤ Java output size with semantic equivalence.
- **Found and fixed 3 bugs in the existing decoder.** The encoder doesn't just produce correct output — it exercises decoder paths that were previously untested.
- **Static analysis.** clang-tidy, cppcheck, ASan + UBSan — all clean on the full test suite.
- **Explicit anti-slop editing pass.** Removed comments that restate what code does, dead code, unused includes, near-duplicate functions. Comments that remain document wire format specs, non-obvious constraints, or algorithm provenance.

### Core encoder (`encoder.hpp` / `encoder.cpp`)

Public API:

```cpp
mlt::Encoder encoder;
auto bytes = encoder.encode(layers, config);
```

`Encoder::Layer` holds features with typed properties (`bool`, `int32_t`, `int64_t`, `uint32_t`, `uint64_t`, `float`, `double`, `string`, `StructValue`), geometry (coordinates + ring sizes + multi-part structure), and optional IDs.

### Encoding primitives

- **varint** — unsigned LEB128 encode
- **zigzag** — signed → unsigned for smaller varints on negative values
- **byte RLE** — ORC-style for boolean present streams
- **integer RLE** — run-length for topology streams (geometry types, ring counts)

### Component encoders

| Encoder | Strategy |
|---------|----------|
| `IntegerEncoder` | Tries plain, delta, RLE, delta+RLE — picks smallest (BTR-Blocks heuristic) |
| `BooleanEncoder` | Bitpack + byte RLE |
| `FloatEncoder` | Raw IEEE 754 LE (matches wire format — doubles stored as floats) |
| `StringEncoder` | Tries plain, dictionary, FSST dictionary — picks smallest |
| `GeometryEncoder` | Tries plain, Hilbert dictionary, Morton dictionary vertex buffers — picks smallest |
| `PropertyEncoder` | Dispatches per scalar type, handles nullable present streams |

### Advanced features

- **FSST string compression** — vendors [cwida/fsst](https://github.com/cwida/fsst) (MIT). Thin adapter extracts symbol table into MLT wire format. StringEncoder considers it as a third candidate alongside plain and dictionary.
- **Hilbert curve** — Skilling algorithm, cross-validated against Java `org.davidmoten.hilbert` (38 test vectors, bits 2–14).
- **Vertex dictionary** — deduplicated vertices sorted by Hilbert index with delta-encoded offsets. Morton dictionary variant uses `MortonEncodedStreamMetadata`. Encoder tries all three and picks the smallest.
- **Feature sorting** — points sorted by Hilbert index, linestrings by first-vertex Hilbert index. Enabled by default, only on homogeneous layers.
- **Struct columns** — `StructValue` (map of string→string) auto-discovers child fields, encodes as STRUCT with shared dictionary (plain or FSST).
- **Pre-tessellation** — vendors [mapbox/earcut.hpp](https://github.com/mapbox/earcut.hpp) (ISC). Triangulates polygons, encodes triangle counts + index buffer as additional streams. Only activates when all features are polygons/multipolygons. Outlines always included.

### Metadata encoding

`StreamMetadata`, `RleEncodedStreamMetadata`, and `MortonEncodedStreamMetadata` got `encode()` methods (inverse of existing `decode()`). `encodeFeatureTable()` added to `tileset.hpp` / `tileset.cpp`. Column type codes use the existing `type_map.hpp` (with a bug fix — nullable/longID bits were swapped).

## Decoder bugs found and fixed

Building the encoder exercised decoder paths that existing tests didn't cover:

1. **Polygon ring closing points** (`geometry_vector.cpp`): the plain vertex buffer path had closing-point logic commented out with a TODO. The dictionary and Morton paths added closing points correctly. Re-enabled it — was causing ring vertex count mismatches when the encoder selected plain over dictionary.

2. **Space-filling curve bit width** (`space_filling_curve.hpp`): `ceil(log2(tileExtent))` → `ceil(log2(tileExtent + 1))`. With extent 4096, the old formula gave numBits=12 (grid 0..4095), mapping coordinate 4096 to the same index as 0.

3. **INT_64 property decoding** (`decode/property.hpp`): `decodeIntStream` was instantiated with `<uint64_t, uint64_t, uint64_t>` for INT_64 columns. The zigzag step produces signed values, but they were stored as `uint64_t`. Fixed to `<uint64_t, uint64_t, int64_t>`.

Also fixed shadow variables in `geometry_vector.cpp` (inner loops reusing `i`/`j` from outer scope — no runtime bug, but warned by `-Wshadow`).

## Test suite

378 new tests (396 total with pre-existing decoder tests) across 13 new suites:

| Suite | Count | What it tests |
|-------|-------|---------------|
| `EncodePrimitives` | 2 | Zigzag, varint roundtrip |
| `EncodeMetadata` | 3 | StreamMetadata, RLE metadata, FeatureTable |
| `Encode` | 32 | All geometry types, property types, nullable variants, edge cases (empty layers, boundary coords, max uint64 IDs, 10k-char strings, UTF-8, degenerate polygons, 10k features), FSST strings, vertex dictionary, feature sorting, struct columns, pre-tessellation |
| `CrossValidate` | 13 | Java fixture decode → C++ re-encode → decode → compare |
| `ByteCompare` | 6 | C++ output ≤ Java output size, semantic equivalence |
| `FSST` | 2 | cwida/fsst adapter encode/decode roundtrip |
| `HilbertCurve` | 2 | 38 Java cross-validation vectors, SFC roundtrip |
| `ReencodeOMT` | 95 | OpenMapTiles corpus roundtrip |
| `ReencodeBing` | 17 | Bing Maps corpus roundtrip |
| `ReencodeAmazon` | 11 | Amazon corpus roundtrip |
| `ReencodeAmazonHere` | 5 | Amazon HERE corpus roundtrip |
| `ReencodeOMTSorted` | 95 | OMT corpus with feature sorting enabled |
| `ReencodeOMTTessellated` | 95 | OMT corpus with pre-tessellation enabled |

Corpus tests decode a Java-produced `.mlt` fixture, convert to `Encoder::Layer`, re-encode, decode again, and compare geometry and properties per feature. Duplicate IDs handled via fingerprint-based set comparison.

## Design decisions

- **Encoding selection matches Java heuristic**: each encoder tries multiple strategies and picks the smallest output. The integer encoder uses the BTR-Blocks-inspired threshold.
- **`encode()` on existing metadata classes** rather than separate encoder metadata types. Keeps the encode/decode symmetry visible.
- **MLT stores polygon rings without closing points**: Java's `ringToLineString` strips them, decoder adds them back. The encoder follows the same contract.
- **FSST: vendor cwida/fsst, not port Java FSST**: the original C library has AVX-512 support and is battle-tested. The wire format is compatible — cwida's decoder stores symbols in length-group order matching Java's `sortSymbolsByLength`.
- **INT32_MAX as long-ID threshold**: delta encoding accumulates in `int32_t`. OSM IDs commonly exceed INT32_MAX but fit in `uint32_t` — without the threshold, they sign-extend to wrong values on zigzag decode.
- **Pre-tessellation always encodes outlines**: the 7-stream format (vs 4-stream without outlines) supports renderer-side polygon clipping.

## Stats

- 5,021 lines added, 32 deleted across 29 files
- 2 new git submodules (cwida/fsst, mapbox/earcut.hpp)
- Encoder source: ~2,500 lines
- Test code: ~2,400 lines


HarelM:
I'm not sure I'll be really the one reviewing this, but as a reference, I asked the contributors for the typescript code to split large PRs into smaller PRs for easier review. 
The benefit is that if there are repeated request for changes you can apply them to "future" PRs and avoid the churn of following a lot of code review comments. 
You can start by introducing the primitive encoders and their unit tests, and slowly build on that from PR to PR.
This helps me review the fastPFor typescript implementation, for example. 
But @louwers and @TimSylvester should probably decide what would work better for them. 

Thanks for taking the time to contribute!!

dannote:
Good idea, agree! Happy to split it — e.g. decoder bug fixes → encoding primitives → component encoders → metadata & top-level encoder, each with its own tests. Open to other ways to slice it too.

@louwers @TimSylvester — whatever works best for you. I'll keep this PR open for a while as a reference for the full picture.

cpp/CMakeLists.txt
```diff
@@ -135,6 +149,18 @@ target_sources(mlt-cpp PRIVATE
 # SIMD Everywhere (needed on ARM)
 target_include_directories(mlt-cpp PRIVATE "${PROJECT_SOURCE_DIR}/vendor/simde")
 
+# FSST (Fast Static Symbol Table) — vendored from https://github.com/cwida/fsst (MIT)
```
louwers: Please create fsst.cmake and earcut.cmake files in `vendor`.

I think you can use `add_subdirectory` and use the CMake targets directly.

cpp/test/test_encode.cpp
```diff
@@ -0,0 +1,2427 @@
...
+                << "property " << name << " presence mismatch for id=" << fa.getID();
+            if (valA.has_value() && valB.has_value()) {
+                EXPECT_EQ(valA->index(), valB->index())
+                    << "property " << name << " type mismatch for id=" << fa.getID();
+            }
+        }
+    }
+}
+
+TEST(CrossValidate, JavaPointBoolean) {
```
louwers: There is a lot of duplication here.

Also I don't think we should skip tests if fixtures cannot be found.

louwers (commented):
Thanks for the heads-up on AI usage.

The bug fixes for the decoder would be good to have in a separate PR. The encoder can come in a single PR as far as I am concerned.

I would like to see a separate library and CMake target for the encoder, since it is a completely different use case and I don't want to rely on dead code elimination to make sure we don't ship it with MapLibre Native builds.

dannote:
@louwers thank you for your swift review! Will follow your advices.

cpp/src/mlt/geometry_vector.cpp
```diff
@@ -31,11 +31,9 @@ std::vector<Coordinate> getLineStringCoords(const std::vector<std::int32_t>& ver
         coords.push_back(coord(x, vertexBuffer[startIndex++]));
     }
 
-    // This breaks embedded tessellation because it changes the indexes.
-    // Disabling it doesn't seem to break anything, is it needed?
-    // if (closeLineString && !coords.empty() && coords.front() != coords.back()) {
-    //     coords.push_back(coords.front());
-    // }
+    if (closeLineString && !coords.empty() && coords.front() != coords.back()) {
```
TimSylvester, line 52: I had forgotten about this... is the issue with tessellation resolved?

This only shows up if the encoder does not include closing points, but they're redundant, so it should not include them.

I think the right solution is for encoders to not encode closing points, but include tessellation indexes based on those closing points being present.  Based on this, the Java encoder, at least, isn't doing that.

dannote: Yes — both encoders strip closing points before writing vertices, and tessellation indices reference those same stripped vertices. The decoder adds closing points back on read. This PR re-enables that for the plain path, which was the only one missing it.

sjg-wdw:
This is really interesting.  We'd love to see a C++ encoder for Postgres and Tilemaker, among others.
We'll be discussing in the MapLibre Native meeting next week.  Might be able to free up a little of Tim's time to help.  Depends on how the maintainers feel about it.

louwers:
We'll also have the MLT call on Monday.

If anyone wants an invite, please message team@maplibre.org. Thanks!

dannote:
@louwers  
 
> If anyone wants an invite, please message team@maplibre.org. Thanks!

Sent a email.

louwers:
Sent you an invite!

dannote:
Made another code-deduplication round. Will do some more later.

dannote:
Update: I've built a PostGIS `ST_AsMLT` aggregate function as a proof of concept, modeled after the existing `ST_AsMVT`. It vendors the encoder from this PR into `deps/mlt/` with a C bridge (`lwgeom_mlt.cpp`) that converts LWGEOM geometries into `mlt::Encoder::Geometry` structs.

The fork is at [dannote/postgis@feat/mlt-support](https://github.com/dannote/postgis/tree/feat/mlt-support) — API is the same shape as MVT:

```sql
SELECT ST_AsMLT(q, 'land', 4096, 'geom', 'id')
FROM (SELECT id, geometry AS geom FROM lands WHERE ...) q;
```

Validated on a production dataset (4.1M land parcels, 428K buildings, 289K zones). Some numbers from real tiles:

| Metric | Value |
| --- | --- |
| Densest z14 tile (2,635 polygons) | 24 KB MLT vs 198 KB MVT (8.1×) |
| z12 tile (3,252 polygons) | 40 KB MLT, 50 ms |
| Multi-layer zone tile (29 layers, 1,193 features) | 12.6 KB, 167 ms |
| Fuzz test | 134,430 tiles, 61K non-empty, zero errors |

Encoding is ~10× slower than MVT but the output is 2–8× smaller, so the tradeoff works well for tiles served behind a caching proxy.

dannote:
Also sharing the other integrations I've built using this encoder:

- **[dannote/tippecanoe@feature/mlt-encoder](https://github.com/dannote/tippecanoe/tree/feature/mlt-encoder)** — `tippecanoe -o output.mlt.pmtiles` generates MLT tilesets. Supports nested JSON properties as STRUCT columns.
- **[dannote/duckdb-spatial@feat/mlt-support](https://github.com/dannote/duckdb-spatial/tree/feat/mlt-support)** — `ST_AsMLT` aggregate + `ST_ReadMLT` table function + `.mlt` replacement scan for DuckDB.

louwers:
Now you're just showing off. Awesome work! 🔥

louwers:
If you want to calculate coverage, [this](https://github.com/maplibre/maplibre-tile-spec/pull/821) should be merged soon.

Edit: You can generate coverage with

```
just coverage-cpp
```

This will create a `build/coverage.txt`.

dannote:
Added 23 tests to improve encoder coverage — type promotion paths (INT_32→INT_64, FLOAT→DOUBLE, mixed→STRING), nullable variants for all property types, int64 RLE/delta-RLE, Morton metadata roundtrip, mixed geometry layers, sparse struct columns.

Also wired `includeOutlines`, `useMortonEncoding`, and `useFsst` through `EncoderConfig` to component encoders (previously hardcoded), so tests can exercise each encoding strategy selection path.

Remaining uncovered lines that aren't practically testable:

- `encodeFastPfor` throw on line 66-68 — only reachable when compiled without `MLT_WITH_FASTPFOR`, but in that case the function body is `#else throw`, which the coverage build can't enter since it compiles with FastPFOR enabled
- `(auto)` fallback lambdas in `extractColumn` visitors (lines 481, 488-489, 499-500, etc.) — `std::visit` requires exhaustive overloads, but the type is already determined by column metadata, so the catch-all arms are dead by construction
- `default: throw` in the scalar type switch (line 565) — same reason, all enum values are handled
- `StructValue` arm in scalar type visitor (line 103) — `StructValue` properties are routed to `structColumns` before reaching the scalar visitor

louwers:
@systemed Did you ever try this out?

systemed:
> @systemed Did you ever try this out?

I haven't yet had the chance to do so unfortunately, but hoping to get the time soon!

TimSylvester:
The C++ encoder is now producing bit-identical output for 325 of the synthetic test cases, and JSON-equivalent results for the 31 others which have expected differences from FSST or tessellation.  Getting them to work turned up several minor bugs, along with some improvements that should be made in the Java encoder.  I also implemented the buffer handling in #1065, which was done after this was ported.

Ideally the separate Rust synthetics would also be checked, but I don't think that's necessary yet.

@louwers Do you want to review it?

louwers:
@TimSylvester Thanks Tim! I will.

@dannote If you're still around, please have a look at Tim's changes as well.

@systemed If you still want to give this a spin before we merge it, now is the time. 😉

nyurik:
you may want to follow the same pattern as in rust:
* after generation, make sure you can parse it and get json-identical results
* if bin generated data is bit-identical to java, validate that java's json version is the same
* if bit-different, save it to `0x01-cpp/...` with the same name as in java, and also dup the json file. Make sure json file is bit-identical to the one in `0x01/`
* if generating something that java cannot produce, save it to `0x01-cpp/...-cpp` to indicate that this is a cpp-unique

TimSylvester:
I missed a bunch of places with the buffer handling change.

It looks like at least some of the need to force specific encodings in order to match is because Java is comparing the raw streams without metadata to select encodings.  That needs to be addressed separately, though.

louwers (approved):

brawer:
Curious, what’s needed to get this merged?

louwers:
@brawer From my side, we just need to get the CI passing (again).

PR #814, opened 2026-02-04
fix(ts): shared dictionary child column names double-seperator

Turtelll:
decodeSharedDictionary inserted a : separator between parent column name and child field name. The decoder was producing name::en / :highway instead of name:en / highway. Which caused GL JS to not filter for properties and double delimiter which were requested from the style.json.

So basically what @TimSylvester said in [PR#810](https://github.com/maplibre/maplibre-tile-spec/pull/810)


Turtelll:
only a part of it

PR #817, opened 2026-02-04
feat(cpp): enable FastPFOR decoding in MLT

TimSylvester:
Replaces #3954

### DIFF
--- a/cpp/mod.just
+++ b/cpp/mod.just
@@ -71,9 +71,10 @@ test-json:
 
 decode-synthetics: cmake-build (just 'cargo-install' 'fd' 'fd-find')
     #!/usr/bin/env bash
+    set -e
     JSON_TOOL=build/tool/mlt-cpp-json
-    fd . ../test/synthetic/0x01 -e mlt -x bash -c "echo {} ; $JSON_TOOL {} >/dev/null || true"
-    fd . ../test/synthetic/0x01-rust -e mlt -x bash -c "echo {} ; $JSON_TOOL {} >/dev/null || true"
+    fd . ../test/synthetic/0x01 -e mlt -x bash -c "echo {} ; $JSON_TOOL {} >/dev/null"
+    fd . ../test/synthetic/0x01-rust -e mlt -x bash -c "echo {} ; $JSON_TOOL {} >/dev/null"
 
 _check-tool command:
     @which {{command}} > /dev/null || (echo "Error: {{command}} is not installed. Please install it using: brew install {{command}} (macOS) or pip3 install {{command}} (Linux)" && exit 1)


cpp/vendor/fastpfor.cmake
louwers: This is a bug I think.

TimSylvester: What's the problem, the path?

louwers: `MLT_ENABLE_FASTPFOR` should be `MLT_WITH_FASTPFOR`

TimSylvester: Ah, that sounds familiar.  So it hasn't even been using it 😬🫣

nyurik:
We may have to resolve #1016 upstream first - I seriously doubt anyone can rely on `native` compilation result

louwers (commented):
Note: FastPFOR is not a safe library...

> It is a low-level library. The inputs are not validated prior to decoding. It is assumed that the provided buffers were generated by the corresponding encoding functions. It is your responsability to ensure the validity of your data. In production systems, the use a strong hashing function to check for data integrity is recommended. The library does not provide this functionality.


We should probably not enable it by default in MapLibre Native.

Can you add a flag to the encoder to enable or disable FastPFor at runtime?

TimSylvester:
The Java CLI encoder already requires FastPFOR to be explicitly enabled, same with FSST.

Or do you mean the decoder?  That would be easy enough, but I'm not sure how it would be controlled.

louwers:
@TimSylvester Yes I meant the decoder.

We'd add platform APIs to turn it on so it is safe by default.

Or we need to validate the incoming data, but I'm not sure how involved that would be.

TimSylvester:
~The C++ FastPFOR path isn't really being tested here.  I added decoding of the synthetic MLTs via the JSON tool in #1031, but a bunch of them fail because they haven't been run so far, and the output isn't comparable with what's checked in.~

Hm, I forgot about #885, so I guess it is.  But I don't know why it was failing when I ran it.

TimSylvester:
@louwers Would you expect that configuration to be per-map-instance, per-source, or global/static?

For now I wired it up to the most direct spot, which is the vector tile source.

louwers:
@TimSylvester Do you think FastPFor it has the potential to just crash, or cause UB? (cc @CommanderStorm)

TimSylvester:
@louwers  Is the distinction between a crash and UB important?

I take that warning to mean that crafted input could cause buffer overflows and thus RCEs.  Simply corrupted/broken.

I don't think that the necessary buffer checks are practical outside the library, as they would effectively require decoding the data.  And I don't think a hash is useful here.

If we're concerned about safely loading un-trusted data, we should probably try to contribute that to the upstream, leaving it disabled by default in the meantime.

PR #818, opened 2026-02-05
feat(java): add parallel conversion for mbtiles and offline database cases in CLI encoder

TimSylvester:
Extends parallel processing to MBTiles and offline database operations.

Splits up and refactors the excessively-large CLI Encoder class.


java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
```diff
@@ -40,4 +73,273 @@ private static Map<String, Object> toJSON(Feature feature) {
...
+        try (final var outputStream = new ByteArrayOutputStream()) {
+          decompressInputStream.transferTo(outputStream);
+          return outputStream.toByteArray();
+        }
+      }
+    } catch (IndexOutOfBoundsException | IOException ex) {
+      System.err.printf("Failed to decompress data: %s%n", ex.getMessage());
+    }
+
+    return srcStream.readAllBytes();
```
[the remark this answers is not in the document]
TimSylvester: already documented

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/PMTilesHelper.java
```diff
@@ -0,0 +1,359 @@
...
+                state.totalTileCount.get());
+          }
+        }
+
+        final var maybeTile = reader.getTile(tileIndex);
+        if (!maybeTile.isPresent()) {
+          if (state.verboseLevel > 1) {
+            System.err.printf("%s :  No tile data present%n", tileLabel);
+          }
+          return false;
```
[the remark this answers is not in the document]
TimSylvester: This state is checked before the call to this method, in `processZoom`, so isn't needed here.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/MBTilesHelper.java
```diff
@@ -0,0 +1,195 @@
...
+                  tile.getZoom(), tile.getColumn(), tile.getRow(), ex.getMessage());
+              if (config.verboseLevel() > 1) {
+                ex.printStackTrace(System.err);
+              }
+            }
+          }
+
+          if (config.threadPool() != null) {
+            config.threadPool().close();
+          }
```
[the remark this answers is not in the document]
TimSylvester: `close()` calls `awaitTermination()`

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/OfflineDBHelper.java
```diff
@@ -0,0 +1,195 @@
...
+                    if (config.verboseLevel() > 1) {
+                      ex.printStackTrace(System.err);
+                    }
+                  }
+                }
+              });
+        }
+        if (config.threadPool() != null) {
+          config.threadPool().close();
+        }
```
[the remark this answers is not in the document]
TimSylvester: `close()` calls `awaitTermination()`

TimSylvester:
@louwers @nyurik There's more I want to do on this, but I think it's time to get this one merged.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
```diff
@@ -106,4 +133,269 @@ private static Object intifyCoordinates(Object obj) {
     }
     return obj;
   }
+
+  public static String printMVT(MapboxVectorTile mvTile) {
```
louwers: The idiomatic Java way to do this is to implement `toString()` I think.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
```diff
@@ -1312,105 +484,22 @@ private static byte[] convertTile(
...
-    if (Objects.equals(a, b)) {
-      return true;
-    }
-    if (a instanceof Double && b instanceof Float) {
-      // We currently encode doubles as floats
-      return ((Double) a).floatValue() == (Float) b;
-    }
-    // Allow for, e.g., int32 and int64 representations of the same number by comparing strings
-    return a.toString().equals(b.toString());
+  private static void printColumnMappings(
```
louwers: This should also be a `toString()` I think.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
```diff
@@ -106,4 +133,269 @@ private static Object intifyCoordinates(Object obj) {
...
+      } catch (IOException ex) {
+        System.err.println("Failed to create directory: " + path);
+        ex.printStackTrace(System.err);
+      }
+    }
+  }
+
+  ///  Compare the content of the given MLT and MVT tiles and throw a RuntimeException if they
+  // differ
+  public static void compare(
```
louwers: This should not be called `compare` I think because it's just a gigantic assert.

I think it should be split up:

- It should exist in its own file since it does not really have anything to do with the CLI. It's just used by the CLI.
- The `compareGeom` part can be its own function.
- The `compareProp` part can be its own function.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
```diff
@@ -106,4 +133,269 @@ private static Object intifyCoordinates(Object obj) {
...
     return obj;
   }
+
+  public static String printMVT(MapboxVectorTile mvTile) {
+    final var gson = new GsonBuilder().setPrettyPrinting().create();
+    return gson.toJson(Map.of("layers", mvTile.layers().stream().map(CliUtil::toJSON).toList()));
+  }
+
+  ///  Execute the given task either directly or on the given thread pool
+  public static void runTask(@Nullable ExecutorService threadPool, @NotNull Runnable task) {
```
louwers: Instead of sticking this on the `CliUtil` class, consider creating a little helper class that can either be constructed with a threadPool, or without one.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
```diff
@@ -106,4 +133,269 @@ private static Object intifyCoordinates(Object obj) {
...
+      }
+
+      if (decompressInputStream != null) {
+        try (final var outputStream = new ByteArrayOutputStream()) {
+          decompressInputStream.transferTo(outputStream);
+          return outputStream.toByteArray();
+        }
+      }
+    } catch (IndexOutOfBoundsException | IOException ex) {
+      System.err.printf("Failed to decompress data: %s%n", ex.getMessage());
```
louwers: Is this really an error we can (and should) recover from?

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
```diff
@@ -106,4 +133,269 @@ private static Object intifyCoordinates(Object obj) {
...
+
+  ///  Execute the given task either directly or on the given thread pool
+  public static <T> Future<T> runTask(
+      @Nullable ExecutorService threadPool, @NotNull Supplier<T> task) {
+    return (threadPool != null)
+        ? threadPool.submit(() -> task.get())
+        : CompletableFuture.completedFuture(task.get());
+  }
+
+  public static byte[] decompress(InputStream srcStream) throws IOException {
```
louwers: The `decompress` and `compress` functions do not need to be added to the `CliUtil` class I think.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
```diff
@@ -106,4 +133,269 @@ private static Object intifyCoordinates(Object obj) {
...
+      var parameters = new DeflateParameters();
+      parameters.setCompressionLevel(9);
+      parameters.setWithZlibHeader(false);
+      return new DeflateCompressorOutputStream(src);
+    }
+    return src;
+  }
+
+  /// Returns the values that are in set a but not in set b
+  static <T> Set<T> getAsymmetricSetDiff(Set<T> a, Set<T> b) {
```
louwers: This utility helper should exist in its own file.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
```diff
@@ -106,4 +133,269 @@ private static Object intifyCoordinates(Object obj) {
...
+    Set<T> diff = new HashSet<>(a);
+    diff.removeAll(b);
+    return diff;
+  }
+
+  static void createDir(Path path) {
```
louwers: This utility helper should exist in its own file.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/MBTilesHelper.java
```diff
@@ -0,0 +1,195 @@
...
+          if (outputPath == null) {
+            dbFile.deleteOnExit();
+          }
+          mbTilesWriter = null;
+
+          // mbtiles4j doesn't support types other than png and jpg,
+          // so we have to set the format metadata the hard way.
+          final var connectionString = "jdbc:sqlite:" + dbFile.getAbsolutePath();
+          try (var connection = DriverManager.getConnection(connectionString)) {
+            if (config.verboseLevel() > 0) {
```
louwers: I think this codebase has outgrown `System.err` and some kind of logger is needed.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/MBTilesHelper.java
```diff
@@ -0,0 +1,195 @@
...
+                }
+                continue;
+              }
+
+              final var data = tile.getData().readAllBytes();
+
+              final var writerCapture = mbTilesWriter;
+              CliUtil.runTask(
+                  config.threadPool(),
+                  () -> {
```
louwers: To make this function slightly more readable I think this lambda should be refactored into a static method.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/MBTilesHelper.java
```diff
@@ -0,0 +1,195 @@
...
+
+          final var dbFile = mbTilesWriter.close();
+          if (outputPath == null) {
+            dbFile.deleteOnExit();
+          }
+          mbTilesWriter = null;
+
+          // mbtiles4j doesn't support types other than png and jpg,
+          // so we have to set the format metadata the hard way.
+          final var connectionString = "jdbc:sqlite:" + dbFile.getAbsolutePath();
```
louwers, line 128: Lines 139-168 should be refactored into a method.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/OfflineDBHelper.java
```diff
@@ -0,0 +1,195 @@
...
+
+          if (config.verboseLevel() > 1) {
+            System.err.printf(
+                "Updated source JSON format to '%s'%n", MBTilesHelper.MetadataMIMEType);
+          }
+        }
+      }
+
+      if (config.verboseLevel() > 1) {
+        System.err.println("Optimizing database");
```
louwers: Duplicated with `MBTilesHelper.java`. Should be extracted into a shared helper.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/OfflineDBHelper.java
```diff
@@ -0,0 +1,195 @@
...
+        if (config.threadPool() != null) {
+          config.threadPool().close();
+        }
+      }
+
+      if (!config.continueOnError() && !success.get()) {
+        return false;
+      }
+
+      // Update metadata
```
louwers: All this can be pulled into a separate method.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/OfflineDBHelper.java
```diff
@@ -0,0 +1,195 @@
...
+
+                byte[] srcTileData;
+                try {
+                  srcTileData = CliUtil.decompress(new ByteArrayInputStream(data));
+                } catch (IOException | IllegalStateException ex) {
+                  success.set(false);
+                  System.err.printf(
+                      "ERROR: Failed to decompress tile '%d': %s%n", uniqueID, ex.getMessage());
+                  if (config.verboseLevel() > 1) {
+                    ex.printStackTrace(System.err);
```
louwers: This log + optionally print stack trace is duplicated 4 times in this file and 3 times in `MBTilesHelper`. Should be put into the logger suggested in the other comment.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/OfflineDBHelper.java
```diff
@@ -0,0 +1,195 @@
...
+            System.err.printf(
+                "Updated source JSON format to '%s'%n", MBTilesHelper.MetadataMIMEType);
+          }
+        }
+      }
+
+      if (config.verboseLevel() > 1) {
+        System.err.println("Optimizing database");
+      }
+      try (var statement = dstConnection.prepareStatement("VACUUM")) {
```
louwers: It's not needed to use `prepareStatement` here because it has no parameters.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/OfflineDBHelper.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+      }
+
+      if (!config.continueOnError() && !success.get()) {
+        return false;
+      }
+
+      // Update metadata
+      var metadataKind = 2; // `mbgl::Resource::Kind::Source`
+      var metadataQuerySQL = "SELECT id,data FROM resources WHERE kind = " + metadataKind;
```
louwers: Nit: could use prepared statement instead of string concatenation.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/PMTilesHelper.java
```diff
@@ -0,0 +1,309 @@
...
+        throw new IllegalArgumentException("Unsupported PMTiles compression type: " + compression);
+      }
+    };
+  }
+
+  private static ReadablePmtiles tryCreateReadablePmtiles(RangeReader reader) {
+    try {
+      return new ReadablePmtiles(reader.asByteChannel());
+    } catch (IOException e) {
+      return null;
```
louwers: Maybe return an `Optional`? Not sure what best practice is in Java.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/PMTilesHelper.java
```diff
@@ -0,0 +1,309 @@
...
+        return false;
+      }
+
+      if (state.encodeConfig().verboseLevel() > 0) {
+        final var extra =
+            (rawSize != tileData.length) ? String.format(", %d compressed", rawSize) : "";
+        System.err.printf("%s : Loaded (%d bytes%s)%n", tileLabel, tileData.length, extra);
+      }
+
+      final MutableBoolean didCompress = null; // force compression
```
louwers: I think this API where you pass a `null` to force compression is a bit smelly.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
TimSylvester: That's in core, though, and this output format is not necessarily universal.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
TimSylvester: Hm, probably not...

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/MBTilesHelper.java
TimSylvester: Maybe agree, though I hate every logging framework.  Nevertheless, I think that should be a PR of its own.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
louwers: Fair point. Maybe you can move `printMVT` and related utilities into a separate class. It would clean up `CliUtil` a bit.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
TimSylvester: I don't think the geometry and properties make sense as separate (public) methods because they only apply in the context of already having matched up individual features within layers.  I have refactored the overall functionality into a helper class in core, though.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/MBTilesHelper.java
TimSylvester: #899

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
TimSylvester: JSON/GeoJSON support moved to `JsonHelper`.  Not moved to core because GeoJSON support is an extra dependency.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/Encode.java
TimSylvester: I refactored one `printColumnMappings`, but the other doesn't seem to make sense as a `toString()` on any of the classes involved.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/CliUtil.java
TimSylvester: Moved to comparison helper, that's the only place it's used.

java/mlt-cli/src/main/java/org/maplibre/mlt/cli/OfflineDBHelper.java
TimSylvester: #899

java/mlt-core/src/test/java/org/maplibre/mlt/MltGenerator.java
```diff
@@ -193,13 +197,13 @@ private byte[] convertMvtToMlt(
       throws IOException {
     var config =
         new ConversionConfig(
```
louwers: Use builder?

java/mlt-core/src/main/java/org/maplibre/mlt/compare/CompareHelper.java
```diff
@@ -0,0 +1,368 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.locationtech.jts.geom.Geometry;
+import org.maplibre.mlt.converter.mvt.MapboxVectorTile;
+import org.maplibre.mlt.data.Feature;
+import org.maplibre.mlt.data.Layer;
+import org.maplibre.mlt.data.MapLibreTile;
+
+public class CompareHelper {
```
louwers, line 20: Isn't this just a test utility?

Shouldn't it live in a different package?

java/mlt-core/src/main/java/org/maplibre/mlt/converter/MltConverter.java
```diff
@@ -19,25 +19,74 @@
...
-      boolean isIdPresent) {
-    return createTilesetMetadata(tile, columnMappings, isIdPresent, false, false);
+      MapboxVectorTile tile, ColumnMappingConfig columnMappingConfig, boolean includeIdIfPresent) {
+    return createTilesetMetadata(tile, null, columnMappingConfig, includeIdIfPresent);
   }
 
+  /// Create tileset metadata from a MVT tile, with optional column mapping configuration and type
+  // mismatch policy.
+  /// @param tile The MVT tile to create metadata from
+  /// @param config Optional configuration
```
louwers: This parameter doesn't exist.

java/mlt-core/src/main/java/org/maplibre/mlt/compare/CompareHelper.java
TimSylvester: It's used in tests and CLI, not sure what the right place is.  Theoretically useful for an API user for validation, I suppose.

java/mlt-core/src/main/java/org/maplibre/mlt/converter/MltConverter.java
```diff
@@ -19,25 +19,74 @@
...
+  /// @param columnMappingConfig Optional column mapping configuration to be applied to all layers
```
louwers, line 46: It's not marked `@Nullable` here. Is it really optional?

java/mlt-core/src/test/java/org/maplibre/mlt/MltGenerator.java
TimSylvester: Ugh, I hate this thing, it's not even used.  @nyurik Does the new synthetic testing support replace `MltGenerator`?  I've only kept it around in case it might be useful.

TimSylvester:
Logging is going to be a ton of changes all over, I'd rather a separate PR.  This one has already grown too much.

#941

java/mlt-core/src/test/java/org/maplibre/mlt/MltGenerator.java
nyurik: @TimSylvester i'm not sure what you mean here

TimSylvester: We have this class, `MltGenerator` that @mactrem added to create scenarios with various combinations of options.  All the tests in it are currently disabled.  I suspect that the synthetic test case generation you've been working on fulfills the goal better than that code would, even if it were being used, allowing it to be removed entirely rather than fixed and re-enabled.

PR #819, opened 2026-02-07
Use geojson for geometries instead of WKT in Java CLI

HarelM:
This changes the output of the CLI json to be "regular" geojson for each layer instead of using WKT for the geometries, which creates a "weird" looking json.

Related to:
- https://github.com/maplibre/maplibre-tile-spec/pull/780

### DIFF
--- a/java/mlt-cli/build.gradle
+++ b/java/mlt-cli/build.gradle
@@ -27,6 +27,7 @@ dependencies {
     implementation 'jakarta.annotation:jakarta.annotation-api:3.0.0'
     implementation 'org.jetbrains:annotations:26.0.2-1'
     implementation 'org.locationtech.jts:jts-core:1.20.0'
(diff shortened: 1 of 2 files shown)


HarelM:
If there's a need for better performance the initialization of the writer can probably be done once, I'm not sure how important that is in this case...

HarelM:
@louwers, @TimSylvester let me know what's the best way to test this, as this is currently not tested, unfortunately.

nyurik:
if this is all it takes to start using geojson i'm all for it.  Are there any checked-in code that relies on the string representation of MLT to be WKT?

nyurik (approved):

HarelM:
This PR can be closed as the other PR is incorporating it, I don't have some feelings in which PR it should be merged, your call.

PR #821, opened 2026-02-10
Enable coverage for C++

louwers:
Run with

```
just cpp-coverage
```

Coverage is created as TXT, XML and HTML in build dir.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ codecov*
 MODULE.bazel.lock
 *.diff.geojson
 *.new.geojson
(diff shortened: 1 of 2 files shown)


louwers:
@CommanderStorm I don't think build scripts/tools should install system dependencies.

I think we should remove that because some people might want to use their package manager to install `fd` instead of `cargo-install`. And in any case we should not install stuff without people's explicit permission. In corporate environments where all software needs to be pre-approved that could even get people into trouble.

PR #824, opened 2026-02-11
chore: update Gradle setup action to version 5

nyurik:


louwers:
Why is dependabot not doing this?

PR #826, opened 2026-02-11
chore(ci): ensure all CI keeps git clean

nyurik:
git should always stay clean during CI as it ensures the commands do not produce junk artifacts in user builds either.  If anything changes, those files should be added to gitignore.

### DIFF
--- a/.github/workflows/ts-bump-version.yml
+++ b/.github/workflows/ts-bump-version.yml
@@ -20,6 +20,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+      - uses: taiki-e/install-action@v2
+        with: { tool: 'just,cargo-binstall' }
 
       - name: Update version, commit, and create tag
         working-directory: ts
--- a/.github/workflows/ts-release.yml
+++ b/.github/workflows/ts-release.yml
@@ -19,6 +19,8 @@ jobs:
         shell: bash
     steps:
       - uses: actions/checkout@v6
+      - uses: taiki-e/install-action@v2
+        with: { tool: 'just,cargo-binstall' }
 
       - name: Check if version has been updated
         id: check
@@ -41,6 +43,8 @@ jobs:
         working-directory: ts
     steps:
       - uses: actions/checkout@v6
+      - uses: taiki-e/install-action@v2
+        with: { tool: 'just,cargo-binstall' }
 
       - uses: ./.github/actions/mlt-setup-node
 
@@ -58,8 +62,7 @@ jobs:
         run: npm ci
 
       - name: Build
-        run: |
-          npm run build
+        run: npm run build
 
       - name: Merge license files into LICENSE.txt
         run: |


.github/workflows/ts-bump-version.yml
```diff
@@ -33,3 +33,5 @@ jobs:
           commit-message: bump @maplibre/mlt version to ${{ env.version }}
           branch: mlt-js-version-${{ env.version }}
           title: bump @maplibre/mlt version to ${{ env.version }}
+
+      - run: just -v assert-git-is-clean
```
HarelM, line 39: This is a manual step to update the package version, this check does not apply here.

.github/workflows/ts-release.yml
```diff
@@ -26,6 +26,8 @@ jobs:
         with:
           file-name: ts/package.json
 
+      - run: just -v assert-git-is-clean
```
HarelM, line 31: I don't think this check is relevant here in this file either as I believe npm pack/publish creates a local tgz file.
This is also not part of the PR CI and only runs on main when the version changes...

.github/workflows/gh-pages.yml
```diff
@@ -23,3 +23,5 @@ jobs:
         if: github.event_name != 'pull_request' && github.ref == 'refs/heads/main' && github.repository_owner == 'maplibre'
         uses: JamesIves/github-pages-deploy-action@4a3abc783e1a24aeb44c16e869ad83caf6b4cc23 # v4.7.4
         with: { branch: gh-pages, folder: mkdocs/site }
+
+      - run: just -v assert-git-is-clean
```
HarelM, line 27: Not sure about this file either...

.github/workflows/java-release.yml
```diff
@@ -46,6 +46,7 @@ jobs:
       - name: Build CLI tool
         working-directory: java
         run: ./gradlew cli
+      - run: just -v assert-git-is-clean
```
HarelM, line 49: This is also a release file, I'm not familiar with java's release process, but I think it might not be relevant to this file as well.

.github/workflows/rust.yml
```diff
@@ -66,6 +67,8 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: target/llvm-cov/codecov.info
+      - run: just -v assert-git-is-clean
+      - run: cd ../.. && just -v assert-git-is-clean
```
HarelM, line 69: Isn't running this on the root folder enough after running all the CI steps? Why is there a need to run this twice?

louwers (changes requested):
Please use a reusable action instead, e.g.

```
  - uses: infotroph/tree-is-clean@v1
```

.github/workflows/ts-release.yml
nyurik: @HarelM it is totally fine to create a local tgz file - we should  just add it to .gitignore

.github/workflows/rust.yml
nyurik: good catch, accident

nyurik:
@louwers why would we want to add a dependency to CI, which is not locally testable?  With `just`, we should (eventually) be able to remove all the complex setup steps, and instead simplify all CI to just oneliners, e.g.

```
# These will go into per-language jobs
run: just ci-test-java
run: just ci-test-c
run: just ci-test-ts
```

And the git status test will go inside of these.  The main important point is that no custom logic will be in the github steps, but rather in justfile - testable locally.

justfile
```diff
@@ -183,6 +189,17 @@ generate-one-expected-mlt file:
 extract-version language tag:
     @echo "{{replace(replace(tag, language + '-', ''), 'v', '')}}"
 
+# Make sure the git repo has no uncommitted changes
+[private]
+assert-git-is-clean:
+    @if [ -n "$(git status --untracked-files --porcelain)" ]; then \
+        >&2 echo "ERROR: git repo is no longer clean. Make sure compilation and tests artifacts are in the .gitignore, and no repo files are modified." ;\
```
louwers: Consider using ::error:: instead.

```
echo "::error::Your error message"
```

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-creating-an-annotation-for-an-error

louwers:
@nyurik So we don't need to maintain it? Well I guess it's simple enough to check in...

But why `just` though, which has its own bespoke programming language, instead of a Bash script or reusable action?

nyurik:
@louwers because any significantly large project uses make-like tooling, and just is by far the most popular modern alternative to it with 30k stars. Also, because just is a container language - i.e. it packages many tiny bash scripts, vs having a directory of script files that do not work well together.  Lastly, because we already use it, so it will be easier to just add a new test into existing infrastructure

nyurik:
oh, and because reusable actions are not (easily) runnable locally

.github/workflows/ts-bump-version.yml
nyurik: same

PR #832, opened 2026-02-12
install gcovr with brew

louwers:
Fixes the workflow.

Also renames it cpp-build so we can make it a required workflow.

Also updates the justfile to print a message when gcovr is not installed.

### DIFF
--- a/justfile
+++ b/justfile
@@ -257,7 +257,7 @@ cpp-test: cpp-cmake-build
     ctest
 
 [working-directory: 'cpp/build']
-cpp-coverage: _check-gcovr cpp-test
+cpp-coverage: check-gcovr-is-installed cpp-test
     gcovr --root ../.. \
         --filter ../src --filter ../include \
         --txt coverage.txt \
@@ -265,5 +265,5 @@ cpp-coverage: _check-gcovr cpp-test
         --html-details coverage.html
     @echo "Coverage report at $PWD/coverage.html"
 
-_check-gcovr:
+check-gcovr-is-installed:
     @which gcovr > /dev/null || (echo "Error: gcovr is not installed. Please install it using: brew install gcovr (macOS) or pip3 install gcovr (Linux)" && exit 1)


.github/workflows/cpp.yml
```diff
@@ -34,7 +34,12 @@ jobs:
...
         with: { tool: 'just,cargo-binstall' }
 
-      - name: install gcovr
+      - name: install gcovr (macOS)
+        if: runner.os == 'macOS'
+        run: brew install gcovr
+
+      - name: install gcovr (Linux)
+        if: runner.os == 'Linux'
         run: pip3 install gcovr
```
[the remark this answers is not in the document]
louwers: It works, shut up.

justfile
```diff
@@ -257,10 +257,13 @@ cpp-test: cpp-cmake-build
...
+cpp-coverage: _check-gcovr cpp-test
     gcovr --root ../.. \
         --filter ../src --filter ../include \
         --txt coverage.txt \
         --cobertura-pretty --cobertura coverage.xml \
         --html-details coverage.html
-    echo "Coverage report at $PWD/coverage.html"
+    @echo "Coverage report at $PWD/coverage.html"
+
+_check-gcovr:
```
CommanderStorm:
```suggestion
# checks if gcovr is installed
[private]
check-gcovr-is-installed:
```

justfile
```diff
@@ -257,10 +257,13 @@ cpp-test: cpp-cmake-build
     ctest
 
 [working-directory: 'cpp/build']
-cpp-coverage: cpp-test
+cpp-coverage: _check-gcovr cpp-test
```
CommanderStorm:
```suggestion
cpp-coverage: check-gcovr-is-installed cpp-test
```

justfile
louwers: isn't that what the underscore does?

louwers: image
It's already private

CommanderStorm: none of the things in https://github.com/louwers/maplibre-tile-spec/blob/main/justfile do that this way.
Lets stay consistent, ok?

louwers: Doch doch https://github.com/maplibre/maplibre-tile-spec/blob/1ba44f87c55da3f4ede2e9d42446d05eb24bc59b/justfile#L139-L161

nyurik: tbh, i prefer the underscore - doesn't waste another line, and it is not important enough to highlight

PR #687, opened 2025-11-07
feat(rust): pass all synthetic tests

nyurik:
This PR ensures java-generated synthetic tests all pass, and the parsing result generates identical "geojson-like" data as stored in git.

This pull request introduces a new `geojson` module for representing decoded MLT data in a GeoJSON-like format, adds serialization dependencies, and makes several improvements to decoding utilities and error handling. It also updates the build and lint configuration for consistency and clarity.

GeoJSON support and serialization:

* Added a new `geojson` module (`rust/mlt-nom/src/geojson.rs`) that provides types and conversion functions to represent decoded MLT data as GeoJSON-like structures with i32 coordinates. This includes `FeatureCollection`, `Feature`, `Geometry`, and CRS support.
* Updated `Cargo.toml` to add `serde`, `serde_json`, and `strum` dependencies, enabling serialization and deserialization for the new GeoJSON types.

Decoding utilities and error handling:

* Improved decoding utilities in `utils.rs` by using `WrappingAdd` for safer arithmetic, adding a new `decode_byte_rle` function for ORC-style RLE decoding, and refactoring debug output for optional sequences. [[1]](diffhunk://#diff-5bea486f07de5d8ce7a11b1bd9aa3b59ac82de26003434a156e0a1ad3cf3904bL4-R4) [[2]](diffhunk://#diff-5bea486f07de5d8ce7a11b1bd9aa3b59ac82de26003434a156e0a1ad3cf3904bL89-R91) [[3]](diffhunk://#diff-5bea486f07de5d8ce7a11b1bd9aa3b59ac82de26003434a156e0a1ad3cf3904bL99-R102) [[4]](diffhunk://#diff-5bea486f07de5d8ce7a11b1bd9aa3b59ac82de26003434a156e0a1ad3cf3904bL109-R116) [[5]](diffhunk://#diff-5bea486f07de5d8ce7a11b1bd9aa3b59ac82de26003434a156e0a1ad3cf3904bR178-R237)
* Added a new `DecodeError` variant to `MltError` for more descriptive decode errors, and updated error handling in decoding-related code. [[1]](diffhunk://#diff-862369872e52deb50f9179ac78a1b35d73dc1f2b19c2aee10e90c4f3824581ceR12-R13) [[2]](diffhunk://#diff-4d2842453dbb31ee497bedeb7e2aa167761634ac26d20c4a8df68b1854aae9b7L17-R26)

Layer and geometry improvements:

[...]

PR #835, opened 2026-02-12
feat(rust): Id32 decoding support

CommanderStorm:
The first blocker in getting the e2e tests to run seems to be the missing id decoding logic.

This PR starts with the process of implementing this.

Doing stuff at a lower level is fun!


CommanderStorm:
Openend a PR against this instead.
The diff is smaller

- https://github.com/maplibre/maplibre-tile-spec/pull/687

PR #841, opened 2026-02-13
chore(tests): tag benchmark tests and exclude from default test run

nyurik:
java CI tests take very long time compared to all others due to running benchmarks in CI - mostly useless and wasteful.

### DIFF
--- a/java/mlt-core/build.gradle
+++ b/java/mlt-core/build.gradle
@@ -39,6 +39,8 @@ dependencies {
 
 test {
     jvmArgs "-Dcom.google.protobuf.use_unsafe_pre22_gencode"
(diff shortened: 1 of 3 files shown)


louwers:
Can you instead reduce the number of iterations on CI so it at least still runs?

nyurik:
@louwers I tried to re-enable them one by one:
* without both: 4 seconds total
* `MltDecoderBenchmarkTest`:  5+ minutes -- seems like it uses a 200 instance
* `CompressionBenchmarksTest`: 1.5 minutes -- there is no loop that I can see

I modified `MltDecoderBenchmarkTest` to only run once per test - this reduces it to 30 seconds (still much worse than 4s, but ok), and kept the `CompressionBenchmarksTest` as bench-only.

PR #844, opened 2026-02-14
test: Add `DELTA_RLE`  and `RLE` encoding for ids

CommanderStorm:


java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -37,14 +37,24 @@ private static void generateIds() throws IOException {
     write("id", feat(p1, 100L), cfg().ids());
     write("id0", feat(p1, 0L), cfg().ids());
```
CommanderStorm: more IDs are nessesary here as otherwise `DELTA_RLE` and `DELTA` don't differ in output.

Both inputs are minimal testcases

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -37,14 +37,32 @@ private static void generateIds() throws IOException {
...
 
-    var pts = new Feature[] {feat(p1, 100L), feat(p2, 101L), feat(p3, 103L)};
+    var pts = new Feature[] {feat(p1, 100L), feat(p2, 101L), feat(p3, 102L), feat(p4, 103L)};
     write(layer("ids", pts), cfg().ids());
     write(layer("ids-delta", pts), cfg(DELTA).ids());
+    // RLE for IDs does not actually make sense, but it is in the spec
+    var ptsRLE = new Feature[] {feat(p1, 100L), feat(p2, 100L), feat(p3, 100L)};
+    write(layer("ids-rle", ptsRLE), cfg(RLE).ids());
```
CommanderStorm: at least I can't think of a case where this makes sense

PR #846, opened 2026-02-14
chore(java): make tests use same data

nyurik:
@CommanderStorm this reworks #843 a bit and disables one of the tests

### DIFF
--- a/java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
+++ b/java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
@@ -119,16 +119,16 @@ public class SyntheticMltGenerator {
                 0.123456789)),
         cfg());
 
-    var feat_delta_ints =
+    var feat_ints =
         array(
             feat(p1, props("int", 42)),
             feat(p2, props("int", 42)),
             feat(p3, props("int", 42)),
             feat(p4, props("int", 42)));
-    write(layer("props-int", feat_delta_ints), cfg());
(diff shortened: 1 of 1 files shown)


java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -123,26 +120,15 @@ private static void generateProperties() throws IOException {
         cfg());
 
     var feat_delta_ints =
```
```suggestion
    var feat_ints =
```

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -123,26 +120,15 @@ private static void generateProperties() throws IOException {
...
-    write(layer("props-int-rle", feat_rle_ints), cfg(RLE));
-
-    var feat_delta_rle_ints =
-        Stream.concat(Arrays.stream(feat_delta_ints), Arrays.stream(feat_rle_ints))
-            .toArray(Feature[]::new);
-    write(layer("props-int-delta-rle", feat_delta_rle_ints), cfg(DELTA_RLE));
+    write(layer("props-int", feat_delta_ints), cfg());
+    write(layer("props-int-delta", feat_delta_ints), cfg(DELTA));
+    write(layer("props-int-rle", feat_delta_ints), cfg(RLE));
+    write(layer("props-int-delta-rle", feat_delta_ints), cfg(DELTA_RLE));
```
```suggestion
    write(layer("props-int", feat_ints), cfg());
    write(layer("props-int-delta", feat_ints), cfg(DELTA));
    write(layer("props-int-rle", feat_ints), cfg(RLE));
    write(layer("props-int-delta-rle", feat_ints), cfg(DELTA_RLE));
```

PR #849, opened 2026-02-14
refactor: further synthetics cleanup

nyurik:
* Use non-zero point for most tests
* add `prop` and `kv` helpers to help with property creation


nyurik:
fixed

PR #852, opened 2026-02-14
feat(java): support `UInt32` and `UInt64` encoding/decoding for synthetics

CommanderStorm:
I thought it best to put this all into one PR to get feedback on the basic design of this feature.

### DIFF
--- a/ts/src/synthetic.spec.ts
+++ b/ts/src/synthetic.spec.ts
@@ -24,7 +24,9 @@ const UNIMPLEMENTED_SYNTHETICS = new Map([
     ["prop_f64_neg_zero", "throws error 'byte length of Float64Array should be a multiple of 8'"],
     ["prop_str_empty", "empty strings not supported"],
     ["prop_i64_min", "wraps to 0 despite it should have been negative"],
+    ["prop_u32_max", "wraps to -1 despite it should have been positive"],
     ["prop_i64_max", "wraps to -1 despite it should have been positive"],
+    ["prop_u64_max", "wraps to -1 despite it should have been positive"],
 ]);
 
 const syntheticDir = resolve(__dirname, "../../test/synthetic/0x01");


java/mlt-core/src/main/java/org/maplibre/mlt/data/Unsigned.java
CommanderStorm, line 1: This class needs a bit of a java person to look at.
There are many ways to do this, I chose the `Unsigned<Long>`, `Unsigned<Integer>`, `Unsigned<Byte>` way.

@nyurik you said something about performance on slack. How did you benchmark that?

java/mlt-core/src/main/java/org/maplibre/mlt/data/Unsigned.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /** Creates an unsigned 64-bit integer (range: 0 to 2^64-1). */
+  public static Unsigned<Long> u64(BigInteger value) {
+    final BigInteger U64_MAX_VALUE = new BigInteger("18446744073709551615");
+
+    if (value.compareTo(BigInteger.ZERO) < 0 || value.compareTo(U64_MAX_VALUE) > 0) {
+      throw new IllegalArgumentException(
+          "Value " + value + " is out of range for UInt64 [0, 2^64-1]");
+    }
```
CommanderStorm: I am using `BigInteger` (and in general the one size up in the constructor) as otherwise I am forcing people to do encode the values in twos complement themselve.. (i.e. what I am doing in the serializer for JSON compatability)

Better for usability, likely worse for performance though..

java/mlt-core/src/main/java/org/maplibre/mlt/data/Unsigned.java
nyurik: `time just test` I think

test/synthetic/0x01/prop-uint64-max.json
```diff
@@ -0,0 +1,26 @@
...
+        "crs": {
+          "type": "name",
+          "properties": {
+            "name": "EPSG:0"
+          }
+        }
+      },
+      "id": 0,
+      "properties": {
+        "bignum": -1,
```
CommanderStorm: still serialized wrong

nyurik (approved):
i'm not a java/encoder expert, but looks good to my non-expert eye

java/mlt-core/src/main/java/org/maplibre/mlt/data/unsigned/Unsigned.java
```diff
@@ -0,0 +1,55 @@
+package org.maplibre.mlt.data.unsigned;
+
+/**
+ * Represents an unsigned integer of a fixed bit width (8, 32, or 64 bits).
+ *
+ * <p>This interface abstracts over different unsigned integer types in Java, which do not natively
+ * support unsigned primitives, providing a common API for working with unsigned values regardless
+ * of their underlying representation.
+ */
+public sealed interface Unsigned permits U8, U32, U64 {
```
HarelM, line 10: I think adding an interface to abstract code makes sense, but this PR has a lot of "instance of", which breaks this idea. If we can't get rid of the instance of calls I would consider removing this interface...

CommanderStorm: shure, I can also move them into the enum items themselves

java/mlt-core/src/main/java/org/maplibre/mlt/decoder/PropertyDecoder.java
```diff
@@ -72,13 +73,37 @@ private static Object decodeScalarPropertyColumn(
       case INT_32:
```
msbarry: Minor: switching over enums can be  a little cleaner now with new switch expression syntax:

```java
return switch (scalarType.physicalType) {
  case UINT_32, INT_32 -> {
    ..
    yield unpack(values, presentStream, presentStreamSize);
  }
```

That gives the benefit of a compiler error when we add a new enum value that's not handled.

java/mlt-core/src/main/java/org/maplibre/mlt/converter/MltConverter.java
```diff
@@ -656,10 +659,15 @@ private static MltMetadata.ScalarType getScalarType(Map.Entry<String, Object> pr
...
+    } else if (propertyValue instanceof U8) {
+      return MltMetadata.ScalarType.UINT_8;
+    } else if (propertyValue instanceof U32) {
+      return MltMetadata.ScalarType.UINT_32;
+    } else if (propertyValue instanceof U64) {
+      return MltMetadata.ScalarType.UINT_64;
+    } else if (propertyValue instanceof Byte) {
+      return MltMetadata.ScalarType.INT_8;
+    } else if (propertyValue instanceof Integer) {
       return MltMetadata.ScalarType.INT_32;
```
msbarry: Minor: can simplify this a bit with:

```java
return switch (propertyValue) {
  case Boolean v -> MltMetadata.ScalarType.BOOLEAN;
  case U8 v -> MltMetadata.ScalarType.UINT_8;
  ...
```

java/mlt-core/src/main/java/org/maplibre/mlt/data/unsigned/U32.java
```diff
@@ -0,0 +1,35 @@
...
+
+  @Override
+  public Integer intValue() {
+    return value;
+  }
+
+  @Override
+  public Long longValue() {
+    return (long) value;
+  }
```
msbarry, line 29: Should this return `Integer.toUnsignedLong` ?

java/mlt-core/src/main/java/org/maplibre/mlt/data/unsigned/U8.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Override
+  public Integer intValue() {
+    return (int) value;
+  }
+
+  @Override
+  public Long longValue() {
+    return (long) value;
+  }
```
msbarry, line 25: should these use Byte.toUnsignedInt ?

java/mlt-core/src/main/java/org/maplibre/mlt/decoder/PropertyDecoder.java
CommanderStorm: The syntax seems to trip up my editor a bit due to some "misplaced constructs", but the other tests seem to work. Pushed it

java/mlt-core/src/main/java/org/maplibre/mlt/data/unsigned/U32.java
CommanderStorm: Sure, I thought that just a long cast is more redable, but we can also change it to this.

java/mlt-core/src/main/java/org/maplibre/mlt/data/unsigned/U8.java
CommanderStorm: sure

CommanderStorm:
I think it is simpler to maintain support for this in main. If someone has comments on this, I would be happy to do a followup

PR #854, opened 2026-02-15
test: Add shared dictionary synthetics

CommanderStorm:
This PR adds a synthetic test for shared dictionarys


java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltUtil.java
```diff
@@ -200,7 +218,21 @@ static void write(String fileName, List<Layer> layers, ConversionConfig.Builder
...
+      var columnMappings = new HashMap<Pattern, List<ColumnMapping>>();
+      if (config.getOptimizations() != null && !config.getOptimizations().isEmpty()) {
+        var allColumnMappings =
+            config.getOptimizations().values().stream()
+                .flatMap(opt -> opt.columnMappings().stream())
+                .toList();
+        if (!allColumnMappings.isEmpty()) {
+          columnMappings.put(Pattern.compile(".*"), allColumnMappings);
+        }
+      }
```
CommanderStorm, line 245: this is making sure that we encode as a struct (?). Is this correct?

rust/mlt-nom/src/v01/root.rs
```diff
@@ -195,16 +208,13 @@ impl OwnedLayer01 {
             // self.id.write_to(writer)?;
             return Err(io::Error::new(
                 io::ErrorKind::Unsupported,
-                format!("{}", MltError::NotImplemented("ID write not implemented")),
+                format!("{}", MltError::NotImplemented("ID write")),
```
CommanderStorm:
```suggestion
                MltError::NotImplemented("ID write").to_string(),
```

rust/mlt-nom/src/v01/root.rs
```diff
@@ -163,11 +161,24 @@ fn parse_columns_meta(
...
+                let mut children = Vec::new();
+                let child_count;
+                (input, child_count) = utils::parse_varint::<usize>(input)?;
+
+                for _ in 0..child_count {
+                    let child;
+                    (input, child) = Column::parse(input)?;
+                    children.push(child);
+                }
+                typ.children = children;
```
CommanderStorm, line 180: I know structs are experiemntal, but is this hack right here actually how it is supposed to work?

Are structs requred for shared dict encoding?

CommanderStorm: https://github.com/maplibre/maplibre-tile-spec/blob/1e6eb05c4ecd823ae627be51c727a4da1beb6eab/cpp/src/mlt/decode/string.hpp#L104

It seems like it is required

Issue #836, opened 2026-02-12
Improve Synthetic test coverage


nyurik:
Our synthetic test cases generated by [SyntheticMltGenerator.java](https://github.com/maplibre/maplibre-tile-spec/blob/main/java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java) are not covering some cases.  Feel free to implement any one of these, run `just generate-synthetic-mlts` and create a PR.

* [ ] `u8` property columns
* [ ] `i8` property columns
* [x] `u32` property columns https://github.com/maplibre/maplibre-tile-spec/pull/852
* [x] `u64` property columns https://github.com/maplibre/maplibre-tile-spec/pull/852
* [ ] `Double.MAX_VALUE` cannot be encoded
* [x] `morton`-encoded data
* [x] `pre-tesselated` data - https://github.com/maplibre/maplibre-tile-spec/pull/837
* [x] Optional IDs (both i32 and i64)
  * [x] vibed PR needs feedback: https://github.com/maplibre/maplibre-tile-spec/pull/829
* [ ] Optional for every possible property type
* [ ] Go through every possible encoding and ensure each test case exists
  * [ ] integers
  * [x] floats #858 
  * [x] strings #858 
* [x] Shared dictionaries https://github.com/maplibre/maplibre-tile-spec/pull/854
* [x] Coordinates inside and outside the default tile extent of 0..4095, e.g. -42 or 4112 - https://github.com/maplibre/maplibre-tile-spec/pull/853
* [x] Non-default extent size - e.g. 512 or 2^20. We should be able to support coordinate representation of `i32` -- everything up to 2^30 (buffer is important, so extent=2^31 would allow negative buffer, but not the coordinates with beyond 2^31) - https://github.com/maplibre/maplibre-tile-spec/pull/853

CommanderStorm:
we have a lot of rust-only synthetics, they are considerd canonical in v2. Closing

PR #858, opened 2026-02-15
test: To infinity and beyond! (adding `prop_{f32,f64}_{min_inf,max_inf,nan}` synthetics)

CommanderStorm:
image

https://www.youtube.com/watch?v=XDC74LEZkE8

### DIFF
--- a/rust/mlt-nom/tests/synthetics.rs
+++ b/rust/mlt-nom/tests/synthetics.rs
@@ -32,7 +32,7 @@ fn test_one(mlt: &Path, json: &Path) {
     let actual_json = normalize_tiny_floats(serde_json::to_value(&actual).unwrap());
     let expected_json = normalize_tiny_floats(serde_json::to_value(&expected).unwrap());
 
-    // here we catch the big issues in an user facing way with a nice diff
+    // here we catch the big issues in a user-facing way with a nice diff
     pretty_assertions::assert_eq!(
         serde_json::to_string_pretty(&actual_json).unwrap(),
         serde_json::to_string_pretty(&expected_json).unwrap(),


java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -139,11 +123,17 @@ private static void generateProperties() throws IOException {
...
     write("prop_f32_max", feat(p0, prop("val", Float.MAX_VALUE)), cfg());
+    write("prop_f32_pos_inf", feat(p0, prop("val", Float.POSITIVE_INFINITY)), cfg());
+    write("prop_f32_nan", feat(p0, prop("val", Float.NaN)), cfg());
     write("prop_f64", feat(p0, prop("val", (double) 3.141592653589793)), cfg());
+    write("prop_f64_neg_inf", feat(p0, prop("val", Double.NEGATIVE_INFINITY)), cfg());
     write("prop_f64_min", feat(p0, prop("val", Double.MIN_VALUE)), cfg());
-    // FIXME: fails in Java
-    // write("prop_f64_max", feat(p0, prop("val", Double.MAX_VALUE)), cfg());
+    write("prop_f64_max", feat(p0, prop("val", Double.MAX_VALUE)), cfg());
+    // FIXME: Produces the same output as prop_f64_max
```
```suggestion
    // FIXME: Fails in Java as it Produces the same output as prop_f64_max
```

nyurik:
do we actually want to allow infinity / nans here?  I am not sure we support that in MVT, and for tag 0x01, we shouldn't add any new functionality if it is not already present in MVT

CommanderStorm:
> do we actually want to allow infinity / nans here? I am not sure we support that in MVT, and for tag 0x01, we shouldn't add any new functionality if it is not already present in MVT

I don't know.
I did this because writing the tests for edge cases is on the bucket list for #836
These are defintive edge cases..

nyurik:
heh, well, i guess it does belong because MVT uses https://github.com/mapbox/vector-tile-spec/blob/5330dfc6ba2d5f8c8278c2c4f56fff2c7dee1dbd/2.1/vector_tile.proto#L20 for float properties:

```protobuf
message Value {
        optional string string_value = 1;
        optional float float_value = 2;
        optional double double_value = 3;
```

and protobuf does define them to be valid. I have no idea how style spec might handle them (cc: @HarelM ?), but if they exist in MVT, they would need to be supported

rust/mlt-nom/tests/synthetics.rs
```diff
@@ -18,10 +19,55 @@ fn test_one(mlt: &Path, json: &Path) {
...
+    );
+}
+
+/// Replace extremely small float values (< 1e-40) with 0.0 to handle codec precision issues
+fn normalize_tiny_floats(value: &Value) -> Value {
+    match value {
+        Value::Number(n) => {
+            if let Some(f) = n.as_f64() {
+                let eps = f64::from(f32::EPSILON);
+                if f.abs() < eps && f != 0.0 {
```
nyurik: not sure - do you need to check if this is a real number, i.e. not a NaN or Inf?   Also, why do you need !=0.0 check?

rust/mlt-nom/src/geojson.rs
```diff
@@ -48,7 +48,7 @@ impl FeatureCollection {
             for i in 0..geom.vector_types.len() {
                 let id = ids.and_then(|v| v.get(i).copied().flatten()).unwrap_or(0);
                 let geometry = geom.to_geojson(i)?;
-                let mut properties = HashMap::new();
+                let mut properties = BTreeMap::new();
```
nyurik, line 51: why do you need btreemap here? we shouldn't be preserving order

CommanderStorm: because otherwise we serialize to JSON in random order.
Not what we want..

We want our keys to be sorted actually

rust/mlt-nom/tests/synthetics.rs
CommanderStorm: yes, good catch

PR #860, opened 2026-02-15
fix(rust): Make tests easier to debug and fix CI

CommanderStorm:
Basically, this will render like this:

image

### DIFF
--- a/rust/mlt-nom/src/v01/geometry.rs
+++ b/rust/mlt-nom/src/v01/geometry.rs
@@ -267,7 +267,7 @@ impl<'a> FromRaw<'a> for DecodedGeometry {
             //
             // return FlatGpuVector::new(vector_types, triangles, index_buffer, vertices);
             return Err(MltError::NotImplemented(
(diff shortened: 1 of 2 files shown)


rust/mlt-nom/tests/synthetics.rs
```diff
@@ -21,7 +22,45 @@ fn test_one(mlt: &Path, json: &Path) {
...
+        serde_json::to_string_pretty(&expected_json).unwrap()
+    );
+}
+
+/// Replace extremely small float values (< 1e-40) with 0.0 to handle codec precision issues
+fn normalize_tiny_floats(value: &Value) -> Value {
+    match value {
+        Value::Number(n) => {
+            if let Some(f) = n.as_f64() {
+                if f.abs() < 1e-40 && f != 0.0 {
```
nyurik: should this be f32::EPSILON?

CommanderStorm: yes

nyurik (approved):
see comment in the other PR that seems to contain this one, otherwise good to go

Issue #847, opened 2026-02-14
Add a JavaScript test to parse all synthetic test files


nyurik:
We now have many synthetic tests in the `test/synthetic/0x01` directory, consisting of a tiny MLT file, and a corresponding GeoJSON file. A test should parse all .mlt files, generate corresponding GeoJSON (in memory), and compare the result with the GeoJSON stored in the dir.  Lastly, add this test to the CI.

### Notes
* we use integer tile coordinates in the range 0..4095. Coordinates are allowed to be somewhat outside of that range, e.g. -42 or 4112 - to represent the surrounding buffer, or the value could be even higher e.g. for the tiles with larger extent
* geojson includes CRS section which can be safely ignored
* Layer name is added as a property to each feature

Similar C++ ticket: https://github.com/maplibre/maplibre-tile-spec/issues/848

HarelM:
I'm working to remove all skipped synthetic tests.
The last issues I now see are with conversion between signed and unsigned numbers.
I've fixed it for long, but for integers it's a bit more complicated given all the places that use integeres.
@DoFabien can you check if the fastPFor can use `Uint32Array` everywhere instead of `Int32Array`? As far as I understand fastPFor works on unsigned integers only, right?
@Turtelll can you help me understand why `IntConstVector` is using `Int32Array` conversion - this basically trims max unsigned integers so I'm not entirely sure how to address this code path.
Help would be appreciated.
I created the following branch to work on this, but I'm not sure how to proceed:
https://github.com/HarelM/maplibre-tile-spec/tree/min-i32

Any help would be appreciated.

DoFabien:
Hi, 
I took a look at `min-i32`.

On the `fastPFor` question: yes, for the actual `fastPFor` value path, `Uint32Array` is the right type. The TS `fastPFor` encoder/decoder already operate on `Uint32Array` there. The remaining issue is elsewhere: unsigned TS values are still flowing through signed vector types after decode.

One thing also looked suspicious in `mltDecoder`: for unsigned `CONST` IDs, the new `isSigned` flag seems to be passed as `true`, which would still pick signed storage there.

So my read is: good direction overall, and the fix needed here is the TS unsigned vector path, not `fastPFor` itself.

I’m going to open a small PR on top of your branch for that narrower TS fix:

- preserve unsigned `UINT_32` / `UINT_64` values in the flat-vector path
- fix the unsigned `CONST` ID path
- add the related TS regression tests

[...]

PR #863, opened 2026-02-16
test(ts): Add synthetic tests for TS

CommanderStorm:
This PR adds support for our synthetic conformance tests for the TS decoder

Resolves https://github.com/maplibre/maplibre-tile-spec/issues/847


HarelM:
Thanks! I was planning on doing that in the next few days, you've beat me to it.
The skipped test are "bugs" that needs to be, right (besides the missing fastpfor)?

Also, why do we need json5 to parse json? Is the synthetic generated json not a valid json that can be parsed using JSON.parse?

nyurik:
json does not support +/- infinity and nans for floats. We need to support everything mvt supports, which seem to mean we must support them.  If neither web nor native support infinity or nans, we can officially declare that we do not support them, and ensure parsers and encoders fail if present

HarelM:
In what user case would that be used? I'm not sure the style spec or the renderers should/do support that. 
Is that part of MVT so it's about feature party? 
I would consider not supporting it at the moment and see what use case requires this. But that might just be me.

nyurik:
my only concern is to either have feature parity with mvt, or to have clear specification that we do not allow certain values.  We already have a mismatch for non-consistent value types in a single column, so it is fine to clearly state it and ensure all encoder and decoders enforce real float values.

CommanderStorm:
`props_mixed` (I think this is what you are talking about) is a real bug where somewhere in the ts decoder we have a leght reading issue.
Threre is a lot of code to read and it does not occur in native.

CommanderStorm:
-> https://github.com/maplibre/maplibre-tile-spec/pull/922

Issue #848, opened 2026-02-14
Add a C++ test to parse all synthetic test files


nyurik:
We now have many synthetic tests in the `test/synthetic/0x01` directory, consisting of a tiny MLT file, and a corresponding GeoJSON file. A test should parse all .mlt files, generate corresponding GeoJSON (in memory), and compare the result with the GeoJSON stored in the dir.  Lastly, add this test to the CI.

### Notes
* we use integer tile coordinates in the range 0..4095. Coordinates are allowed to be somewhat outside of that range, e.g. -42 or 4112 - to represent the surrounding buffer, or the value could be even higher e.g. for the tiles with larger extent
* geojson includes CRS section which can be safely ignored
* Layer name is added as a property to each feature

Similar JavaScript ticket:  #847

PR #866, opened 2026-02-16
test(cpp): Add synthetic tests for MLT decoder

CommanderStorm:
I am having a bit of trouble getting this to run reliably.

Most tests don't seem to work due to
- either build system issues (I tampered with some cmake files, @louwers I am not an expert at this) or
- the `Decoder::Impl::makeFeatures` -> 0 features bug I noted below

Resolves https://github.com/maplibre/maplibre-tile-spec/issues/848

Disclaimer: I got some help on the json parts, doing json is not as simple in cpp as I thought.

### DIFF
--- a/cpp/include/mlt/util/file_io.hpp
+++ b/cpp/include/mlt/util/file_io.hpp
@@ -2,7 +2,7 @@
 
 namespace mlt::util {
 /// Load binary file contents
-inline std::vector<char> loadFile(const std::filesystem::path& path) {
+std::vector<char> loadFile(const std::filesystem::path& path) {
     std::ifstream file(path, std::ios::binary | std::ios::ate);
     if (!file.is_open()) {
         throw std::runtime_error("Failed to open file: " + path.string());
@@ -23,7 +23,7 @@ inline std::vector<char> loadFile(const std::filesystem::path& path) {
 }
 
 /// Load text file contents
-inline std::string loadTextFile(const std::filesystem::path& path) {
+std::string loadTextFile(const std::filesystem::path& path) {
     std::ifstream file(path);
     if (!file) {
         throw std::runtime_error("Failed to open file: " + path.string());


cpp/test/test_synthetics.cpp
```diff
@@ -0,0 +1,591 @@
...
+std::vector<std::pair<std::string, fs::path>> findSyntheticTests(const fs::path& dir);
+
+// Instantiate tests for all synthetic test files found
+INSTANTIATE_TEST_SUITE_P(Synthetic0x01,
+                         SyntheticTest,
+                         ::testing::ValuesIn(findSyntheticTests("../../../test/synthetic/0x01")),
+                         [](const testing::TestParamInfo<SyntheticTest::ParamType>& info) {
+                             // Use test name as the test case name
+                             return info.param.first;
+                         });
```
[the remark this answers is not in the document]
CommanderStorm: `just test-cpp` and running in CLion works..
CI might need to be fixed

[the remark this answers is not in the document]
CommanderStorm: actually, clion had it wrong

cpp/test/test_synthetics.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+    // Replace JavaScript-style special values with valid JSON
+    // We need to preserve them as special markers that we can handle later
+
+    // Handle -Infinity first (before Infinity)
+    size_t pos = 0;
+    while ((pos = result.find("-Infinity", pos)) != std::string::npos) {
+        const bool isWord = (pos > 0 && std::isalnum(result[pos - 1])) ||
+                            (pos + 9 < result.length() && std::isalnum(result[pos + 9]));
+        if (!isWord) {
```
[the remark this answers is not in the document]
CommanderStorm: for fucks sake, really?

CommanderStorm:
> Not sure why so much code is needed to 'preprocess' the JSON

Because the -+infinitys or NaN otherwise trip up the json decoder

cpp/include/mlt/util/geojson.hpp
```diff
@@ -0,0 +1,337 @@
...
+                                     [](std::string_view v) -> json { return std::string(v); },
+                                     [](const std::string& v) -> json { return v; },
+                                     [](auto&& opt) -> json {
+                                         if constexpr (requires { opt.has_value(); }) {
+                                             return opt.has_value() ? propertyToJson(*opt) : nullptr;
+                                         } else {
+                                             return nullptr;
+                                         }
+                                     }},
+                          prop);
```
CommanderStorm: this is me trying to do 
`match property ...`

Would need advice how to write this ideomatic cpp

CommanderStorm: After consulting with gemini, I think the enterprise-ready solution of using a visitor it is

cpp/include/mlt/util/file_io.hpp
```diff
@@ -0,0 +1,36 @@
+#pragma once
```
louwers, line 1: Since this is a test utility (not part of the library API) it should be added in a different location.

cpp/include/mlt/util/geojson.hpp
```diff
@@ -0,0 +1,320 @@
+#pragma once
```
louwers, line 1: These are test utility and should not be part of the library API.

cpp/include/mlt/util/geojson.hpp
```diff
@@ -0,0 +1,320 @@
...
+        std::swap(this->features, other.features);
+        return *this;
+    }
+};
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(GeoJsonFeatureCollection, type, features)
+
+class JsonComparator {
+    std::vector<std::string> path_;
+    /// Check if two floating point values are approximately equal
+    static bool floatsApproxEqual(const double actual, const double expected) {
```
louwers, line 166: C++ is not Java. You can have free standing functions.

Implementations should go in `.cpp` files, definitions in `.hpp` files.

cpp/include/mlt/util/file_io.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace mlt::util {
+/// Load binary file contents
+inline std::vector<char> loadFile(const std::filesystem::path& path) {
```
louwers: Only trivial functions should be `inline`d.

cpp/include/mlt/util/geojson.hpp
```diff
@@ -0,0 +1,320 @@
...
+            }
+        }
+
+        if (!floatsApproxEqual(actualVal, expectedVal)) {
+            std::ostringstream ss;
+            ss << "Numeric mismatch at " << currentPath() << ": expected " << expectedVal << ", got " << actualVal;
+            throw std::runtime_error(ss.str());
+        }
+    }
+    void assertEqualArray(const json& expected, const json& actual) {
```
louwers, line 229: Why are these functions needed?

nlohmann implements an `operator==` for the `json` type. https://json.nlohmann.me/api/basic_json/operator_eq/#nlohmannbasic_jsonoperator

cpp/vendor/fastpfor.cmake
```diff
@@ -23,5 +23,5 @@ endif()
 
 target_link_libraries(mlt-cpp FastPFOR)
 target_include_directories(mlt-cpp PRIVATE SYSTEM "${PROJECT_SOURCE_DIR}/vendor/fastpfor/headers")
-target_compile_definitions(mlt-cpp PUBLIC MLT_ENABLE_FASTPFOR=1)
+target_compile_definitions(mlt-cpp PUBLIC MLT_WITH_FASTPFOR=1)
```
louwers: The name is an oversight but I think it's on purpose that FastPFor is not enabled yet.

cpp/test/test_synthetics.cpp
```diff
@@ -0,0 +1,204 @@
...
+        if (std::filesystem::exists(mltPath) && std::filesystem::exists(jsonPath)) {
+            tests.emplace_back(name, dir);
+        }
+    }
+
+    std::ranges::sort(tests);
+    return tests;
+}
+
+/// Preprocess JSON text to handle NaN and Infinity values
```
louwers, line 143: NaN and Infinity are not valid JSON values.

Should we get rid of those test cases?

CommanderStorm: they are valid json5 and this is a good testcase since it found a crash in JS

cpp/vendor/fastpfor.cmake
CommanderStorm: so like this?

```suggestion
target_compile_definitions(mlt-cpp PUBLIC MLT_WITH_FASTPFOR=0)
```

cpp/include/mlt/util/geojson.hpp
CommanderStorm: I am not quite sure how to hack this into the build system.

Headers were resonable to add without too much effort, but cpp files do seem to need some magic there.
Not sure how to do that exactly

CommanderStorm: all the other headers include headers also don't seem to do that

cpp/include/mlt/util/file_io.hpp
CommanderStorm: clang-tidy suggested it

cpp/include/mlt/util/geojson.hpp
CommanderStorm: if we want good error messages, yes.

If we are fine with dumping a hole lot of json, they are likely nessesary

CommanderStorm:
shure, we can add a js based test to the cpp side

PR #868, opened 2026-02-17
feat(rust): Inital fuzzing implementation and underflow protection

CommanderStorm:
This comes with a few code-changes where I needed to fix a few of the buffer underflows and related panics (such as no stream_count, but an optional stream)

Not sure if the latter case is correctly handled tbh.
It could also be that for that case the optional stream should not be accessed?

### DIFF
--- a/rust/mlt-nom/src/v01/root.rs
+++ b/rust/mlt-nom/src/v01/root.rs
@@ -72,7 +72,7 @@ impl Layer01<'_> {
                     }
                     if stream_count == 0 {
                         return Err(MltError::MinLength {
-                            ctx: "geometry stream",
+                            ctx: "geometry type, but without streams",
                             min: 1,
                             got: 0,
                         });


rust/mlt-nom/src/v01/root.rs
```diff
@@ -106,9 +129,23 @@ impl Layer01<'_> {
...
                     // if optional has a value, one stream has already been consumed
-                    let stream_count = stream_count - usize::from(optional.is_some());
+                    let optional_stream_count = optional.is_some() as usize;
+                    let Some(stream_count) = stream_count.checked_sub(optional_stream_count) else {
+                        return Err(MltError::ExpectedValues {
+                            ctx: "stream count does not include optional stream count",
+                            expected: stream_count,
+                            got: optional_stream_count,
+                        });
+                    };
```
CommanderStorm: this is concretely the case which I don't know how to handle.
- abort with error
- don't parse the optional stream if `stream_count == 0`

rust/mlt-nom/src/v01/root.rs
```diff
@@ -168,11 +205,18 @@ fn parse_columns_meta(
...
 
-                for _ in 0..child_count {
+                // Each stream requires at least 1 byte (ColumnType without name)
+                if input.len() < child_column_count {
+                    return Err(MltError::BufferUnderflow {
+                        needed: child_column_count,
+                        remaining: input.len(),
+                    });
+                }
+                let mut children = Vec::with_capacity(child_column_count);
```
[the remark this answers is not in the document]
CommanderStorm:
```suggestion
                // Each collumn requires at least 1 byte (ColumnType without name)
                if input.len() < child_column_count {
                    return Err(MltError::BufferUnderflow {
                        needed: child_column_count,
                        remaining: input.len(),
                    });
                }
                let mut children = Vec::with_capacity(child_column_count);
```

rust/mlt-nom/fuzz/README.md
```diff
@@ -0,0 +1,137 @@
...
+When the fuzzer finds an issue, you can reproduce it using the test infrastructure:
+
+1. The failing input is saved to `artifacts/layer/crash-<hash>`
+2. Minimize the input using `cargo fuzz tmin layer artifacts/layer/crash-<hash>`
+3. Edit `tests/reproduce.rs` and update the filename:
+   ```rust
+   let bytes = include_bytes!("../artifacts/layer/minimized-from-<hash>");
+   ```
+
+3. Run the test:
```
CommanderStorm:
```suggestion
4. Run the test:
```

rust/mlt-nom/src/v01/root.rs
nyurik: my overall understanding:

* there must not be any panics (any error in data -> err result)
* any duplicated data, e.g. if some data is somehow duplicated in the data/metadata
  * we should minimize that in v2 to make our metadata more concise
  * in v0x01, or if we decide we need some duplication -- it must match and it is an error otherwise

PR #874, opened 2026-02-17
fix(rust): handle vertex_offsets in GeoJSON geometry conversion

nyurik:
vibed, geojson parsing was crashing for some mlts, passes now and uses a nicer helper

The `to_geojson` method ignored the `vertex_offsets` indirection layer, causing index-out-of-bounds panics on real-world MLT tiles that use dictionary-encoded vertices. When `vertex_offsets` is present, topology offsets must be mapped through it to get actual vertex buffer positions.

Also replace all direct array indexing with bounds-checked `.get()` calls that return descriptive errors instead of panicking, and add an integration test covering all 134 MLT files in `test/expected/`.

### DIFF
--- a/rust/mlt-nom/README.md
+++ b/rust/mlt-nom/README.md
@@ -9,7 +9,7 @@ This approach allows us to easily extend the format in the future by adding new 
 Note the ordering -- `tag` is after the `size` because it is possible to treat it as a single byte for now until the parser supports more than 127 types, and can efficiently skip unknown ones without doing a more expensive varint parsing.
 
 ## Layer 0x01 - MVT compatibility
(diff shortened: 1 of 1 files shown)


rust/mlt-nom/src/v01/property.rs
```diff
@@ -359,31 +358,34 @@ fn raw_bytes(s: Stream<'_>) -> Vec<u8> {
...
-fn decode_plain_strings(lengths: &[u32], data: &[u8]) -> Vec<String> {
+/// Split `data` into string slices using `lengths` as byte lengths for each entry.
+fn split_strings_by_lengths<'a>(lengths: &[u32], data: &'a [u8]) -> Result<Vec<&'a str>, MltError> {
     let mut strings = Vec::with_capacity(lengths.len());
     let mut offset = 0;
     for &len in lengths {
         let len = len as usize;
-        strings.push(String::from_utf8_lossy(&data[offset..offset + len]).into_owned());
+        let s = str::from_utf8(&data[offset..offset + len])
+            .map_err(|e| MltError::DecodeError(format!("invalid UTF-8 in string: {e}")))?;
```
CommanderStorm: should be a proper error

nyurik: good catch

PR #876, opened 2026-02-17
feat(rust): add MLT TUI

nyurik:
```bash
cd rust/mlt-nom
cargo run ui ../../test/expected
```

image


CommanderStorm (approved):
least maintainable code ever.

I tried to make it at least a bit less horrible, still some way to go

PR #877, opened 2026-02-18
feat(rust): pre-Stream serialiser

CommanderStorm:
These are the remaining missing serialisers:
- geometry
- stream (regular, bitvec-optional, boolean)

### DIFF
--- a/rust/mlt-nom/src/v01/geometry/mod.rs
+++ b/rust/mlt-nom/src/v01/geometry/mod.rs
@@ -5,6 +5,7 @@ use std::io::Write;
 use std::ops::Range;
 
 use borrowme::borrowme;
+use integer_encoding::VarIntWriter as _;
 use num_enum::TryFromPrimitive;
 
 use crate::MltError;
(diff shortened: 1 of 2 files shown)


rust/mlt-nom/src/v01/geometry/mod.rs
```diff
@@ -51,7 +53,7 @@ impl Analyze for Geometry<'_> {
 impl OwnedGeometry {
     pub(crate) fn write_columns_meta_to<W: Write>(&self, writer: &mut W) -> Result<(), MltError> {
         match self {
-            Self::Raw(r) => r.write_columns_meta_to(writer),
+            Self::Raw(_) => OwnedRawGeometry::write_columns_meta_to(writer),
```
nyurik, line 56: this is weird - wouldn't serializer need to have the instance to serialize, even if not implemented yet

CommanderStorm: this is just the metadata, for geometry this does not depend on self.

the actual body comes later

CommanderStorm: image

PR #879, opened 2026-02-18
refactor: `Analysis` trait-fn naming `decoded` -> `collect_statistic`

CommanderStorm:
it is kind of misinformative what it actually does otherwise

### DIFF
--- a/rust/mlt-nom/src/cli/ui/mod.rs
+++ b/rust/mlt-nom/src/cli/ui/mod.rs
@@ -29,6 +29,7 @@ use crate::cli::ls::{FileSortColumn, LsRow, MltFileInfo, analyze_mlt_files};
 use crate::cli::ui::rendering::files::{
     render_file_browser, render_file_filter_panel, render_file_info_panel,
 };
+use crate::cli::ui::rendering::help::render_help_overlay;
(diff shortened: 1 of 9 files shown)


rust/mlt-nom/src/cli/ls.rs
```diff
@@ -210,9 +210,9 @@ fn analyze_mlt_file(path: &Path, base_path: &Path) -> Result<MltFileInfo> {
     for layer in &mut layers {
         layer.decode_all()?;
         if let Some(layer01) = layer.as_layer01() {
-            data_size += layer01.decoded(StatType::PayloadDataSizeBytes);
-            meta_size += layer01.decoded(StatType::MetadataOverheadBytes);
-            feature_count += layer01.decoded(StatType::FeatureCount);
+            data_size += layer01.decoded_statistics_for(StatType::PayloadDataSizeBytes);
```
nyurik: too verbose... how about `compute(DecodedDataSize)` ?

CommanderStorm: too generic!

how about

```suggestion
            data_size += layer01.collect_statistic(PayloadDataSizeBytes);
```

PR #885, opened 2026-02-18
feat(cpp): make feature ids optional, use mlt-cpp-json for synthetic tests of C++ decoder

louwers:
- Creates a new JavaScript package `test/synthetic/synthetic-test-tool` to share the synthetic test logic between the TypeScript and C++ decoder.
- Adds script that calls out to `mlt-cpp-json` (which uses the C++ decoder) and output JSON.
- Because there is JavaScript in different parts of the project, I added Biome to the pre-commit config to format JavaScript/TypeScript across the project (except the `ts` directory which uses prettier).
- Some changes are made to the C++ decoder to pass the synthetic tests. Most notably doubles are now encoded as doubles and feature IDs are made optional. The other changes are bug fixes.

### DIFF
--- a/test/synthetic/synthetic-test-tool/index.ts
+++ b/test/synthetic/synthetic-test-tool/index.ts
@@ -1,4 +1,4 @@
-import { glob, readFile } from "node:fs/promises";
+import { glob, readFile, writeFile } from "node:fs/promises";
 import { basename, join } from "node:path";
 import JSON5 from "json5";
 import { expect } from "vitest";
@@ -34,6 +34,19 @@ expect.addEqualityTesters([
 export class SyntheticTestRunner {
   shouldSkip(_testName: string): false | string {
     return false;
+  }
+
+  private async writeActualOutput(
+    mltFile: string,
+    actual: Record<string, unknown>,
+  ) {
+    const actualFile = mltFile.replace(/\.mlt$/, ".actual.json");
+    await writeFile(
+      actualFile,
+      `${JSON5.stringify(actual, null, 2)}\n`,
+      "utf-8",
+    );
+    return actualFile;
   }
 
   decodeMLT(_mltFilePath: string): Promise<Record<string, unknown>> {
@@ -77,12 +90,8 @@ export class SyntheticTestRunner {
         passed++;
       } catch (_err) {
         console.log(`FAIL - ${name}`);
-        console.error(
-          "expected:\n",
-          JSON5.stringify(expected, null, 2),
-          "\nactual:\n",
-          JSON5.stringify(actual, null, 2),
-        );
+        const actualFile = await this.writeActualOutput(mltFile, actual);
+        console.log(`wrote actual output to ${actualFile}`);
         failed++;
       }
     }
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.gcov
 *.log
 *.new.geojson
+*.actual.json
 .DS_Store
 .cache/
 .idea/


cpp/tool/test-synthetic.mjs
```diff
@@ -0,0 +1,81 @@
...
+function normalizeFloats(val) {
+  if (typeof val === "number") return Math.fround(val);
+  if (Array.isArray(val)) return val.map(normalizeFloats);
+  if (val && typeof val === "object") {
+    return Object.fromEntries(Object.entries(val).map(([k, v]) => [k, normalizeFloats(v)]));
+  }
+  return val;
+}
+
+// Expected JSON files that contain NaN or Infinity (not valid JSON)
```
CommanderStorm: please use json5 instead of json as a dependency
our files are json5

cpp/tool/synthetic-geojson.cpp
```diff
@@ -0,0 +1,93 @@
+#include "synthetic-geojson.hpp"
```
louwers, line 1: Maybe we can get rid of the original MLT to JSON serialization code.

There seems to be some overlap.

cpp/tool/mlt-json.cpp
```diff
@@ -56,7 +57,13 @@ int main(int argc, char** argv) {
 
     mlt::Decoder decoder;
     const auto tileData = decoder.decode({buffer.data(), buffer.size()});
-    const auto tileJSON = mlt::json::toJSON(tileData, {.x = x, .y = y, .z = z}, geoJSON);
+
+    nlohmann::json tileJSON;
+    if (geoJSON) {
+        tileJSON = mlt::json::toJSON(tileData, {.x = x, .y = y, .z = z}, true);
+    } else {
+        tileJSON = mlt::test::toFeatureCollection(tileData);
```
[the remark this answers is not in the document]
louwers: Nobody depends on it.

cpp/tool/test-synthetic.mjs
```diff
@@ -0,0 +1,81 @@
...
+
+console.log(`Found ${mltFiles.length} MLT files\n`);
+
+let passed = 0;
+let failed = 0;
+let skipped = 0;
+
+for (const mltFile of mltFiles) {
+  const name = mltFile.split("/").pop();
+  const jsonFile = mltFile.replace(/\.mlt$/, ".json");
```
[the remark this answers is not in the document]
louwers: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/maplibre/maplibre-tile-spec/pull/885#discussion_r2823203437)

cpp/tool/mlt-json.cpp
```diff
@@ -56,7 +57,13 @@ int main(int argc, char** argv) {
...
+    }
```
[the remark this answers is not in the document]
louwers: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/maplibre/maplibre-tile-spec/pull/885#discussion_r2823203455)

cpp/src/mlt/decoder.cpp
```diff
@@ -93,13 +93,15 @@ Layer Decoder::Impl::parseBasicMVTEquivalent(BufferStream& tileData) {
...
+    const auto featureCount = geometries.size();
+    if (!ids.empty() && ids.size() != featureCount) {
+        throw std::runtime_error("ID count (" + std::to_string(ids.size()) + ") does not match geometry count (" +
+                                 std::to_string(featureCount) + ")");
     }
 
     return util::generateVector<Feature>(featureCount, [&](const auto i) {
-        return Feature{ids[i], std::move(geometries[i]), static_cast<std::uint32_t>(i)};
+        const auto id = ids.empty() ? Feature::id_t{0} : ids[i];
+        return Feature{id, std::move(geometries[i]), static_cast<std::uint32_t>(i)};
```
CommanderStorm, line 124: Assigning zero as an Id is a fix for this, but not sure if it is the correct fix.

cpp/tool/test-synthetic.mjs
```diff
@@ -0,0 +1,81 @@
...
+
+// Expected JSON files that contain NaN or Infinity (not valid JSON)
+const SKIP_NAN_INF = new Set([
+  "prop_f32_nan.mlt",
+  "prop_f32_neg_inf.mlt",
+  "prop_f32_pos_inf.mlt",
+  "prop_f64_nan.mlt",
+  "prop_f64_neg_inf.mlt",
+  "prop_f64_max.mlt",
+]);
```
CommanderStorm: our test fixtures are JSON5
Please use a node library like I am using in rust or ts instead of ignroing the tests ^^

cpp/tool/test-synthetic.mjs
```diff
@@ -0,0 +1,81 @@
...
+
+// FastPFOR-encoded tiles (requires MLT_WITH_FASTPFOR=ON at build time)
+const SKIP_FASTPFOR = new Set([
+  "polygon_fpf.mlt",
+  "polygon_hole_fpf.mlt",
+  "polygon_morton_tes.mlt",
+  "polygon_multi_fpf.mlt",
+]);
```
CommanderStorm, line 20: can't we enable it on the build side of things?

That would enable us to get to the real underlying issues

louwers: Blocked by https://github.com/maplibre/maplibre-tile-spec/pull/817/changes

ts/src/synthetic-test-util.ts
HarelM, line 1: Adding the `addEqualityTesters` for every comparison seems a bit weird, and might affect performance, but mostly weird.
I would recommend the following instead in order to share the comparison code.
Before running the test, run something that would generate the json files.
When this completes, use vitest to run the test itself.
You can even use `npx vitest compare-jsons.ts` - `compare-jsons.ts` would be the name of the new script both platforms will run (or can be made for all platforms as we will simply compare json files).
This way there's no need to write something similar to the test harness that's written here.
The folder where the output json files are created can be either hardcoded for all platforms or use an ENV variable.

What do you think?

Parsing the code and comparing it sounds like something that should be shared, so this is a step in the right direction I believe, just needs a bit refinement.

CommanderStorm:
@louwers
Could we please merge this?
I would not know how many tests on the cpp side my recent PR with geometry generation synthetics would break.

louwers:
@CommanderStorm Let me make some final changes.

ts/src/synthetic-test-util.ts
louwers: I removed the Vitest dependeny because it is a bit overkill for the C++ synthetic tests.

I put everything in a shared package for the synthetic tests.

Please take a look if it's OK for you.

cpp/src/mlt/decode/property.hpp
```diff
@@ -188,6 +187,15 @@ class PropertyDecoder {
                 checkBits(presentStream, result);
                 return {scalarType, std::move(result), std::move(presentStream)};
             }
+            case ScalarType::DOUBLE: {
```
louwers, line 190: @CommanderStorm @TimSylvester Did we change how doubles are encoded? This seems to be needed right now.

cpp/tool/test-synthetic.mjs
```diff
@@ -0,0 +1,38 @@
...
+	"polygon_fpf",
+	"polygon_hole_fpf",
+	"polygon_morton_tes",
+	"polygon_multi_fpf",
+	"polygon_fpf_tes",
+]);
+
+class SyntheticTestRunnerCpp extends SyntheticTestRunner {
+	shouldSkip(testName) {
+		// if (SKIP_NAN_INF.has(testName)) return "NaN/Infinity not valid JSON";
```
louwers:
```suggestion
```

cpp/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "module",
+  "dependencies": {
+    "json5": "^2.2.3",
+    "synthetic-test-tool": "file:../test/synthetic/synthetic-test-tool"
```
HarelM, line 5: I would condider removing this and simply import the relevant file...

test/synthetic/synthetic-test-tool/index.ts
HarelM, line 1: This replaces like 20 lines with ~150 lines just to remove vitest as a dependency? Seems a bit of an overkill to me, especially since there is an internal like dependency, so I find this solution a bit weird...

cpp/package.json
louwers: Considered it, but I like this package to manage its own dependencies...

test/synthetic/synthetic-test-tool/index.ts
louwers: Reverted.

louwers:
@HarelM I don't want to change it to output files, and then compare those, because we have different encoders so we would need to use different output directories which is messy and awkward.

That said, I agree that it is great for debugging. So I changed it so that it writes out `[testname].actual.json` files on unexpected output.

test/synthetic/synthetic-test-tool/index.ts
```diff
@@ -0,0 +1,125 @@
...
+        failed++;
+        continue;
+      }
+
+      const expectedRaw = await readFile(jsonFile, "utf-8");
+      const expected = JSON5.parse(expectedRaw);
+
+      try {
+        expect(actual).toEqual(expected);
+        console.log(`OK - ${name}`);
```
HarelM, line 89: Can all these console log be removed? I think vitest can take care of that, can't it?

louwers: @HarelM No, because as far as Vitest is concerned it is just a single test that calls `run()`.

If you want I can return some kind of iterator so you can let Vitest handle it on the JS side.

louwers: Making a follow-up PR.

HarelM:
I think the test output files can use a {testname}.{language/runner}.json to avoid file name collision.
They can be ignored, and you could even be able to see differences in different language failures, although I'm sure no one actually will do it.
Having some preprocessing code output json files and later on run the test harness to compare can make the comparison code exactly the same and language agnostic.

test/synthetic/synthetic-test-tool/index.ts
HarelM: Check out my last comment, I really think it can solve a lot of quirkiness...

HarelM: Maybe even use some kind of `beforeAll` to generate these json files.

louwers:
@HarelM The comparison code is already language agnostic, it's only the encoding part that is not. I prefer having a class as a contract between the test different implementations, rather than some particular way the encoded MLTs need to be written out.

I also like to avoid cluttering the source tree with test or build artifacts.

I also think the more functional approach of `expectEquals(actual, expected)` is cleaner than the imperative `writeOutABunchOfFiles(); compareABunchOfFiles()`.

HarelM:
I think the render tests (at least in maplibre-gl-js) are a great example of how placing files can help understanding the problem and allowing good debug, so there is a merit to writing files, even if it cluttering the source tree, especially if they are git ignored (I sometime can use these files to rename them in order to see the diff using vscode diff tool which is a great advantage).
There's some redundant code from my point of view of logging skipped tests, counting how many have been successful etc which should be handled by a test harness, and not "written be hand", IMHO.
But I respect your professional judgment so I'll let you decide how to proceed with this.

louwers:
@HarelM I'm working on that. I will let each platform decide how to report on test fail/skip/success.

> how placing files can help understanding the problem and allowing good debug,

Agreed, that's how I am writing out actual output now. But it's just for debugging, not for doing the comparison.

PR #888, opened 2026-02-18
refactor(rust): rename `mlt-nom` -> `mlt`

CommanderStorm:


nyurik (commented):
lets make the working directory in CI be  `./rust` rather than a subdir of that. This way we will start fully testing all of our rust code in a workspace, and if needed, we can always delist some archived project in the root cargo.toml.  Consider this PR approved, just do this change before merging (i don't want to actually click approve because that would auto-merge)

PR #891, opened 2026-02-19
ci: fix rust ci not running

CommanderStorm:
this merges our two justfiles by the outer one calling the inner one instead of having duplicates.

It also fixes a few pathing issues which I did not catch before


justfile
```diff
@@ -57,7 +57,7 @@ env-info-java:
     ./gradlew --version
 
 # Run all formatting in every language
-fmt: fmt-java fmt-js fmt-rust
+fmt: fmt-rust fmt-java fmt-js
```
CommanderStorm, line 60: ordered by amount of execution time, not 4s, 10s, 0s

PR #893, opened 2026-02-19
feat(rust): MVT tile visualizer

nyurik:
needs a bit of post-vibe work

### DIFF
--- a/.github/actions/mlt-setup-rust/action.yml
+++ b/.github/actions/mlt-setup-rust/action.yml
@@ -1,0 +1,12 @@
+name: mlt-setup-rust
+description: Setup Rust requirements
+
+inputs: {}
+
+runs:
+  using: "composite"
+  steps:
+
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):

CommanderStorm (approved):
I don't like hinging more of the logic on the geojson part.
It is likely fine though


One thing you have is this:
```rust
 _ => Geometry::Point([0, 0]),
```

could you add a comment why we have this line 😉

nyurik:
> ```rust
>  _ => Geometry::Point([0, 0]),
> ```
> 
> could you add a comment why we have this line 😉

@CommanderStorm, are you saying you have a problem with the [Null Island](https://en.wikipedia.org/wiki/Null_Island)?!?  :rofl:

CommanderStorm:
No, just that I want to know that you have traced trhough the idea of making this a warning and why this is invalid data instead

PR #895, opened 2026-02-19
ci(rust): publishing `mlt` and other misc ci improvements

CommanderStorm:
This PR bringes cargo-dist support.
Lets see how this works out


rust/Cargo.toml
```diff
@@ -1,11 +1,13 @@
 [workspace]
 resolver = "3"
-members = ["mlt", "mlt-core", "mlt-legacy"]
-exclude = ["mlt-core/fuzz"]
+members = ["mlt", "mlt-core"]
+exclude = ["mlt-core/fuzz", "mlt-legacy"]
```
CommanderStorm: lets not waste time debugging this:

thread 'main' (6666) panicked at mlt-legacy/build.rs:17:10:
  called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "protoc failed: mlt_tileset_metadata.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.\n" }

.github/actions/mlt-setup-rust/action.yml
```diff
@@ -7,6 +7,19 @@ runs:
   using: "composite"
   steps:
 
-    - name: Install protobuf compiler for MVT support
+    - name: Install protobuf compiler (Ubuntu)
```
nyurik, line 10: Is there an action we can use to install protobuf? I'm sure someone already had to deal with all this logic

.github/workflows/rust.yml
```diff
@@ -102,10 +102,63 @@ jobs:
...
+  build-release-binaries:
+    name: Build release binaries for ${{ matrix.target }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+            artifact_name: mlt-linux-x86_64.tar.gz
+          # Requires mlt > fastpfor to support apple.
```
CommanderStorm:
```suggestion
          # Requires mlt > fastpfor to support arm via the existing simde code
```

PR #896, opened 2026-02-19
ci(rust): Split CI for better debuggability

CommanderStorm:
This PR splits our CI and makes a few metadata improvements that are unrelated.
The metadata changes are requred for cargo-dist, but otherwise mundane so  in this PR


nyurik:
on the second thought, this seems really redundant:

* clippy run automatically includes check run -- it is the same infrastructure as far as i know, so any check warnings will show when running clippy
* test also includes check run
* running check and then clippy/test automatically reuses the internal build artifacts

so in short - doing it all in one job seems far better for performance reasons

PR #901, opened 2026-02-19
build(java): Use gradle version catalogs to centralize dependencies declaration for java projects

### LINKED ISSUE #887: Simplify java dependencies management using gradle version catalog
Wondering if it's worth using gradle's version catalog to simplify the dependencies declaration for the gradle project ?
https://docs.gradle.org/current/userguide/version_catalogs.html

I've noticed currently the java dependencies are declared inline between the subprojects and duplicate dependencies declared in each project's `build.gradle`

Using the version catalog it would simplify a central point for all dependency delcaration for the project.

sonalranjit:
Introduces gradle's version catalog for declaring all java dependencies used across the various java subprojects.
https://docs.gradle.org/current/userguide/version_catalogs.html

Version catalog centralizes the version management of all the java dependencies, all the dependecies declared now in `gradle/libs.versions.toml` is accessible by all the projects under `java`.

Closes #887


sonalranjit:
Thanks @CommanderStorm feel free to merge. I dont think I have write access to merge.

CommanderStorm:
I'd like a cursorary look over this by barry since he is the primary consumer of this (thanks for that work) and I don't fully know how to test that this does not break his tooling, even though I don't think it should.

We can also merge and see if it breaks something -> revert if it does.

PR #903, opened 2026-02-19
test(rust): improve the fuzzabiltiy and add fuzzing to the CI

CommanderStorm:


rust/mlt-core/src/layer/v01/geometry/mod.rs
```diff
@@ -50,14 +50,16 @@ impl Analyze for Geometry<'_> {
 }
 
 impl OwnedGeometry {
-    pub(crate) fn write_columns_meta_to<W: Write>(&self, writer: &mut W) -> Result<(), MltError> {
+    #[doc(hidden)]
+    pub fn write_columns_meta_to<W: Write>(&self, writer: &mut W) -> Result<(), MltError> {
```
CommanderStorm, line 54: I need this and the others to be `pub` for the fuzzer to enable better debug encoding, but I am not sure they add any value for others, so not public API.

If someone asks, we can make them public api.

rust/mlt-core/src/layer/v01/stream/physical.rs
```diff
@@ -24,6 +24,10 @@ impl PhysicalStreamType {
         let high4 = value >> 4;
         let low4 = value & 0x0F;
         Some(match high4 {
+            #[cfg(fuzzing)]
+            // when fuzzing, we cannot have ignored bits, to preserve roundtripa-biltiy
+            0 if low4 == 0 => PhysicalStreamType::Present,
+            #[cfg(not(fuzzing))]
```
CommanderStorm, line 30: a bit ugly to sprinkle these tiny `cfg(fuzzings)` into the codebase when we have something that messes with roundtrip-ability (not sure if spelled correctly).

I don't think this is usefull in terms of an actual `cfg(feature=..)`

PR #905, opened 2026-02-20
feat(rust): add encoding utilitys for symetry with the decoders

CommanderStorm:
lets build out a base for an encoding system to stand on.

Currently ony tested for empty and symmetry, could not come up with better tests for this.
Since decode is already anchored, the tests are likely complete.


rust/mlt-core/src/errors.rs
```diff
@@ -94,6 +94,8 @@ pub enum MltError {
     FastPforFfi(#[from] Exception),
     #[error("invalid RLE run length (cannot convert to usize): value={0}")]
     RleRunLenInvalid(i128),
+    #[error("invalid RLE run length (cannot convert to usize): value={0}")]
+    RleRunCountInvalid(usize),
```
CommanderStorm:
```suggestion
```

rust/mlt-core/src/utils/decode.rs
```diff
@@ -171,15 +171,45 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_decode_zigzag_delta_empty() {
+        assert!(decode_zigzag_delta::<i32, i32>(&[]).is_empty());
+    }
+
+    #[test]
+    fn test_decode_rle_empty() {
+        assert!(decode_rle::<u32>(&[], 0, 0).unwrap().is_empty());
```
CommanderStorm, line 201: do note that these decoding functions are not as defensive as the rest of the code.

Passing a value larger than 0 here for example panics.
Can fix this here or in a future PR.
I have not audited if this is actualyl a problem in the code that uses this.

rust/mlt-core/src/utils/encode.rs
```diff
@@ -9,3 +13,195 @@ pub fn encode_str(data: &mut Vec<u8>, value: &[u8]) {
...
+
+    (runs, values)
+}
+
+/// Encode byte-level RLE as used in ORC for boolean and present streams.
+///
+/// Format: control byte determines the run type:
+/// - `control >= 128`: literal run of `(256 - control)` bytes follow
+/// - `control < 128`: repeating run of `(control + 3)` copies of the next byte
+pub fn encode_byte_rle(data: &[u8]) -> Vec<u8> {
```
CommanderStorm, line 69: a bit more complicated than I would like, but likely works for now.

nyurik (approved):

PR #906, opened 2026-02-20
feat(rust): Id encoder impelmentation

CommanderStorm:
This is the most simple of the decoders to write, but progress.

Same as before, tests for symetry but is not plumped in usefully.

### DIFF
--- a/rust/mlt-core/src/layer/v01/id.rs
+++ b/rust/mlt-core/src/layer/v01/id.rs
@@ -242,9 +242,9 @@ fn apply_present_bitset(
     Ok(result)
 }
 
-/// ID encoding configuration
+/// How to encode IDs
 #[derive(Debug, Clone, Copy)]
-pub enum IdEncodingConfig {
+pub enum IdEncodingType {
     /// 32-bit encoding
     Id32,
     /// 32-bit encoding with nulls
@@ -257,10 +257,10 @@ pub enum IdEncodingConfig {
 
 impl ToRaw<'_> for OwnedRawId {
     type Output = DecodedId;
-    type Config = IdEncodingConfig;
+    type Config = IdEncodingType;
 
-    fn to_raw(decoded: &DecodedId, config: IdEncodingConfig) -> Result<Self, MltError> {
-        use IdEncodingConfig as CFG;
+    fn to_raw(decoded: &DecodedId, config: IdEncodingType) -> Result<Self, MltError> {
+        use IdEncodingType as CFG;
 
         // skipped one level higher
         let DecodedId(Some(ids)) = decoded else {
@@ -347,7 +347,7 @@ impl ToRaw<'_> for OwnedRawId {
 
 #[cfg(test)]
 mod tests {
-    use IdEncodingConfig::*;
+    use IdEncodingType::*;
     use proptest::prelude::*;
     use rstest::rstest;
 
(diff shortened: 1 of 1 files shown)


rust/mlt-core/src/layer/v01/id.rs
```diff
@@ -220,76 +242,349 @@ fn apply_present_bitset(
...
 
-        let present_meta = StreamMeta {
-            physical_type: PhysicalStreamType::Present,
-            num_values: 3, // 3 features total
-            logical_decoder: LogicalDecoder::None,
-            physical_decoder: PhysicalDecoder::None,
-        };
+    #[cfg(test)]
+    mod proptests {
+        use proptest::prelude::*;
```
nyurik: did you mean to make this as a sub-module under `mod tests`?  With a dup `cfg()`?

rust/mlt-core/src/layer/v01/id.rs
```diff
@@ -220,76 +242,349 @@ fn apply_present_bitset(
...
+            #[test]
+            fn test_correct_variant_produced_id64(
+                ids in prop::collection::vec(any::<u64>(), 1..50)
+            ) {
+                let ids_u64: Vec<Option<u64>> = ids.iter().map(|&id| Some(id)).collect();
+                assert_produces_correct_variant(ids_u64, Id64)?;
+            }
+        }
+
+        enum ExpectedVariant {
```
nyurik: if enum is identical, why do you need to dup it?

rust/mlt-core/src/utils/mod.rs
```diff
@@ -1,5 +1,5 @@
 mod encode;
-//pub(crate) use encode::*;
+pub use encode::*;
```
nyurik, line 2: do we really need to use wildcards? I thought there was a lint for that

CommanderStorm: lets move that to another PR

rust/mlt-core/src/layer/v01/id.rs
```diff
@@ -220,76 +242,298 @@ fn apply_present_bitset(
...
-            "Decoded IDs should match expected values"
-        );
+    #[rstest]
+    #[case::id32_basic(Id32, &[Some(1), Some(2), Some(100), Some(1000)])]
+    #[case::id32_single(Id32, &[Some(42)])]
+    #[case::id32_boundaries(Id32, &[Some(0), Some(u64::from(u32::MAX))])]
+    #[case::id64_basic(Id64, &[Some(1), Some(2), Some(100), Some(1000)])]
+    #[case::id64_single(Id64, &[Some(u64::MAX)])]
+    #[case::id64_boundaries(Id64, &[Some(0), Some(u64::MAX)])]
+    #[case::id64_large_values(        Id64,        &[Some(0), Some(u64::from(u32::MAX)), Some(u64::from(u32::MAX) + 1), Some(u64::MAX)])]
```
nyurik: spacing?

CommanderStorm: I thought cargo fmt would shurely fix that ^^

rust/mlt-core/src/layer/v01/id.rs
```diff
@@ -220,76 +242,298 @@ fn apply_present_bitset(
     Ok(result)
 }
 
+/// ID encoding configuration
+#[derive(Debug, Clone, Copy)]
+pub enum IdEncodingConfig {
```
nyurik: Why `Config` here?  Maybe `IdType` ?

CommanderStorm: It is only for How to encode.. so likely a config-ish thing.

How about `IdEncodingType`

PR #908, opened 2026-02-20
chore(ci): cleanup ci workflows

nyurik:


.github/workflows/ts.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - run: just -v ts::test
       - run: just -v ts::build
       - name: Codecov upload
-        if: '!cancelled()'
+        if: "github.repository_owner == 'maplibre' && !cancelled()"
```
HarelM, line 33: Im not sure this is needed, worst case it will upload to codecov for a different repo, it won't show for this repo.
But I don't mind either way...

HarelM (approved):

.github/workflows/ts.yml
nyurik: why do we have `!cancelled` here?  If the CI is canceled, it will simply stop, no biggie

HarelM: The idea is that even if the tests fail we want to upload the coverage report.

nyurik: how valuable are the coverage results if tests don't pass???

HarelM: Can be valuable if only one test falls for example...

PR #918, opened 2026-02-21
refactor(rust): Shift id encoding to the stream

CommanderStorm:
This PR adds some stubs for @nyurik to look into the geometry encoding part if he has time later in the day/weekend.

This PR does:
- sort the `decode_*` method in a sensible order
- add stubs for `encode_*`
- de-generifiy `decode_unsigned_int_stream` (I don't think this was saving code here, just mental overhead)
- add currently ignored tests (so they can be gradually enabled again)

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1162,7 +1162,6 @@ dependencies = [
  "insta",
  "integer-encoding",
  "json5",
- "log",
  "mvt-reader",
  "num-traits",
  "num_enum",
--- a/rust/mlt-core/Cargo.toml
+++ b/rust/mlt-core/Cargo.toml
@@ -17,7 +17,6 @@ fastpfor.workspace = true
 geo-types.workspace = true
 hex.workspace = true
 integer-encoding.workspace = true
-log = "0.4.29"
 mvt-reader.workspace = true
 num-traits.workspace = true
 num_enum.workspace = true


rust/mod.just
```diff
@@ -96,6 +96,8 @@ clean:
 
 # Run cargo clippy to lint the code
 clippy *args:
+    #!/usr/bin/env bash
```
nyurik, line 99: why is this needed?

rust/mlt-core/src/errors.rs
```diff
@@ -51,7 +53,7 @@ pub enum MltError {
     #[error("unsupported logical decoder {0:?} for {1}")]
     UnsupportedLogicalDecoder(LogicalDecoder, &'static str),
     #[error("unsupported combination of logical techniques: {0:?} + {1:?}")]
-    UnsupportedLogicalTechnique(LogicalTechnique, LogicalTechnique),
+    UnsupportedLogicalTechniqueCombination(LogicalTechnique, LogicalTechnique),
```
nyurik, line 56: a bit too long, perhaps `InvalidLogicalEncoding` ?

rust/mlt-core/src/utils/serialize.rs
```diff
@@ -3,16 +3,15 @@ use std::io::Write;
...
 use crate::v01::{OwnedStream, OwnedStreamData};
 
 pub trait BinarySerializer: Write + VarIntWriter {
     fn write_u8(&mut self, value: u8) -> io::Result<()> {
         self.write_all(&[value])
     }
     fn write_string(&mut self, value: &str) -> io::Result<()> {
-        let size =
-            u64::try_from(value.len()).map_err(|_| io::Error::other(MltError::IntegerOverflow))?;
+        let size = u64::try_from(value.len()).map_err(|e| io::Error::other(TryFromIntError(e)))?;
```
nyurik, line 14: perhaps a generic helper with the int restriction?

rust/mod.just
CommanderStorm: Otherwise fails will show multiple times, annoying to have twice the log

nyurik: what do you mean it shows multiple times?

CommanderStorm: cargo clippy shows erros in sub-crates (correctly, it could not build otherwise!).

If there is an error in martin-core, this means it is currently printed twice, which is a bit much. lets only print one error.

PR #922, opened 2026-02-21
test(ts): Add synthetic tests for TS, fix nullable issues for IDs decoding

HarelM:
~I don't think this is a good enough solution, but it does solve the problem for the failing tests related to ids with nullable vector.~
Based mostly on this one:
- https://github.com/maplibre/maplibre-tile-spec/pull/863

~I'll continue looking into it.~
CC: @CommanderStorm @Turtelll

### DIFF
--- a/rust/mlt-core/src/errors.rs
+++ b/rust/mlt-core/src/errors.rs
@@ -3,9 +3,7 @@ use std::convert::Infallible;
 use fastpfor::cpp::Exception;
 use num_enum::TryFromPrimitiveError;
 
-use crate::v01::{
-    GeometryType, LogicalDecoder, LogicalTechnique, PhysicalDecoder, PhysicalStreamType,
-};
+use crate::v01::{GeometryType, LogicalCodec, LogicalTechnique, PhysicalCodec, PhysicalStreamType};
 
 pub type MltRefResult<'a, T> = Result<(&'a [u8], T), MltError>;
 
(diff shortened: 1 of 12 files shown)


HarelM:
Tests are now passing, but I'm not sure I like the fragility of the code that I have added.
I'll look into improving that.
@Turtelll if you have a better idea how to handle the nullability stuff do let me know.

ts/src/mltDecoder.ts
```diff
@@ -100,7 +100,7 @@ export default function decodeTile(
                 }
 
                 const idDataStreamMetadata = decodeStreamMetadata(tile, offset);
-                numFeatures = idDataStreamMetadata.decompressedCount;
+                numFeatures = nullabilityBuffer ? nullabilityBuffer.size() : idDataStreamMetadata.decompressedCount;
```
CommanderStorm, line 104: 👀 is this nessesary? what happens without it?

HarelM: `idDataStreamMetadata.decompressedCount` returns 4 when I looked instead of 5
It returns only the number of non-null IDs (as far as I understand) - 4, which is not the same as nullability buffer size - 5.

CommanderStorm: makes sense.
How about a code comment?

```suggestion
                // decompressedCount is the count WITHOUT nulls, but we may have nulls
                numFeatures = nullabilityBuffer ? nullabilityBuffer.size() : idDataStreamMetadata.decompressedCount;
```

HarelM:
Ok, I moved the nullability vector down to the decoder, so I'm good with the current code.
Also did some minor refactoring to the sythentic tests.
Since there's a mechanism to skip test in case there's a problem with the decoder there's no problem adding more tests for other languages and adding to the skip list.

HarelM:
I'll merge this now and continue to remove the skipped tests as much as possible.

PR #924, opened 2026-02-21
feat(rust): impl stream encoding

CommanderStorm:
this is rather of a bit of a hacky/duplicated code, not sure how to make the line count go down, but too be fair with me am currently a bit tired

### DIFF
--- a/mkdocs/docs/specification.md
+++ b/mkdocs/docs/specification.md
@@ -113,45 +113,33 @@ All integers in metadata sections are `Varint`-encoded (for u32) or bit-packed (
 
 ```mermaid
 ---
-title: FeatureTableSchema
 config:
   class:
     hideEmptyMembersBox: true
+title: FeatureTableSchema
 ---
 classDiagram
-    note for LogicalScalarType "[EXPERIMENTAL]"
-    note for ComplexColumn "[EXPERIMENTAL]"
-    note for ComplexField "[EXPERIMENTAL]"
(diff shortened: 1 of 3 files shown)


rust/mlt-core/src/layer/v01/stream/logical.rs
```diff
@@ -41,6 +43,124 @@ pub enum LogicalCodec {
     PseudoDecimal,
 }
 
+impl LogicalCodec {
+    /// Logically encode `u32` values, returning the physically-stored sequence and the concrete decoder.
+    ///
+    /// [`RleMeta`] is derived from the actual data.
+    /// See [`LogicalValue::decode_u32`] for the reverse operation.
+    pub fn encode_u32(self, values: &[u32]) -> Result<(Vec<u32>, Self), MltError> {
```
nyurik: The more i look at it, the more i think this is a wrong API.  `LogicalCodec` is the "state" of the encoding, not an algorithm configuration. As such, we may need to rename it once again, and move this function out of it.

Something like this (name suggestions are very welcome):

```rust
fn delta_encode(values: &[f32]) -> Result<(Vec<u8>, LogicalCodec), MltError> {}
```

CommanderStorm: how about the API that I just added? 
That kind of an API is what I would need in the next PR anyhow.

PR #940, opened 2026-02-23
fix(ts): DOUBLE/Float64 wire format to match Java/Rust/C++ implementations

Turtelll:
TypeScript decoder failed 6 Float64 synthetic tests with error: byte length of Float64Array
should be a multiple of 8

Root cause: TS encoder/decoder incorrectly used Float64 (8 bytes) for DOUBLE wire format, while
Java/Rust/C++ implementations correctly use Float32 (4 bytes).


HarelM (approved):
Nice work!

HarelM:
@Turtelll 
Can you check if the following PR is better in terms of readability:
https://github.com/maplibre/maplibre-tile-spec/pull/786

It seems a lot simpler and I think it solves the same problem, doesn't it?

Turtelll:
I think with this PR we preserve the right way if we want to change to real Float64 support. But @keiya01's approach is for sure more simpler and fixes the same issue. I'm fine with either.

HarelM:
I tend to say that the other PR is handling this higher up, where the logic should be, but this PR changes how doubles are encoded, which can be "confusing" as you call a method that returns a Float64, but it doesn't actually use Float64 to decode.
So I think the other one makes more sense, keeping the double decode as a building block without "magic" that can be used higher up the decoder where the logic should happen.
Any chance you can open a PR with reverting this and adding the other one?
Sorry for the extra work...

PR #947, opened 2026-02-24
feat(rust): Synthetic generation

CommanderStorm:


nyurik (approved):

rust/mlt-core/src/layer/v01/geometry/encode.rs
```diff
@@ -288,19 +528,151 @@ pub fn encode_geometry(
...
+        items.push(encode_vertex_buffer(verts, config.vertex().physical)?);
     }
 
     Ok(OwnedEncodedGeometry { meta, items })
 }
 
+pub trait GeometryEncoder {
+    /// Encoding settings for the geometry types (meta) stream.
+    fn meta(&self) -> Encoder {
+        unreachable!()
```
michaelkirk: What's the idea behind the `unreachable`s? 

Is this just a TODO before moving out of draft status? Otherwise, why have the default impls at all?

CommanderStorm: The idea here is that we were not quite sure which of the geometrys require which stream.
If we are not sure, likely our users are also not.

So if points don't require ring offsts, is unrechable.

CommanderStorm: but yea, defintively a footgun and I removed it

nyurik: the bigger concern with this is that "multi-typed" (mixed) geometries in a layer get really messy - encoding uses one stream if it is mixed, and another if it is just one type.  So perhaps we may need to move away from the explicit configuration, and instead introduce a builder pattern for the geom encoder config

Issue #911, opened 2026-02-21
Encoder use cases


nyurik:
For encoders, we should keep in mind these scenarios we want to handle:
* simplest (synthetics):  manually drive the encoding process by dictating the encoding method for each stream/field/... 
  * this is used mostly for debugging
  * side benefit is that the encoder API is cleanly separated from the optimizer
* automatic:  for each field, pick the absolute best encoding possible
  * try all relevant encoders for each ID/Property/Geometry(?), but not in their combination - search space N x M , not (N x M)^2
  * (?) consider ID-based sorting (which tbh i am not sure is that valuable)
  * shared dictionaries might be considered
* profile-driven: do exhaustive encoding for a subset of tiles, generate a profile with the optimal combination, and use it for the rest of the tiles
  * (?) in some cases it may make sense for the profile to have two encodings per property, i.e. one encoding is a clear winner in one subset of  tiles, and another encoding elsewhere
  * profile may be fine-tuned manually

#### Java
* [ ] simplest
* [x] automatic
* [ ] profile-driven

#### Rust
* [x] simplest
* [x] automatic
* [x] profile-driven

PR #950, opened 2026-02-24
feat(rust): Stream encoding optimiser

CommanderStorm:
This is the medium strategy apprach.
Not n x m, but also not entirely static.

Works towards #911

### DIFF
--- a/rust/mlt-core/Cargo.toml
+++ b/rust/mlt-core/Cargo.toml
@@ -18,7 +18,6 @@ harness = false
 [dependencies]
 borrowme.workspace = true
 fastpfor.workspace = true
-probabilistic-collections.workspace = true
 fsst-rs.workspace = true
 geo-types.workspace = true
 hex.workspace = true
@@ -26,6 +25,7 @@ integer-encoding.workspace = true
 mvt-reader.workspace = true
 num-traits.workspace = true
 num_enum.workspace = true
+probabilistic-collections.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 strum.workspace = true


rust/mlt-core/src/layer/v01/stream/optimizer.rs
```diff
@@ -0,0 +1,502 @@
...
+        candidates
+            .iter()
+            .copied()
+            .min_by_key(|&enc| encoded_size_u32(data, enc))
+            .unwrap_or_else(Encoder::fastpfor)
+    }
+    pub fn min_size_encoding_u64s(candidates: &[Encoder], data: &[u64]) -> Encoder {
+        candidates
+            .iter()
+            .copied()
```
michaelkirk, line 170: Perf nit: could you do the copy *after* choosing rather than copying every element and choosing one?

rust/mlt-core/src/layer/v01/stream/optimizer.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+
+    /// Returns `true` if RLE is a sensible candidate based on this profile.
+    ///
+    /// Requires both a sufficient average run length (consecutive repetition)
+    /// and a low enough distinct ratio (global repetition) so that the
+    /// run-length and unique-value arrays stored by the encoder are small.
+    #[must_use]
+    fn rle_is_viable(&self) -> bool {
+        self.avg_run_length >= MIN_AVG_RUN_LENGTH && self.distinct_ratio < 1.0
```
michaelkirk: If distinct_ratio == 1.0, wouldn't avg_run_length also == 1? 

If I'm understanding the code correctly, it seems like the `distinct_ratio` check would be completely redundant here. Or can you provide an example where it'd make a difference?

rust/mlt-core/src/layer/v01/stream/optimizer.rs
```diff
@@ -0,0 +1,502 @@
...
+    /// A low `distinct_ratio` means most values are repeated, which is exactly
+    /// the case where a dictionary trades a small code-book for a much shorter
+    /// value stream.
+    #[must_use]
+    pub fn distinct_ratio(&self) -> f64 {
+        self.distinct_ratio
+    }
+}
+
+fn block_sample<T: Clone + Copy>(values: &[T], target: usize) -> Vec<T> {
```
michaelkirk: perf nit: based on current usage it seems like this could return a slice rather than allocating a vec, but maybe it's not performance sensitive.

rust/mlt-core/src/layer/v01/stream/optimizer.rs
```diff
@@ -0,0 +1,502 @@
...
+
+/// Sampling-based encoder selection
+///
+/// # Strategy
+///
+/// 1. **Profile**: Compute lightweight statistics over a representative sample
+///    of the data (average run length, sort order, max bit-width, distinct
+///    ratio) and use them to prune obviously unsuitable candidates early.
+///
+/// 2. **Compete**: Encode the same sample with every surviving candidate and
```
michaelkirk: For my own understanding, this "Compete" portion of the work is not included in this PR. Is that correct?

CommanderStorm: This is what I meant with this

```rust
candidates
    .iter()
    .copied()
    .min_by_key(|&enc| encoded_size_u64(data, enc))
    .unwrap_or_else(Encoder::fastpfor)
```

rust/mlt-core/src/layer/v01/stream/optimizer.rs
```diff
@@ -0,0 +1,502 @@
...
+            .copied()
+            .min_by_key(|&enc| encoded_size_u32(data, enc))
+            .unwrap_or_else(Encoder::fastpfor)
+    }
+    pub fn min_size_encoding_u64s(candidates: &[Encoder], data: &[u64]) -> Encoder {
+        candidates
+            .iter()
+            .copied()
+            .min_by_key(|&enc| encoded_size_u64(data, enc))
+            .unwrap_or_else(Encoder::fastpfor)
```
CommanderStorm:
```suggestion
        candidates
            .iter()
            .copied()
            .min_by_key(|&enc| encoded_size_u64(data, enc))
            .unwrap_or_else(Encoder::varint)
```

rust/mlt-core/src/layer/v01/stream/optimizer.rs
CommanderStorm: We are copying an 2-8 long vector of Copy'able enums.
Basically, this struct is one u8

I would need to benchmark this, but very unlikely that this actually can affect perf.

rust/mlt-core/src/layer/v01/stream/optimizer.rs
michaelkirk: 🤦 of course! 

Maybe worth linking to the method names for dummies like me.

rust/mlt-core/src/layer/v01/stream/optimizer.rs
CommanderStorm: Also the refence logic looks not so nice, I am going to stick with the current version

rust/mlt-core/src/layer/v01/stream/optimizer.rs
CommanderStorm: likely compliler can inlne this but good catch

rust/mlt-core/src/layer/v01/stream/optimizer.rs
CommanderStorm: None of the values here have been benchmarked, they are likely all fairly off from optimal.

That needs a bit more higher level infrastructure.

If all values in a slice are estimated to be unique, RLE does certainly not make sense.
The value here should likely be lower, but how much? need to data-mine this.

michaelkirk: Just in case there's misunderstanding, my point wasn't so much that 1.0 is the wrong value, but that `distinct_ratio` is the wrong metric, because it is not independent from run_length and it is strictly less precise.

The actual run length of the encoded data is what ultimately determines the efficacy of RLE encoding, right? 

`distinct_ratio` seems like a way to guess what the run length might be. At 1.0 and 0.0 you can actually use it precisely, in the middle you're guessing. But we already have the actual run_length metric. Adding `distinct_ratio` into the mix seems like a way to layer on something strictly less precise for no benefit.

That said, there seems little harm in leaving it in as an experiment if you'd like to try to tune it later.

CommanderStorm: Yea, BtrBlocks also does not use it, there might be a reason.

On the other hand, Max also did not have as many different string-dictionarys to support. 🤷🏻‍♂️ 
We will see.

PR #952, opened 2026-02-25
test(synthetics): fix morton naming being misleading

CommanderStorm:
This does not contain morton.

Rust does not implement mroton and we are bit for bit identical!


rust/mlt-synthetics/src/main.rs
```diff
@@ -75,10 +75,13 @@ fn generate_geometry(d: &Path) {
     geo_varint().geo(LineString(vec![C1, C2])).write(d, "line");
     geo_varint().geo(pol()).write(d, "polygon");
     geo_fastpfor().geo(pol()).write(d, "polygon_fpf");
-    geo_varint().tessellated(pol()).write(d, "polygon_tes");
+    geo_varint().tessellated(pol()).write(d, "polygon
+      
+      
+      ");
```
CommanderStorm:
```suggestion
    geo_varint().tessellated(pol()).write(d, "polygon_tes");
```

PR #960, opened 2026-02-25
test(synthetic): add Morton fixture synthetic test

DoFabien:
- Add `line_morton` synthetic fixture.
- In TS synthetic tests, any fixture with `morton` in its name must decode with `geometryVector.mortonSettings`.

And fix skip reason for `polygon_fpf_tes` to `FastPFor not implemented` (not Morton)


ts/src/synthetic.spec.ts
```diff
@@ -67,6 +67,10 @@ describe("MLT Decoder - Synthetic tests", () => {
             ]);
 
             const featureTables = decodeTile(mltBuffer, null, false);
+            if (testName.includes("morton")) {
```
HarelM: This code should not be here, no need to verify our test code, code coverage should take care of this.

HarelM:
@nyurik @louwers can you check the java code? I'm not familiar enough with the synthetic generation coding standards.

ts/src/synthetic.spec.ts
DoFabien: Understood, I’ll remove it.

HarelM:
@nyurik @CommanderStorm can you check the rust CI? I also think it falls on main...?

CommanderStorm:
No, not a flake. CI on main is green (-+ macos-intel runners being a bit flaky today).
I had a fail in the release-creation-automation, but nothing in CI.

This PR is differeent:
- The logical techniques here are `Morton` + `Delta`.
- We expected that only `Morton` + `None` would make sense, so only added this to the parser. If this combination is legal too, we should support it.

Here is the relevant code:

https://github.com/maplibre/maplibre-tile-spec/blob/9ef9ae4ed92972fa347522c48a4a85882c3a8491/rust/mlt-core/src/layer/v01/stream/mod.rs#L414-L450

You can simply add it to the ignore list, I or @nyurik will look at this tomorrow (after having a sleep)

```diff
  mod java {
      use super::*;
-    test_each_path! { for ["mlt", "json"] in "../test/synthetic/0x01" => pair_test }
+    test_each_path! { 
+        for ["mlt", "json"] in "../test/synthetic/0x01" => pair_test
+        ignore: {
+            "line_morton": "Morton + Delta logical encoding is not currently supported"
+        }
    }
}
````

Which logical1 + logical2 + physical technique are legal (and which meatadata comes after each) is a bit underdefined, so unclear what to implement how.

File: java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java [hunk 2]
```diff
@@ -6,6 +6,7 @@
 import java.io.IOException;
 import java.math.BigInteger;
 import java.nio.file.Files;
+import org.locationtech.jts.geom.Coordinate;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
```
CommanderStorm: faulty merge by me

```suggestion
import java.util.ArrayList;
import java.util.List;
import java.util.stream.Collectors;
import org.locationtech.jts.geom.Coordinate;
```

rust/mlt-core/src/layer/v01/stream/mod.rs
```diff
@@ -436,15 +436,20 @@ impl StreamMeta {
                     LogicalEncoding::DeltaRle(rle)
                 }
             }
-            (LT::Morton, LT::None) => {
+            (LT::Morton, LT::None) | (LT::Morton, LT::Rle) | (LT::Morton, LT::Delta) => {
```
CommanderStorm: clippy complaint

```suggestion
            (LT::Morton, LT::None | LT::Rle | LT::Delta) => {
```

CommanderStorm:
Merging since this makes maintaining these tests simpler

PR #964, opened 2026-02-25
feat(ts): add FastPFOR encoder/decoder support

DoFabien:
This PR integrates FastPFOR support in the TypeScript implementation, including both decoding and encoding paths.

### Since the stacked reviews

- Rebased cleanly on latest `upstream/main`.
- Updated synthetic TS tests after `#922`:
  - `polygon_fpf`
  - `polygon_hole_fpf`
  - `polygon_multi_fpf`
  - `polygon_fpf_tes`


DoFabien:
Oops, rebased locally and forgot to push 😅 Just force-pushed

ts/src/encoding/fastPforEncoder.ts
```diff
@@ -0,0 +1,493 @@
...
+}
+
+/**
+ * Encodes an int32 stream using the FastPFOR wire format (pages + VByte tail).
+ *
+ * If `workspace` is omitted, a new workspace is created per call.
+ */
+export function encodeFastPforInt32WithWorkspace(
+    values: Int32Array,
+    workspace?: FastPforEncoderWorkspace,
```
HarelM: Make this non-optional for this method.
If the other method (`encodeFastPforInt32`) is not used in production scenario, consider removing it.

ts/src/decoding/bigEndianDecode.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Decodes big-endian bytes into `out` without allocating the output buffer.
+ *
+ * This function does not copy `bytes`; it writes decoded words into the provided `out` array.
+ * For aligned inputs it may create a temporary typed-array view (`Uint32Array`) over `bytes.buffer`
+ * to speed up decoding.
+ *
+ * If `byteLength` is not a multiple of 4, the final word is padded with zeros.
+ *
+ * @returns Number of int32 words written.
+ * @throws RangeError If `(offset, byteLength)` is out of bounds, or if `out` is too small.
```
HarelM, line 13: Are the exceptions thrown here different from the exception thrown in unpackExceptionStreams or other places that throws exceptions?
Please go over all the exceptions thrown as part of this PR and see if some are redundant or not.

HarelM:
I've added two last comments.
I'm wondering why we don't have code coverage for this PR... Let me check that.

HarelM:
@CommanderStorm Rust CI is failing here, can you check please?

CommanderStorm:
Hard to debug since it
- is only happening on this PR and
- only in cargo-llvm-cov instrumented code

Feel free to comment it out in your PR if it is anoying you.
I don't quite have the badwith left to debug this issue right now.

ts/src/encoding/fastPforEncoder.ts
DoFabien: I made workspace non-optional for encodeFastPforInt32WithWorkspace, and I removed encodeFastPforInt32 since it wasn’t used in production scenario. Tests now call `encodeFastPforInt32WithWorkspace` directly

ts/src/decoding/bigEndianDecode.ts
DoFabien: `decodeBigEndianInt32sInto` uses RangeError for caller precondition violations (input/output bounds).
FastPFOR decode paths (decodePage / unpackExceptionStreams) use Error for malformed/truncated encoded payloads.
I also removed one redundant check in decodePage (`packedEnd >= inValues.length`), since it was already implied by the previous whereMeta bounds validation.

CommanderStorm:
@HarelM Likely fixed it

HarelM:
Looks like java is now failing :-(

HarelM:
Ha! I finally got the coverage report back on.
@DoFabien can you check what part of the decoder is missing coverage and see if you can add the relevant tests?

HarelM:
I see that most of the missing coverage is related to exceptions, which if there's no way to reach them then please remove them.
Other non exceptions code needs coverage I believe as it looks like some missing edge cases.

ts/src/decoding/bigEndianDecode.ts
HarelM: Thanks, I would recommend a through look at all the exceptions and see if they can really happen or is it only defensive programming, or otherwise checked in more than one place.
Also see my comment about code coverage in regards to these exceptions.

HarelM:
I've reviewed this in small chucks in @DoFabien fork, so the comments here are minimal as most of the work was done on the fork.
Thanks again @DoFabien for all the hard work on this, and sorry for all my annoying code review comments...
Once this is merged I'll create a new release, which also has the double/float bug fix.

DoFabien:
> I see that most of the missing coverage is related to exceptions, which if there's no way to reach them then please remove them. Other non exceptions code needs coverage I believe as it looks like some missing edge cases.

Thanks, I’m currently working through this and will finish it tomorrow.

I’ve started by separating truly reachable paths from defensive checks. A significant part of the uncovered “exceptions” are not actually reachable with the current wire types (for example values read from Uint8Array are always in 0-255, so checks like >= 256 can’t trigger). 

For the remaining gaps, I’m adding focused tests around real edge cases (exception and non-exception paths), so the coverage increase comes from meaningful behavior rather than test-only scenarios.

HarelM:
Great! Thanks! This is really exciting!

HarelM:
@DoFabien let me know when you need another review. You can ignore the rust CI failures.
@nyurik @CommanderStorm are you aware of the instability of the rust CI?

DoFabien:
Hi @HarelM,
I’ve just pushed an update to the PR, it should be ready for another review. Coverage should now be in a good state.
I won’t have access to a computer until tomorrow, but I’ll address any feedback as soon as I’m back.
Thanks again.

HarelM:
I'm not sure when you pushed, but typescript CI us failing...
No rush, fix it when you can.

DoFabien:
Hi @HarelM,
I missed that, thanks for pointing it out
I just pushed a follow-up commit to fix that, and lint now passe.

Ready for another review.
Thanks again for all the feedback.

ts/src/encoding/fastPforEncoder.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+        const workspace = createFastPforEncoderWorkspace();
+        workspace.dataToBePacked[2] = new Int32Array(UNDERSIZED_PREALLOCATED_STREAM);
+
+        const encoded = encodeFastPforInt32WithWorkspace(values, workspace);
+        const decoded = decodeFastPforInt32(encoded, values.length);
+
+        expect(decoded).toEqual(values);
+        const exceptionStream = workspace.dataToBePacked[2];
+        expect(exceptionStream).toBeDefined();
+        if (!exceptionStream) throw new Error("expected exception stream for bitWidth=2");
```
HarelM: You should not throw in a test, use the vitest infra for that or remove completely

ts/src/encoding/fastPforEncoder.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+        values[EXCEPTION_POS_B] = EXCEPTION_OUTLIER_VALUE;
+
+        const workspace = createFastPforEncoderWorkspace();
+        workspace.dataToBePacked[2] = new Int32Array(UNDERSIZED_PREALLOCATED_STREAM);
+
+        encodeFastPforInt32WithWorkspace(values, workspace);
+
+        const exceptionStream = workspace.dataToBePacked[2];
+        expect(exceptionStream).toBeDefined();
+        if (!exceptionStream) throw new Error("expected exception stream for bitWidth=2");
```
HarelM: Same here

ts/src/decoding/fastPforDecoder.spec.ts
```diff
@@ -0,0 +1,432 @@
...
+        const encoded = encodeFastPforInt32WithWorkspace(values, createFastPforEncoderWorkspace());
+        const workspace = createDecoderWorkspace();
+        workspace.byteContainer = new Uint8Array(4096);
+        workspace.byteContainerI32 = new Int32Array(4);
+
+        const decoded = decodeFastPforInt32(encoded, values.length, workspace);
+        expect(decoded).toEqual(values);
+        const byteContainerI32 = workspace.byteContainerI32;
+        expect(byteContainerI32).toBeDefined();
+        if (!byteContainerI32) throw new Error("expected byteContainerI32 to be rebuilt");
```
HarelM: Same here

HarelM:
I see that coverage was improved, thanks! Can you check if the encoder coverage is sufficient or some edge cases are not interesting to kerp or test?

HarelM:
@CommanderStorm @nyurik Rust CI is still failing? Should we disable it until you can make it stable?

DoFabien:
Thanks. I addressed both points: 

- I added a large round-trip test to cover the remaining encoder lines
- and I removed custom throw statements from tests

PR #965, opened 2026-02-25
fix(ts): double column decoded as float

Turtelll:
Like we dicussed in #940 we reverted to the cleaner approach from PR #786. 
- I removed the unused low level decod/encodeDouble functions.
- encodeDoubleColumn and encodeNullableDoubleColumn are still featured to round trip test the propertyDecoder.


HarelM (approved):
LGTM, thanks!

HarelM:
@nyurik I don't see a codecov comment regarding covrage, can you check if this is related to recent changes you made in the CI?

PR #969, opened 2026-02-25
Convert CLI apps to Kotlin

TimSylvester:
Mostly automated conversion via Android Studio (not LLM-based, I think), with some manual cleanup.

Also changes the order of elements in the JSON output to insertion order for everything except properties, for better readability, and updates the fixture files to match.  Equality was checked by sorting with `jq`.

### DIFF
--- a/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/ConversionHelper.kt
+++ b/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/ConversionHelper.kt
@@ -216,7 +216,7 @@ fun convertTile(
                     outputStream,
                     config.compressionType,
                 ).use { compressStream ->
-                    compressStream!!.write(tileData)
+                    compressStream.write(tileData)
                 }
                 // Evaluate whether the compressed version is worth using.
                 if ((
--- a/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/OfflineDBHelper.kt
+++ b/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/OfflineDBHelper.kt
@@ -177,7 +177,7 @@ private fun updateMetadata(
                         config.compressionType,
                     ).use { compressStream ->
                         compressed = (compressStream !== outputStream)
-                        compressStream!!.write(jsonString.toByteArray(StandardCharsets.UTF_8))
+                        compressStream.write(jsonString.toByteArray(StandardCharsets.UTF_8))
                     }
                     data = outputStream.toByteArray()
                 }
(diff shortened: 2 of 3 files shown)


louwers:
@TimSylvester Do you prefer using Kotlin for the CLI?

TimSylvester:
I'd prefer to use Kotlin for everything, but it has a runtime dependency, so the CLI is all I can really get away with, I think.

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/JsonHelper.kt
```diff
@@ -0,0 +1,193 @@
...
+import org.maplibre.mlt.data.Layer
+import org.maplibre.mlt.data.MapLibreTile
+import java.util.SortedMap
+import kotlin.Any
+import kotlin.Boolean
+import kotlin.String
+import kotlin.math.floor
+import kotlin.streams.asSequence
+
+object JsonHelper {
```
louwers: I think Kotlin supports free standing functions.

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/EncodeConfig.kt
```diff
@@ -0,0 +1,186 @@
...
+            .willPrintMLT(this.willPrintMLT)
+            .willPrintMVT(this.willPrintMVT)
+            .compareProp(this.compareProp)
+            .compareGeom(this.compareGeom)
+            .willTime(this.willTime)
+            .dumpStreams(this.dumpStreams)
+            .taskRunner(this.taskRunner)
+            .continueOnError(this.continueOnError)
+
+    class Builder {
```
louwers: I think this can be a `data class` as well.

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/JsonHelper.kt
TimSylvester: There's more stuff in there than the one method, and the three "container" helpers inherit common stuff from the base class.

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/EncodeConfig.kt
TimSylvester: Yes, that's just left over from the conversion, I'm removing it, along with unnecessary annotations, etc.

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/JsonHelper.kt
TimSylvester: I could reference those instead of inheriting, I'll try that...

TimSylvester: Yeah, that's probably better

PR #982, opened 2026-02-26
test: More testing around geometry combinations

CommanderStorm:
Okay, I found another error.

@nyurik
Do you have an idea how this can be fixed?
The code where this happens is very alien to me and I don't understand geometrys enough to be able to pin-point how we should handle this.


nyurik:
not really - but as always, lets start with generating more synthetics, probably in Java encoder as it is more tested and understood, although at some point once we are certain, we could start moving towards rust encoder

nyurik (approved):

Issue #978, opened 2026-02-26
Support NaN and ±Infinity for Rust ⇔GeoJSON synthetics


nyurik:
In order to have comprehensive testing, we must ensure every possible value we support is readable in every implementation.  GeoJSON does not support non-numeric floats, which means we should serialize it as strings to make sure we are consistent in every language.  I used Rust names for the values, I hope we don't start a holy naming war here :).  f32 and f64 are needed to be explicit about the originating data type.

Make sure to create pull requests against the `inf-float-as-str` integration branch.

## Requirements

* Serialize `f32::NAN` → `"f32::NAN"`
* Serialize `f32::INFINITY` → `"f32::INFINITY"`
* Serialize `f32::NEG_INFINITY` → `"f32::NEG_INFINITY"`
* Serialize `f64::NAN` → `"f64::NAN"`
* Serialize `f64::INFINITY` → `"f64::INFINITY"`
* Serialize `f64::NEG_INFINITY` → `"f64::NEG_INFINITY"`

* Deserialize `"f32::NAN"` → `f32::NAN`
* Deserialize `"f32::INFINITY"` → `f32::INFINITY`
* Deserialize `"f32::NEG_INFINITY"` → `f32::NEG_INFINITY`
* Deserialize `"f64::NAN"` → `f64::NAN`
* Deserialize `"f64::INFINITY"` → `f64::INFINITY`
* Deserialize `"f64::NEG_INFINITY"` → `f64::NEG_INFINITY`

@CommanderStorm

CommanderStorm:
makes sense.

Honestly, I would prefer if we would serialize our floats as strings wholesale.
The amount of logic that I have written due to f64 <-> f32 is not very nice

HarelM:
Do we really need an issue for each language?

nyurik:
@HarelM easier to track

nyurik:
@CommanderStorm hmm, that's actually a very interesting idea for synthetics...  That said, at the moment, json serializer is somewhat generically useful - any tile can be converted to a proper GeoJson that can be used in qgis and elsewhere. If we use `"prop": "3.14_f32"` it is good for validation, but useless as geojson for most usecases...

CommanderStorm:
> but useless as geojson for most usecases

Is it?

Would you not look at the content in QGis anyhow?
So strings have the same content as floats, so fine?

PR #983, opened 2026-02-26
Support NaN and ±Infinity in f32/f64 JSON conversion

caglareker:
## Summary

Add proper handling for special floating point values (NaN and ±Infinity) when converting between Rust and GeoJSON formats.

## Changes

- Update `f32_to_json()` to return string representations for NaN and infinite values ("f32::NAN", "f32::INFINITY", "f32::NEG_INFINITY")
- Introduce `f64_to_json()` function with equivalent special value handling
- Update `PropValue::to_geojson()` to use dedicated f64 converter instead of casting through f32
- Update test synthetic files to reflect correct handling of special values

## Rationale

Special floating point values require explicit handling during serialization since JSON itself doesn't natively support NaN or infinity. Using string tags ensures these values roundtrip correctly and are unambiguous in the wire format.

Fixtures #978


HarelM:
Thanks for opening this PR, there's an integration branch to support this throughout different languages, I think it will make sense to target that branch. 
Once all the implantations will support this and CI passes we could merge to main.

caglareker:
Got it, I'll retarget this to the integration branch. Makes sense to coordinate across all the language implementations before merging to main.

nyurik:
@caglareker i will close this PR - please re-open it against the `inf-float-as-str` branch - this way we can consolidate all language changes at the same time. Thank you!

nyurik:
@caglareker i'm so sorry, i asked you to do this, but forgot to do it myself, and now both java and rust code is in main for this :'(

nyurik:
~reopening...~  apparently I can't reopen it now

caglareker:
no worries, i checked #1000 and it already covers everthing. nice cleanup with the json5 removal btw 👍

PR #985, opened 2026-02-26
ci(rust): add more debugging information to narrow down the target-cpu flakyness

CommanderStorm:


rust/mod.just
```diff
@@ -213,7 +213,7 @@ sync-pyo3-version:
 alias t := test
 # Run all tests
 test:
-    cargo test --all-targets --all-features{{if CI}} -- --test-threads=1{{endif}}
+    cargo test --all-targets --all-features{{if ci_mode == '1' {' -- --test-threads=1'} {{endif}}
```
nyurik: why endif?

CommanderStorm: stupidity and not testing my change locally

rust/mod.just
```diff
@@ -35,24 +35,36 @@ check: (just 'cargo-install' 'cargo-hack')
...
     mkdir -p target/llvm-cov
     # ATTENTION: the output-path here is used in the CI workflow
     {{just}} rust::coverage '--codecov --output-path target/llvm-cov/codecov.info'
 
 # Run minimal subset of tests to ensure compatibility with MSRV
-ci-check: env-info check
+ci-check: ci-env-info check
+
+# Print environment info for CI
+ci-env-info:
```
nyurik, line 47: envinfo is mostly used for CI anyway, no need to have two different ones here, just add this to envinfo

CommanderStorm: I don't want this polluting my `just rust::test` calls

nyurik (approved):
@CommanderStorm please consolidate envinfo changes into the existing envinfo recipe, but approving otherwise

rust/mod.just
nyurik: you may want to remove it from rust::test, and only keep it in ci-test

PR #987, opened 2026-02-27
test(synthetic): Add more geometric combination tests

CommanderStorm:
This PR adds more geometric combination tests.

The rust version seems to have some resonably large bugs, so some portion of them is commented out.


java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -66,18 +67,40 @@ private static void generateMultiLineStrings() throws IOException {
   }
 
   private static void generateMixed() throws IOException {
-    write(layer("mixed_pt_line", feat(p0), feat(line(c1, c2))), cfg());
-    write(layer("mixed_pt_poly", feat(p0), feat(poly(c1, c2, c3, c1))), cfg());
-    write(layer("mixed_line_poly", feat(line(c1, c2)), feat(poly(c1, c2, c3, c1))), cfg());
-    write(layer("mixed_pt_mline", feat(p0), feat(multi(line(c1, c2), line(h1, h2, h3)))), cfg());
+    record GeomType(String name, Geometry geom) {}
+    // we deduplicate duplicate geometry, so yes this is nessesary
```
```suggestion
```

CommanderStorm:
> I would disable some ts and rust tests if they fail on this, just do that e can have all these tests in main

Ok 👍🏻

PR #988, opened 2026-02-27
test(synthetic): more floating point tests

CommanderStorm:
Floats are a bit weird, these are a few of the values I learned exists -> we likely should cover them too


nyurik (approved):

nyurik:
@CommanderStorm Java CI broke here :(

PR #993, opened 2026-02-27
chore(java): rework all Java synthetics

nyurik:
Another pass at a more comprehensive test suit:
* add all combinations of different geometries, including a polygon with a hole
* introduce better coordinates - two non-overlapping triangles, and a hole in one of them
* use coordinates that are 0..90 to pretend we are in lat/lng space for simpler QGis debugging
* reduce default extent to 80 to make the tile coordinates better fit the extent

image


java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -63,44 +66,53 @@ private static void generateMultiPoints() throws IOException {
...
+
+  private static void generateMixedCombine(GeomType[] arr, int k, int start, List<GeomType> current)
+      throws IOException {
+    if (current.size() == k) {
+      var name = "mixed_" + current.stream().map(GeomType::sym).collect(Collectors.joining("_"));
+      var feats = current.stream().map(t -> t.feat).toArray(Feature[]::new);
+      write(layer(name, feats), cfg());
+    } else {
+      for (int i = start; i < arr.length; i++) {
+        if (i > start && arr[i] == arr[i - 1]) {
```
CommanderStorm, line 83: I think because quality is by default a pointer comparison we want this, not sure.

```suggestion
        if (i > start && arr[i].sym().equals(arr[i - 1].sym())) {
```

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -63,44 +66,53 @@ private static void generateMultiPoints() throws IOException {
...
-        // due to how offsets work, this is still a good test to have
-        if (!t2.name().equals(t1.name())) {
-          var l2 = layer(name + "_" + t1.name(), feat(t1.geom()), feat(t2.geom()), feat(t1.geom()));
-          write(l2, cfg());
+    for (var a : types) {
+      for (var b : types) {
+        if (!a.sym.equals(b.sym)) {
+          // Create A-B-A variants
+          var sym = "mixed_" + a.sym + "_" + b.sym + "_" + a.sym;
+          write(layer(sym, a.feat, b.feat, a.feat), cfg());
```
CommanderStorm: I think it might still be valuable to have an A-A variant, just to be sure.
This is a bit unlikely to actually cause any failousrs, but not sure.

```suggestion
    for (var a : types) {
      // Create A-A variants
      var sym_aa = "mixed_" + a.sym + "_" + a.sym;
      write(layer(sym_aa, a.feat, b.feat, a.feat), cfg());
      for (var b : types) {
        if (!a.sym.equals(b.sym)) {
          // Create A-B-A variants
          var sym_aba = "mixed_" + a.sym + "_" + b.sym + "_" + a.sym;
          write(layer(sym_aba, a.feat, b.feat, a.feat), cfg());
```

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```suggestion
    int[] extents = {512, 4096, 131072, 1073741824};
    for (int e : extents) {
        write(layer("extent_" + e, e, feat(line(c(0, 0), c(e - 1, e - 1)))), cfg());
        write(layer("extent_buf_" + e, e, feat(line(c(-42, -42), c(e + 42, e + 42)))), cfg());
    }
```

CommanderStorm:
Also please make sure to ignore the failing tests on the TS side

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
nyurik: meh, readability wins, i don't care about cpu cycles here

PR #995, opened 2026-02-27
test(ts): Optimize mltDecoder MVT comparison assertions

DoFabien:
## What
Replaced hot-path `expect(...).toEqual(...)` with `assert.deepEqual(...)` / `assert.equal(...)` and removed one redundant assertion (`Object.keys(...).length`) already covered by deep equality.

## Perf
We go from 63s to 9s without coverage, and from 101s to 20s with coverage (on my old local machine).

## Trade-offs

- Error message format changes slightly (`expect` vs `assert`) but still shows clear expected/actual diff.
- Example:
  - `expect`: `expected [ [ Point{ x: 2, y: 2 }, … ] ] to deeply equal []`
  - `assert`: `AssertionError: Expected values to be strictly deep-equal: + actual - expected`

### DIFF
--- a/ts/src/mltDecoder.spec.ts
+++ b/ts/src/mltDecoder.spec.ts
@@ -1,5 +1,5 @@
 import assert from "node:assert/strict";
-import { expect, describe, it } from "vitest";
+import { describe, it } from "vitest";
 import { readdirSync, readFileSync } from "fs";
 import { parse, join } from "path";
 import { VectorTile, type VectorTileFeature } from "@mapbox/vector-tile";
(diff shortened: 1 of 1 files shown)


HarelM:
Are you sure about the performance gain? I've asked Gemeni and it said that they should be have similar performance overall. 
How did you thought about doing this change? Did you see it in another project?

DoFabien:
Yes, I’m sure about these numbers.

I used this approach because I hit the same issue a few months ago on my main project: in very hot test loops, a lot of time was spent in the test framework matcher layer.

Node’s built-in `assert` is lighter in this context. It avoids part of the Vitest matcher overhead (`expect`/`toEqual` pipeline, matcher state, rich formatting path), which is why this specific test gets significantly faster.

You can validate it directly in CI by comparing this PR run with the latest `main` run for the same test file and coverage settings.  
Example CI run: https://github.com/maplibre/maplibre-tile-spec/actions/runs/22483642996/job/65127758558?pr=995#step:6:52

HarelM:
Yup, I see the difference from main ~100s, and on this PR ~20s.
Although there might be different machine allocation, the fact that you saw it on your project is what I needed to confirm this change, THANKS!

HarelM:
Can we completely remove the import of `expect` then?
What about the synthetic tests? Can they benefit from this as well?

DoFabien:
I have remove `expect`

For this test scope, we are comparing around 1M features in total, so this optimization targets a real hot path.
For synthetic fixtures, runtime is already very small,  so the potential gain is much smaller.
Also, switching synthetics to Node `assert` would reduce the richness of Vitest failure messages there.

HarelM:
Fair enough.

PR #996, opened 2026-02-27
ci(rust): automatically rerun CI if it flakes

CommanderStorm:
A bit of a nuklear option, but likely still helpfull to unblock us on this

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -127,19 +127,16 @@ jobs:
     steps:
       - name: Result of the needed steps
         run: echo "${{ toJSON(needs) }}"
-      - if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
-        run: exit 1
-  # FIXME: Some of our tests require AVX, which is not avaliable on AMD. THIS IS A BUG!
-  rerun-on-failure:
-    needs: [ test, test-msrv, lint, build-release-binaries, test-publish ]
-    if: failure() && fromJSON(github.run_attempt) < 3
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -127,7 +127,15 @@ jobs:
...
       - name: Result of the needed steps
         run: echo "${{ toJSON(needs) }}"
-      - if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
+      # FIXME: Some of our tests require AVX, which is not avaliable on AMD. THIS IS A BUG!
+      - run: gh run rerun ${{ github.run_id }} --failed
+        if: (contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')) && fromJSON(github.run_attempt) < 3
+        name: Rerun on failour
+        env:
+          GH_REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ secrets.RERUN_RUST_CI_PAT }}
```
CommanderStorm, line 136: I think this needs to be in main for this secret to be avaliable.

It is registered in our settings.

nyurik: I don't think we should enable this. The rest is not wrong, not flaky. We simply retry until we get an Intel runner. This doesn't seem like a good stategy

CommanderStorm: I don't know how to fix this issue otherwise.

Neither the FPF, nor the CXX bindings have obvious soundness holes and all of my machines are too new to have AVX..

PR #998, opened 2026-02-27
chore(java): export non-finite floats as strings

### LINKED ISSUE #977: Support NaN and ±Infinity for Java ⇔GeoJSON synthetics
In order to have comprehensive testing, we must ensure every possible value we support is readable in every implementation.  GeoJSON does not support non-numeric floats, which means we should serialize it as strings to make sure we are consistent in every language.  I used Rust names for the values, I hope we don't start a holy naming war here :).  f32 and f64 are needed to be explicit about the originating data type.

## Requirements
* Serialize `f32::NAN` → `"f32::NAN"`
* Serialize `f32::INFINITY` → `"f32::INFINITY"`
* Serialize `f32::NEG_INFINITY` → `"f32::NEG_INFINITY"`
* Serialize `f64::NAN` → `"f64::NAN"`
* Serialize `f64::INFINITY` → `"f64::INFINITY"`
* Serialize `f64::NEG_INFINITY` → `"f64::NEG_INFINITY"`

The reverse is not needed because we don't want to support magical geojson -> mlt encoding, only in synthetics

@TimSylvester

nyurik:
fixes #977

### DIFF
--- a/rust/mlt-core/Cargo.toml
+++ b/rust/mlt-core/Cargo.toml
@@ -15,6 +15,10 @@ rust-version.workspace = true
 name = "decoding"
 harness = false
 
+[[bench]]
+name = "encoding"
+harness = false
+
 [features]
 arbitrary = ["dep:arbitrary", "geo-types/arbitrary"]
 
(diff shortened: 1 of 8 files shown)


nyurik:
that's because our java code apparently not handling f64 values correctly.  I disabled those tests

PR #1003, opened 2026-02-28
feat(java): add geometry enc option to GeometryEncoder

nyurik:
This allows us to generate synthetics with controlled geometry stream encoding parameters, so that we can test encoding of geometries themselves, rather than streams used underneath.


CommanderStorm (approved):
But have a look at copiot 😉

PR #1005, opened 2026-02-28
feat(rust): add rust morton encoding

CommanderStorm:
This PR is for working towards enabling rust based morton encoding in the synthetics

Disclaimer: I got some LLM help what to change where (and how to test it)

### DIFF
--- a/rust/mlt-synthetics/src/main.rs
+++ b/rust/mlt-synthetics/src/main.rs
@@ -27,10 +27,6 @@ const C0: Coord<i32> = coord! { x: 13, y: 42 };
 const C1: Coord<i32> = coord! { x: 11, y: 52 };
 const C2: Coord<i32> = coord! { x: 71, y: 72 };
 const C3: Coord<i32> = coord! { x: 61, y: 22 };
-// triangle 2, clockwise winding, X ends in 3, Y ends in 4
-const C21: Coord<i32> = coord! { x: 23, y: 34 };
-const C22: Coord<i32> = coord! { x: 73, y: 4 };
-const C23: Coord<i32> = coord! { x: 13, y: 24 };
 // hole in triangle 1 with counter-clockwise winding
 const H1: Coord<i32> = coord! { x: 65, y: 66 };
 const H2: Coord<i32> = coord! { x: 35, y: 56 };
@@ -72,11 +68,12 @@ static MIX_TYPES: LazyLock<[(&'static str, Geom32); 7]> = LazyLock::new(|| {
 
 fn main() {
     let dir = Path::new("../test/synthetic/0x01-rust/");
-    fs::create_dir_all(dir).unwrap_or_else(|_| panic!("to be able to create {}", dir.display()));
+    fs::create_dir_all(dir)
+        .unwrap_or_else(|e| panic!("to be able to create {}: {e:?}", dir.display()));
 
     let dir = dir
         .canonicalize()
-        .unwrap_or_else(|_| panic!("bad path {}", dir.display()));
+        .unwrap_or_else(|e| panic!("bad path {}: {e:?}", dir.display()));
     println!("Generating synthetic test data in {}", dir.display());
 
     let writer = SynthWriter::new(dir);


rust/mlt-core/src/layer/v01/geometry/mod.rs
```diff
@@ -832,18 +831,23 @@ impl<'a> Geometry<'a> {
 
 impl Debug for DecodedGeometry {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let DecodedGeometry {
```
CommanderStorm, line 834: so that if we add a field, this is a compiler error instead of us ignoring the field

rust/mlt-core/src/layer/v01/geometry/mod.rs
```diff
@@ -960,13 +964,31 @@ impl<'a> FromEncoded<'a> for DecodedGeometry {
...
+                offsets
+                    .iter()
+                    .flat_map(|&i| {
+                        let i = i as usize;
+                        [dict[i * 2], dict[i * 2 + 1]]
+                    })
+                    .collect(),
+            );
+        }
+        // vertex_offsets is left as None — fully decoded
```
CommanderStorm:
```suggestion
```

rust/mlt-core/src/layer/v01/geometry/mod.rs
```diff
@@ -162,7 +162,6 @@ pub struct DecodedGeometry {
     pub geometry_offsets: Option<Vec<u32>>,
     pub part_offsets: Option<Vec<u32>>,
     pub ring_offsets: Option<Vec<u32>>,
```
CommanderStorm, line 164: not sure of removing `vertex_offsets` is the right call..

rust/mlt-synthetics/src/main.rs
```diff
@@ -113,25 +113,44 @@ fn main() {
...
 
 fn generate_geometry(w: &SynthWriter) {
     w.geo_varint().geo(P0).write("point");
     w.geo_varint().geo(line1()).write("line");
+    // Morton (Z-order) line: de-interleave index bits into x/y (even/odd bits).
+    let num_points = 16; // 4x4 complete Morton block
+    let scale = 8;
+    let morton_bits = 4;
+    let mut morton_curve = Vec::with_capacity(num_points);
+    for i in 0_i32..16 {
```
michaelkirk:
```suggestion
    for i in 0_i32..num_points {
```

I think this is referring to the same thing, right? Not just a coincidence that they are both 16?

rust/mlt-core/src/layer/v01/geometry/mod.rs
```diff
@@ -1239,4 +1263,84 @@ mod tests {
...
+        let (remaining, parsed) = EncodedGeometry::parse(&buffer).unwrap();
+        assert!(remaining.is_empty());
+        let decoded = DecodedGeometry::from_encoded(parsed).unwrap();
+
+        assert_eq!(decoded.vertices, Some(vec![0i32, 0, 4, 0, 0, 4, 4, 0]));
+
+        let geom = decoded.to_geojson(0).unwrap();
+        assert_eq!(
+            geom,
+            GeoGeom::LineString(LineString(vec![
```
michaelkirk: No big deal, but in case you weren't aware, my favorite way to instantiate geo_types is with the `wkt!` macro.

So this could be: `wkt!(LINESTRING(0 0,4 0,0 4,4 0))`

Note: it is **compile** time, so there's no issue with running into run-time wkt parsing errors with this macro. 

The main limitation is you can't interpolate variables:

```rust
let point = coord(x: 4, y: 0);
// WONT COMPILE
wkt!(LINESTRING(0 0,coord,0 4,coord))
```

Up to you though!

michaelkirk (approved):

rust/mlt-core/src/layer/v01/geometry/mod.rs
CommanderStorm: that is actually very useful to know, we should use that more.

rust/mlt-synthetics/src/main.rs
```diff
@@ -49,48 +49,23 @@ const fn c(x: i32, y: i32) -> Coord<i32> {
...
-
 static MIX_TYPES: LazyLock<[(&'static str, Geom32); 7]> = LazyLock::new(|| {
     [
-        ("pt", p(38, 29).into()),
-        ("line", line![c(5, 38), c(12, 45), c(9, 70)].into()),
-        (
-            "poly",
-            poly![c(55, 5), c(58, 28), c(75, 22), c(55, 5)].into(),
-        ),
+        ("pt", wkt!(POINT(38 29)).into()),
```
michaelkirk, line 54: I confess I'm 50/50 on using `wkt!` for `Point`/`Coord` because there are more concise alternatives, but I suppose there is some value in consistency. I don't mind either way.

PR #1008, opened 2026-03-01
fix(ts-decoder) ID column dispatch by metadata type

DoFabien:
While benchmarking the TS decoder, I wanted to include `bing` fixtures as well (they use FastPFOR).
During that run, I hit a decoding failure.

## What failed

On some Bing `tag0x01` tiles, TS decoding failed with cascading errors such as:

- unsupported technique/stream combinations
- FastPFOR errors like `invalid byte length`

Those errors were not the root cause. They were side effects of stream offset desynchronization.

## Root cause

In `decodeTile`, column routing was name based:

- `"id"` was treated as the technical ID column
- `"geometry"` was treated as geometry

Some Bing schemas contain a regular feature property named `id` (for example a string field).
That property was incorrectly decoded through the ID path, which shifted stream reads and then broke downstream decoding.

## Fix

This PR switches routing from name based to metadata/type based:

- keep ID type codes (`0..3`) as logical ID metadata in TS
- detect ID columns via `LogicalScalarType.ID`
- detect geometry via `ComplexType.GEOMETRY`
- resolve ID physical width (`UINT_32` or `UINT_64`) from logical ID metadata

## Commit split

1. commit 1

- adds only the metadata unit test
- this commit is intentionally red on its own (the test fails without the fix)

2. commit 2

- adds the decoder fix
- makes the test from commit 1 pass

For now, this PR does not add the Bing fixture integration test because this branch does not include FastPFOR integration yet.

Once FastPFOR is integrated here, we can add the Bing integration coverage to complete the regression surface.

### DIFF
--- a/ts/src/metadata/tileset/typeMap.spec.ts
+++ b/ts/src/metadata/tileset/typeMap.spec.ts
@@ -27,7 +27,7 @@ describe("typeMap helpers", () => {
         }
     });
 
-    it("should detect logical ID columns by metadata type, not by name", () => {
+    it("should return false for physical scalar columns even when column is named id", () => {
         const physicalIdNamedColumn = {
             name: "id",
             type: "scalarType",
@@ -37,6 +37,10 @@ describe("typeMap helpers", () => {
             },
         } as Column;
 
+        expect(isLogicalIdColumn(physicalIdNamedColumn)).toBe(false);
+    });
+
+    it("should return true for logical ID columns regardless of column name", () => {
         const logicalIdAnyNameColumn = {
             name: "my_custom_id",
             type: "scalarType",
@@ -47,7 +51,6 @@ describe("typeMap helpers", () => {
             },
         } as Column;
 
-        expect(isLogicalIdColumn(physicalIdNamedColumn)).toBe(false);
         expect(isLogicalIdColumn(logicalIdAnyNameColumn)).toBe(true);
     });
 


ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.spec.ts
```diff
@@ -214,5 +214,32 @@ describe("embeddedTilesetMetadataDecoder", () => {
             expect(metadata.featureTables[0].name).toBe("");
             expect(metadata.featureTables[0].columns[0].complexType.children).toHaveLength(1);
         });
+
+        const logicalIdTypeCodeCases = [
```
HarelM: This looks very similar to the typemap tests, I would recommend testing only the relevant code in each unit test file and avoid duplication as much as possible.

ts/src/metadata/tileset/typeMap.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+            type: "complexType",
+            complexType: {
+                type: "physicalType",
+                physicalType: ComplexType.GEOMETRY,
+                children: [],
+            },
+        } as Column;
+
+        expect(isGeometryColumn(structNamedGeometryColumn)).toBe(false);
+        expect(isGeometryColumn(geometryAnyNameColumn)).toBe(true);
```
HarelM, line 82: Split to two diffenetnt tests please.

ts/src/metadata/tileset/typeMap.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+            { typeCode: 1, longID: false },
+            { typeCode: 2, longID: true },
+            { typeCode: 3, longID: true },
+        ];
+
+        for (const { typeCode, longID } of cases) {
+            const column = decodeColumnType(typeCode);
+
+            expect(column).not.toBeNull();
+            if (!column) {
```
HarelM: Avoid throwing in tests.

HarelM:
Thanks for taking the time to open this PR.
I've added some comments about the tests, looks good otherwise.

DoFabien:
Thanks, I addressed your test comments:
- deduplicated overlapping test coverage
- split geometry checks into separate tests
- removed manual `throw` in tests

Would you be okay if we also enable the Bing tile tests in `ts/src/mltDecoder.spec.ts` in this PR?

ts/src/metadata/tileset/typeMap.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+            type: "scalarType",
+            scalarType: {
+                type: "logicalType",
+                logicalType: LogicalScalarType.ID,
+                longID: true,
+            },
+        } as Column;
+
+        expect(isLogicalIdColumn(physicalIdNamedColumn)).toBe(false);
+        expect(isLogicalIdColumn(logicalIdAnyNameColumn)).toBe(true);
```
HarelM, line 54: This should be split as well as I don't see any relation between the two checks and their input.

HarelM:
If it's not a lot of code to add this test, then sure, feel free to add the bing tile.

DoFabien:
I split this into two separate tests as requested.
I also enabled the Bing tile tests in `ts/src/mltDecoder.spec.ts` (17 tiles).

PR #1018, opened 2026-03-02
fix(ts):  MultiPoint offset counters and MultiPolygon offset

Turtelll:
- MultiPoint: After decoding a MultiPoint geometry with N points, the `partOffsetCounter` and `ringOffsetsCounter` were not incremented 
- MultiPolygon: The old `getGeometry()` function returned a flat coordinate array for Polygons. The helper function reconstructs the hierarchical polygon structure using the TopologyVectors's offsets.


ts/src/synthetic.spec.ts
```diff
@@ -245,8 +125,24 @@ function getGeometry(geometry: Geometry): GeoJSON.Geometry {
...
             return { type: "MultiLineString", coordinates: coords };
         case GEOMETRY_TYPE.MULTIPOLYGON:
-            return { type: "MultiPolygon", coordinates: coords.map((r) => [r]) };
+            return getMultiPolygon(geometry, table, featureIndex);
         default:
             throw new Error(`Unsupported geometry type: ${geometry.type}`);
     }
 }
+
+function getMultiPolygon(geometry: Geometry, table: FeatureTable, featureIndex: number): GeoJSON.MultiPolygon {
```
HarelM: Why is this part here? Seems like it should not be part of the test, what am I missing?

ts/src/synthetic.spec.ts
```diff
@@ -207,18 +85,21 @@ describe("MLT Decoder - Synthetic tests", () => {
 function featureTablesToFeatureCollection(featureTables: FeatureTable[]): GeoJSON.FeatureCollection {
     const features: GeoJSON.Feature[] = [];
     for (const table of featureTables) {
-        for (const feature of table.getFeatures()) {
+        const tableFeatures = table.getFeatures();
```
HarelM: I would expect this code to remain the same...

ts/src/synthetic.spec.ts
Turtelll: The coordinates array stores a flattened array of coordinates with no information. 

- Input:  [ [[ring1], [ring2]], [[ring3]] ]  // 2 polygons: first has 2 rings, second has 1 ring
- Output: [ [ring1], [ring2], [ring3] ]      // Just 3 rings, no idea where polygons split

Thats why we have:
- geometryOffsets: Maps feature index
- partOffsets: Maps polygon index
- ringOffsets: Maps ring index

Turtelll: I think moving this to geometryVectorConverter.ts would be a good idea. Let me know what you think.

Turtelll:
@HarelM I looked deeper into this and found out that this is a constraint from MVT. There features are loaded into VectorTileFeatureLike which only include coordinate arrays with 2D Array
```
export interface VectorTileFeatureLike {
    type: 0 | 1 | 2 | 3;
    properties: Record<string, number | string | boolean>;
    id: number | undefined;
    extent: number;
    loadGeometry(): Point[][];
}
```
But for MultiPolygon this is not sufficient. That is why their coordinate array is flattened.  Maplibre uses [classifyRings ](https://github.com/maplibre/maplibre-style-spec/blob/0d6c1f8f0bdc889357901da7b9f6890133a12188/src/util/classify_rings.ts#L4)in the [fill_bucket ](https://github.com/maplibre/maplibre-gl-js/blob/main/src/data/bucket/fill_bucket.ts)which classifies an array of rings into polygons with outer rings and holes.

The construction using the topologyVector also works but was unnecessary. This is much cleaner.

HarelM:
@Turtelll check out the following TODO: Seems relevant to the change you made with classifyrings, isn't it:
https://github.com/maplibre/maplibre-tile-spec/blob/b5e100e3deb89bf8ba1203a85dc16a16aa59a611/ts/src/vector/geometry/geometryVectorConverter.ts#L29

Turtelll:
Yea i don't really know what that TODO is supposed to mean

PR #1021, opened 2026-03-02
refactor(java): Optimize PMTiles reading, Kotlin cleanup

TimSylvester:
- Refactor PMTiles access to avoid secondary lookup of data offset per tile (~10x speedup)
- Use lazy evaluation of headers to start converting before loading the entire tile directory
- Fix an off-by-one bug in PMTiles coordinate iteration
- Track unique tile hashes to detect duplicate tile coordinates and repeated tile contents outside a run-length range (when debug output is enabled)
- Remove "helper" from filenames, left over from Java classes
- Eliminate `Environment` class in favor of free functions
- Use lazy evaluations for environment config
- Warn on failures to parse config values
- Make debug/trace messages grey
- Eliminate ANSI output when redirecting
- Add log markers to allow for filtering
- Eliminate use of `atTrace` which doesn't work with markers (slf4j bug)
- Tweak cache parameters (long timeouts lead to big lags when they expire together)
- Tweak log levels, only high-volume messages are trace


java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/Environment.kt
```diff
@@ -1,37 +1,72 @@
...
+const val DEFAULT_THREAD_QUEUE_SIZE = 10_000
+val threadQueueSize by lazy {
+    parse(ENV_THREAD_QUEUE_SIZE, DEFAULT_THREAD_QUEUE_SIZE, String::toInt).also {
+        if (it < 1) {
+            throw IllegalArgumentException("Thread queue size must be positive")
+        }
     }
 }
+
+private fun <T> parse(
```
louwers: `parse` is quite terse of a name.

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/ReadablePmtiles.kt
```diff
@@ -21,35 +21,30 @@ import java.io.UncheckedIOException
...
+        length: Long,
+    ) = getBytes(offset, length).let {
+        if (header.internalCompression() == Pmtiles.Compression.GZIP) {
+            Gzip.gunzip(it)
+        } else {
+            it
         }
     }
 
+    @Throws(IOException::class)
```
louwers: Are these annotations needed?

louwers:
@TimSylvester Looks like coverage is not calculated for the CLI right.

TimSylvester:
Coverage is only from the new tests, not from CLI apps being run in CI.

PR #1026, opened 2026-03-03
chore(rust): rework string parsing

nyurik:


rust/mlt-core/src/layer/v01/stream/mod.rs
```diff
@@ -347,12 +350,99 @@ impl OwnedStream {
...
-        Ok(vec![
-            symbol_length_stream,
-            symbol_table_stream,
-            value_length_stream,
-            compressed_stream,
-        ])
+        // Stream 5: Offset indices (0, 1, 2, ...) for string lookup
+        let offsets: Vec<u32> = (0..values.len())
+            .map(|i| u32::try_from(i).unwrap())
+            .collect();
```
CommanderStorm:
```suggestion
        let value_cnt = u32::try_from(values.len())?;
        let offsets = (0..value_cnt).collect::<Vec<_>>()
```

rust/mlt-core/src/layer/v01/stream/mod.rs
```diff
@@ -347,12 +350,99 @@ impl OwnedStream {
...
+            encoding.dict_lengths,
+            StreamType::Offset(OffsetType::String),
+        )?;
+
+        Ok(OwnedEncodedStrProp::FsstDictionary {
+            symbol_lengths: symbol_length_stream,
+            symbol_table: symbol_table_stream,
+            length: value_length_stream,
+            corpus: compressed_stream,
+            offset: offset_stream,
```
CommanderStorm, line 367: using `Range<u32>` would be possible too..

rust/mlt-core/src/layer/v01/property/strings.rs
```diff
@@ -70,6 +99,508 @@ impl StrEncoder {
...
+            corpus,
+            offset,
+        })
+    }
+
+    /// Build from a parsed stream list (2 = plain, 3 = dictionary, 4 or 5 = FSST). Does not validate stream types.
+    pub fn from_streams(vec: Vec<Stream<'a>>) -> Result<Self, MltError> {
+        let n = vec.len();
+        let mut it = vec.into_iter();
+        Ok(match n {
```
CommanderStorm: consider using as_slice and do strctural pattern matching on the actual slice contents.

Much cleaner and removes the .next() which cannot actually fail.

rust/mlt-core/src/layer/v01/property/strings.rs
```diff
@@ -70,6 +99,508 @@ impl StrEncoder {
...
+    }
+
+    /// Number of streams (for stream count varint).
+    #[must_use]
+    pub fn stream_count(&self) -> usize {
+        match self {
+            Self::Plain { .. } => 2,
+            Self::Dictionary { .. } => 3,
+            Self::FsstPlain { .. } => 4,
+            Self::FsstDictionary { .. } => 5,
```
CommanderStorm: nit: consider if maybe we want to `a: _` values here.
This way, we get a compilaton warning if a field is added.

rust/mlt-core/src/layer/v01/property/strings.rs
nyurik: from_streams is actually not needed - removed

rust/mlt-core/src/layer/v01/property/strings.rs
nyurik: removed

rust/mlt-core/src/layer/v01/stream/mod.rs
nyurik: i'm not certain i understand

CommanderStorm: Yea, that was just dumb. Sorry

CommanderStorm (approved):

PR #1027, opened 2026-03-03
chore(ts): improve readability of geometry conversion code

HarelM:
This should improve readability of geometry conversion code.
Moves conversion to use switch-case, remove unneeded factory, remove some redundant methods and introduce more reusable ones.
This has the benefit of being 20% shorter, which is not a goal but rather nice to have.


HarelM:
Seems like the code is not covered enough with test to facilitate this kind of refactoring.
I'll make it 100% covered in a different PR I guess to make sure this doesn't break anything...
The synthetic tests are nice, but the geometry coverage is at 70% right now, so I guess there's more type of geometries needed to fully cover all this code...

PR #1029, opened 2026-03-03
chore(ts): add coverage to `geometryVectorConverter`

HarelM:
This adds coverage to geometry vector converter code.
I also ran format as it wasn't ran for a while.
After the coverage was at 100% I did a small refactoring of the code to avoid unnecessary extra complicated test.

This is requisite for:
- #1027

As in order to do a refactoring it's better to have the code covered with tests.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -4,12 +4,6 @@ project(mlt-cpp LANGUAGES CXX)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
 include(AddCXXCompilerFlag)
 
-# Use the same deployment target as the parent project, if set.
-if (NOT MLT_OSX_DEPLOYMENT_TARGET)
-    set(MLT_OSX_DEPLOYMENT_TARGET 16)
-endif()
-set(CMAKE_OSX_DEPLOYMENT_TARGET ${MLT_OSX_DEPLOYMENT_TARGET})
-
 # generate compile_commands.json https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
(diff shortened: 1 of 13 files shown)


HarelM:
@nyurik @CommanderStorm should the autofix.ci run as part of this PR when there's no changes besides changes in typescript code?

CommanderStorm:
@HarelM It does not matter. pre-commit also runs on your PR.
Unless this creates an issue, I am not going to invest time into this

louwers:
It's quite hard to review this due to the formatting changes. I think main should always be in a formatted state.

nyurik:
don't we use precommit.ci for automatically fixing PRs to force formatting?

HarelM:
I can run formatting on a different branch if that would help. 
Generally speaking I've added the geometry encoder code and the unit tests for the converter, the rest is automatic formatting. Can you help me configure the pre-commit to run the formatting?

HarelM:
I've updated the branch here so I would expect pre-commit to fix/normalize formatting, but I still see files that the only change is formatting... What am I missing?

HarelM:
Nevermind, it removed the relevant formatting, this was a merge error where a "," was wrongly added. looks good otherwise.
Ready for review.
There's another PR that will solve the build issue, so this can be reviewed separately.

ts/src/decoding/fastPforDecoder.ts
```diff
@@ -93,7 +93,9 @@ export function createDecoderWorkspace(): FastPforDecoderWorkspace {
     };
 }
 
-export function createFastPforWireDecodeWorkspace(initialEncodedWordCapacity = 16): FastPforWireDecodeWorkspace {
+export function createFastPforWireDecodeWorkspace(
+    initialEncodedWordCapacity: number = 16,
```
HarelM, line 97: The formatting adds a "," here, which is not needed and generally "wrong".
@louwers do you know how I can fix that?

HarelM:
Actually the refactoring was needed in order to be able to test it more easily, I removed a class and replaced it with a type.
In any case, there's no real logic charge here, just more tests...

PR #1032, opened 2026-03-03
chore: let platforms handle synthetic test reporting/registration

louwers:
Remove reporting from shared synthetic test logic. Instead let platform handle it.

TS uses Vitest, C++ uses simple `console.log`.

### DIFF
--- a/test/synthetic/synthetic-test-tool/index.ts
+++ b/test/synthetic/synthetic-test-tool/index.ts
@@ -52,7 +52,7 @@ export type SyntheticCaseResult =
       error: unknown;
     };
 
-export class SyntheticTestRunner implements AsyncIterable<SyntheticCaseResult> {
+export class SyntheticTestRunner {
   shouldSkip(_testName: string): false | string {
     return false;
   }
@@ -112,7 +112,7 @@ export class SyntheticTestRunner implements AsyncIterable<SyntheticCaseResult> {
         testName,
       };
     } catch (error) {
-      const _actualFile = await this.writeActualOutput(mltFile, actual);
+      await this.writeActualOutput(mltFile, actual);
       return {
         status: "fail",
         testName,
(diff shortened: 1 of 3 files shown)


test/synthetic/synthetic-test-tool/index.ts
```diff
@@ -31,7 +36,33 @@ expect.addEqualityTesters([
   },
 ]);
 
-export class SyntheticTestRunner {
+export type SyntheticCaseResult =
```
HarelM, line 39: Consider removing testName and keep only name.

test/synthetic/synthetic-test-tool/index.ts
```diff
@@ -31,7 +36,33 @@ expect.addEqualityTesters([
...
+      reason: string;
+    }
+  | {
+      status: "fail";
+      testName: string;
+      name: string;
+      error: unknown;
+    }
+  | {
+      status: "crash";
```
HarelM: Looking at the code I would recommend keeping only fail and have the reason as "failed decoding + exception message" instead of critical.

ts/src/synthetic.spec.ts
```diff
@@ -50,10 +47,35 @@ class SyntheticTestRunnerJS extends SyntheticTestRunner {
...
+describe("MLT Decoder - Synthetic tests", async () => {
+    const runner = new SyntheticTestRunnerJS();
+
+    for await (const result of runner) {
+        switch (result.status) {
+            case "skip":
+                it.skip(`${result.name} (${result.reason})`, () => undefined);
+                break;
+            case "ok":
+                it(result.name, () => undefined);
```
HarelM: This is looks odd, let me know when you have a few minutes to talk, I'm failing to see the reasoning behind these decisions.

louwers: This registers the test result with Vitest.

HarelM:
Both options are very much similar now, removing the need for inheritance from this code, which I think is not really needed here and makes things a bit more complicated without good justification, would even make them more similar.
I can change my code to use console logging for c++, but I really think vitest reporting has better visualization (time the tests took, color for skip, failed and success, etc...).
Let me know how you would like to proceed.

PR #1035, opened 2026-03-04
chore(ts, cpp): share code between ts and cpp for synthetic tests

HarelM:
This is another take on how can we share code between ts and c++.
It creates a utils package with a single file that has all kinds of methods to facilitate code sharing.
The rest is done in the relevant platform.
It uses vitest in both cases to report success, failure etc.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1547,6 +1547,7 @@ dependencies = [
  "mvt-reader",
  "num-traits",
  "num_enum",
+ "pretty_assertions",
  "probabilistic-collections",
  "proptest",
  "proptest-derive",
--- a/rust/mlt-core/Cargo.toml
+++ b/rust/mlt-core/Cargo.toml
@@ -44,6 +44,7 @@ zigzag.workspace = true
 criterion.workspace = true
 geo-types = { workspace = true, features = ["arbitrary"] }
 insta.workspace = true
+pretty_assertions.workspace = true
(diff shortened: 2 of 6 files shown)


HarelM:
This now passes for CPP with the following report:
image

cpp/tool/synthetic.test.ts
```diff
@@ -0,0 +1,49 @@
...
+describe("MLT Decoder - Synthetic tests", () => {
+  expect.addEqualityTesters([compareWithTolerance]);
+  const testCases = getTestCases(syntheticDir, SKIPPED_TESTS);
+  for (const { name, content, fileName } of testCases.active) {
+    it(name, async () => {
+      const actual = await decodeMLT(fileName);
+      try {
+        expect(actual).toEqual(content);
+      } catch (error) {
+        writeActualOutput(fileName, actual);
```
louwers, line 32: Much of this code is duplicated now.

HarelM: I'm ok with duplicating a little bit of code when it comes to tests readability. 
Also the try-catch is mainly for you as I just wrote the actual results to files regardless of the results in typescript...

HarelM:
I've removed the need to pass the synthetic folder, didn't reduce much code, but I didn't mind removing it.

HarelM:
@nyurik @CommanderStorm Rust CI is flickering, can you please take care of this?

PR #1036, opened 2026-03-04
feat(cpp): Allow FastPFOR decoding support to be controlled at runtime

TimSylvester:
Also fixes a minor bug where the geometry decoder was creating its own instance of integer decoder rather than reusing the one created by the top-level decoder.


TimSylvester:
Merged into #817

PR #1039, opened 2026-03-05
fix: remve HLL metrics collection from DataProfile since it is not usefull for integer streams

CommanderStorm:
@michaelkirk was right, obviously.

In the intial design, I confused how strings actually work with how I thought they worked.


CommanderStorm:
accidentally merged as part of #1043

PR #1042, opened 2026-03-05
feat(rust): property encoder optimiser

CommanderStorm:
Stacked on top of https://github.com/maplibre/maplibre-tile-spec/pull/1039

This PR adds the property and ID optimisers.
Both are fairly feautre rich and should provide good results.

Overall, IDs are fairly efficent, but strings I could not get past `O(n^2 + n * m)`[^1][^2] even with HLL and MinHash. Given that that square is comparing two hashes (i.e. fast for a PC and vectorizable), that is likely fine.

Disclaimer:
I took advantage of claude to generate the test data and some of the more mechanical parts of the rest, I am not that creative for test data (except tha colourado joke) and the rest was just plumbing it up, so nothing special.
All parts of the code are something I can fully stand behind.

[^1]: n = number of columns with PropValue::Str
[^2]: m = average number of non-null values per column

### DIFF
--- a/rust/mlt-core/src/layer/v01/property/optimizer.rs
+++ b/rust/mlt-core/src/layer/v01/property/optimizer.rs
@@ -50,7 +50,7 @@ struct StringProfile {
 ///    and compute a `MinHash` signature over them.
 /// 2. **Group**
 ///    compare [`MinHash`] signatures with Jaccard similarity; columns
-///    whose similarity exceeds [`MINHASH_SIMILARITY_THRESHOLD`] are clustered
+///    whose similarity exceeds `MINHASH_SIMILARITY_THRESHOLD` are clustered
 ///    into a shared dictionary.  The struct column name is derived from the
 ///    longest common prefix of the grouped property names.
 /// 3. **Compete & Select** - choose the best `IntEncoder` for integer columns


rust/mlt-core/src/layer/v01/property/optimizer.rs
```diff
@@ -0,0 +1,665 @@
...
+
+    /// Cluster string profiles into groups using greedy [`MinHash`] Jaccard merging.
+    ///
+    /// Returns a `Vec` of groups; each group is a sorted list of `col_idx`
+    /// values.  Every string column appears in exactly one group.  Singleton
+    /// groups (length 1) represent standalone scalar string columns.
+    fn group_strings(
+        profiles: &[StringProfile],
+        min_hash: &MinHash<IntoIter<String>, String>,
+    ) -> Vec<Vec<usize>> {
```
CommanderStorm: I don't know if MinHash LSH banding + union-find would improve or decrease performance here.
Need to check.

Algorithmically, it is better... but in practice? Not sure.

CommanderStorm: actually, lets always add union find.
LSH banding not sure still.

rust/mlt-core/src/layer/v01/property/optimizer.rs
```diff
@@ -0,0 +1,627 @@
+//! Optimizer that analyses a batch of [`DecodedProperty`] values in a
+//! "Profile -> Group -> Compete -> Select" pipeline and produces a
+//! [`MultiPropertyEncoder`] with near-optimal per-column encoding settings.
+
+use std::collections::hash_set::IntoIter;
```
nyurik, line 5: what about ID and Geometry columns? Wouldn't they also benefit from the same optimization, e.g. given a list of IDs, first compute that they must be stored as `Presence + Int32`, and then optimize both the presence stream and the Int32 stream?

rust/mlt-core/src/layer/v01/property/optimizer.rs
```diff
@@ -0,0 +1,618 @@
...
+            let presence = to_presence(count_nulls(v));
+            // FIXME: inaccurate, but encoders don't support u8 widely. Sometimes, plain might be more efficient for this, but is estimated less effective
+            let non_null: Vec<u32> = v.iter().flatten().copied().map(u32::from).collect();
+            PropertyEncoder::Scalar(ScalarEncoder::int(
+                presence,
+                IntEncoder::auto_u32(&non_null),
+            ))
+        }
+        PropValue::I32(v) => {
+            let presence = to_presence(count_nulls(v));
```
nyurik, line 326: my limited understanding is that we have zigzag for signed ints. Moreover, protobuf itself defines 3 types of integers - ints (mostly positive), sints (lots of negatives), uints (all positive) -- and these 3 optimize differently.  Do we have any of that?

rust/mlt-core/src/layer/v01/property/optimizer.rs
```diff
@@ -0,0 +1,618 @@
...
+        }
+        PropValue::SharedDict => {
+            // `SharedDict` columns are produced by the encoder, not consumed
+            // as decoder input.  Hitting this branch means the caller passed
+            // invalid data, which is a programming error.
+            debug_assert!(
+                false,
+                "PropValue::SharedDict must not appear in decoded input"
+            );
+            PropertyEncoder::Scalar(ScalarEncoder::bool(PresenceStream::Absent))
```
nyurik, line 387: Are we mixing two distinct concepts into one?    Is this what you check here?
* Logical column storage - the way columns look from the user's perspective
* Encoded shared-dictionary storage - the way shared-dict columns are serialized

Some hypothetical workflows:
* Synthetics mode: User constructs exact data structure, including shared dict and all encoding params, and just needs it to be encoded. Should be perfectly roundtrippable MLT->decoded->back to MLT (although some algorithms might produce slightly different bit patterns)
* Full auto mode: User gives a set of independent property columns, and the optimizer picks some to be similar enough to group into shared dictionaries -- this seems what you are doing here
* Semi-auto mode (uncertain if we want to support this, but may be easy enough to do and avoids debug_assert):  user sets up shared dicts by hand, but lets optimizer pick the correct encoding of each stream

rust/mlt-core/src/layer/v01/property/optimizer.rs
CommanderStorm: each IDs and geometrys has distinct challenges.
Presence for IDs is for both the least challenging thing.
I would not share code here because it is so simple and this sounds like false coupling.

rust/mlt-core/src/layer/v01/property/optimizer.rs
CommanderStorm: Yes, we the optimiser picks the optimal solution for each logical and shared dictionarys.
That was the goal.

PR #1043, opened 2026-03-05
feat(rust): Id optimiser impl

CommanderStorm:
Stacked on top of https://github.com/maplibre/maplibre-tile-spec/pull/1039

This PR adds the ID optimiser.
It is are fairly feautre rich and should provide good results in resonable time.

Disclaimer:
I took advantage of claude to generate the test data and some of the more mechanical parts of the rest, I am not that creative for test data and the rest was just plumbing it up, so nothing special.
All parts of the code are something I can fully stand behind.

### DIFF
--- a/rust/mlt-core/src/layer/v01/id/optimizer.rs
+++ b/rust/mlt-core/src/layer/v01/id/optimizer.rs
@@ -64,14 +64,12 @@ impl IdOptimizer {
         let mut has_nulls = false;
         let mut is_sequential = true;
         let mut is_constant = true;
-        let mut max_value;
 
         // Find first non-null
         let mut ids_iter = ids.iter();
         let first_non_null = loop {
             if let Some(opt_id) = ids_iter.next() {
                 if let Some(id) = opt_id {
-                    max_value = *id;
                     break *id;
                 }
                 has_nulls = true;
@@ -81,7 +79,9 @@ impl IdOptimizer {
             }
         };
 
-        let mut prev_non_null: u64 = first_non_null;
+        let mut max_value = first_non_null;
+        let mut prev_non_null = first_non_null;
+
         for &id in ids_iter {
             match id {
                 None => has_nulls = true,


rust/mlt-core/src/layer/v01/id/optimizer.rs
```diff
@@ -0,0 +1,173 @@
...
+/// 1. If the input contains no data, immediately return a zero-cost default.
+/// 2. **Single-pass deduction**
+///    One iteration over `Vec<Option<u64>>` collects `has_nulls`, `max_value`,
+///    `is_sequential`, and `is_constant` simultaneously.  `IdWidth` is derived
+///    deterministically from `has_nulls` and `max_value` - it is not a choice
+///    but a strict consequence of the data.
+/// 3. **Select `LogicalEncoder`**
+///    * **Fast-path:**
+///      * `is_sequential` -> `DeltaRle`
+///      * `is_constant` -> `Rle`.
```
nyurik, line 22: how can an ID be constant?  Per https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features

> A feature MAY contain an id field. If a feature has an id field, the value of the id SHOULD be unique among the features of the parent layer.

rust/mlt-core/src/layer/v01/id/optimizer.rs
```diff
@@ -0,0 +1,173 @@
...
+        // Fast-path: all consecutive non-null values increment by exactly 1.
+        // DeltaRle is optimal; skip allocation and trial encoding.
+        if is_sequential && ids.len() > 4 {
+            return IdEncoder::new(LogicalEncoder::DeltaRle, id_width);
+        }
+
+        // Fast-path: every non-null value is identical.
+        // Rle is optimal; skip allocation and trial encoding.
+        if is_constant {
+            return IdEncoder::new(LogicalEncoder::Rle, id_width);
```
nyurik, line 56: I think we may want to introduce some `warning` api - be able to report a suspicious data, e.g. all IDs are the same in a single layer.

rust/mlt-core/src/layer/v01/id/optimizer.rs
CommanderStorm: maybe, just maybe forrests I thought.

image

PR #1044, opened 2026-03-05
test(rust): Fuzz the return path for roundtrip-ability

CommanderStorm:
Did not find something in a bit, will run it overnight


.github/workflows/rust.yml
```diff
@@ -41,6 +41,7 @@ jobs:
       matrix:
         include:
           - target: layer
+          # - target: decoded_layer
```
CommanderStorm, line 44: strategy: if the overnight run finds something, we commit this change

```suggestion
          - target: decoded_layer
```

nyurik (approved):

PR #1045, opened 2026-03-05
feat(rust): Geometry optimiser

CommanderStorm:
This PR adds the geometry optimiser code.

I still need to do a thorugh self-review, but it is getting a bit late here and I need to take a break.
Will continue working on this after the break.

Opening here mostly to get the copilot review

### DIFF
--- a/rust/mlt-core/Cargo.toml
+++ b/rust/mlt-core/Cargo.toml
@@ -42,9 +42,9 @@ zigzag.workspace = true
 
 [dev-dependencies]
 criterion.workspace = true
-pretty_assertions.workspace = true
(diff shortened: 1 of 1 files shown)


nyurik:
sigh, it was set to auto-merge :'(

CommanderStorm:
> Do we encode everything twice? E.g. if we run through all encoding variants, pick one, and encode with that one, does it mean we do N+1 encodings? If so, can we keep the result of each encoding just in case it is the smallest, and use that instead of re-encoding it?

That is something for an optimisation pass ^^
I initially wanted this as simple as possible.

PR #1049, opened 2026-03-05
Automatically format ts code with pre-commit

louwers:
Not sure if @HarelM likes pre-commit, but if you want to use it, here's how you'd do it.

I like it because you never have to worry about formatting manually anymore.


nyurik:
doesn't look like it was - @louwers could you perhaps just change the config files, and let precommit commit another PR?

HarelM:
@louwers this can't be ran locally without installing pre-commit:
`sh: pre-commit: command not found`

I suggest to remove it from the package.json script if this is the case...

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
```diff
@@ -4,11 +4,8 @@ import { decodeFsst } from "../../decoding/fsstDecoder";
 import { decodeString } from "../../decoding/decodingUtils";
 
 export class StringFsstDictionaryVector extends VariableSizeVector<Uint8Array, string> {
-    private readonly textEncoder: TextEncoder;
```
HarelM, line 7: @louwers are you sure about this tool?
Seems like it broke the build due to this removal...

HarelM:
Oh, I accidentally merged this although the build did not pass... damn...

louwers:
@HarelM That's probably because in pre-commit we configured the `--unsafe` flag for Biome, which can make potentially breaking changes. Thanks for the follow-up to fix it...

PR #1052, opened 2026-03-05
chore(rust): bump fastpfor, enable SIMD test, cleanup

nyurik:
* re-enable SIMD test because now all FastPFOR code is compiled with portable default
* MLT ui resize is now correctly working

### DIFF
--- a/ts/src/decoding/fastPforDecoder.ts
+++ b/ts/src/decoding/fastPforDecoder.ts
@@ -93,7 +93,9 @@ export function createDecoderWorkspace(): FastPforDecoderWorkspace {
     };
 }
 
-export function createFastPforWireDecodeWorkspace(initialEncodedWordCapacity = 16): FastPforWireDecodeWorkspace {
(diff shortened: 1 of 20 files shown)


rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -561,7 +561,7 @@ impl ScalarValueEncoder {
             ScalarValueEncoder::String(_) => "string",
             ScalarValueEncoder::Float => "float",
             ScalarValueEncoder::Bool => "bool",
-            ScalarValueEncoder::Struct => "struct",
+            ScalarValueEncoder::SharedDict => "struct",
```
CommanderStorm: like this?

```suggestion
            ScalarValueEncoder::SharedDict => "shared_dict",
```

nyurik: i wasn't sure what its used for, but sure :)

PR #1053, opened 2026-03-05
refactor(rust): move geotype code to its own mod

nyurik:


CommanderStorm:
merging to make my life simpler

PR #1062, opened 2026-03-06
feat(rust): add wasm support

CommanderStorm:
Very much unoptimised, maybe we are even not going to merge this.

Not sure

Current fairly unoptimised version is:
```
156K	pkg/mlt.wasm
52K	pkg/mlt.wasm.gz
```

This compares to the JS version:

```
64K	bundle.js
20K	bundle.js.gz
```

This is a bit of a flase comparison, wasm can be streamed in as fast as the network can hold, while JS needs to be parsed & such.

### DIFF
--- a/rust/mlt-wasm/js/synthetic.spec.ts
+++ b/rust/mlt-wasm/js/synthetic.spec.ts
@@ -33,6 +33,8 @@ const UNIMPLEMENTED_SYNTHETICS = new Map([
   ["prop_f64_nan", "needs investigation"],
   ["prop_u32_null_val", "needs investigation"],
   ["prop_u32_val_null", "needs investigation"],
+  ["prop_i32_null_val", "needs investigation"],
+  ["prop_i32_val_null", "needs investigation"],
 ]);
 
 describe("MLT WASM Decoder - Synthetic tests", () => {


test/synthetic/synthetic-test-utils/index.ts
```diff
@@ -1,4 +1,4 @@
-import { globSync, readFileSync, writeFileSync } from "node:fs";
+import { readdirSync, readFileSync, writeFileSync } from "node:fs";
```
nyurik, line 1: lets not mix this change with webasm

rust/mlt-wasm/js/synthetic.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+import { decodeTile } from "./vectorTile";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const EARCUT_MAX_RINGS = 500;
+
+// The WASM decoder handles F64, U32/U64 natively — skip list is much shorter
+// than the TS decoder's.
+const UNIMPLEMENTED_SYNTHETICS = new Map([
+  ["poly_multi_morton_ring_morton", "decodes to zero coordinates"],
+  ["poly_multi_morton_ring_no_morton", "decodes to zero coordinates"],
```
CommanderStorm:
```suggestion
const UNIMPLEMENTED_SYNTHETICS = new Map([
```

rust/mlt-core/src/layer/v01/root.rs
```diff
@@ -179,6 +179,15 @@ impl Layer01<'_> {
         }
     }
 
+    /// Decode only the geometry and ID columns, leaving properties in their encoded form.
+    ///
+    /// Use this instead of [`Self::decode_all`] when properties will be accessed lazily
+    pub fn decode_geometry_and_id(&mut self) -> Result<(), MltError> {
```
nyurik, line 185: while this is fine for now, i suspect even the ID and geometries are optional in some cases, e.g. you may want to filter by geometry type and not decode some parts of it, or even filter by a property and decide not to draw anything at all in a layer, so no need for geometry in some cases...

rust/mlt-core/src/utils/decode.rs
```diff
@@ -211,21 +215,187 @@ pub fn decode_fastpfor_composite(data: &[u8], num_values: usize) -> Result<Vec<u
...
+/// bitmap+1..         for each set bit k (2..=32):
+///                        [0]  size  — number of exceptions at this width
+///                        [1..] bit-packed exception values
+///                               words = greatest_multiple(size+31, 32) * k / 32
+///                               (with overflow adjustment)
+/// ```
+///
+/// `n_primary` must be a multiple of the block size (256).
+#[cfg(feature = "fastpfor-cpp")]
+fn fastpfor_page_scanner(input: &[u32], n_primary: usize) -> Result<usize, MltError> {
```
nyurik: i'm fine with this here for now - i will need to understand better the reason for this - but do consider upstreaming it to fastpfor-rs crate if it makes sense

rust/mlt-core/src/utils/encode.rs
```diff
@@ -121,24 +124,102 @@ pub fn encode_byte_rle(data: &[u8]) -> Vec<u8> {
     output
 }
 
-/// Encode a `u32` sequence using `FastPFOR256` (composite codec).
+/// Encode a `u32` sequence using the composite FastPFOR+VarInt codec.
```
nyurik: is this for decoding or encoding??  If encoding, please move it to a separate PR

rust/mlt-core/src/layer/v01/root.rs
CommanderStorm: would need the API contract with maplibre-native changed.

Lets see what the Bucketing PR will bring in that regard

rust/mlt-core/src/utils/encode.rs
```diff
@@ -128,17 +127,48 @@ pub fn encode_fastpfor(values: &[u32]) -> Result<Vec<u8>, MltError> {
...
+        // Over-allocate: FastPFOR may write a header and padding beyond the input length.
+        let mut compressed = vec![0u32; values.len() + 1024];
+        let mut comp = Composition::new(FastPFOR::default(), VariableByte::new());
+        let mut input_offset = std::io::Cursor::new(0u32);
+        let mut output_offset = std::io::Cursor::new(0u32);
+
+        comp.compress(
+            values,
+            u32::try_from(values.len())?,
+            &mut input_offset,
```
CommanderStorm:
```suggestion
        let mut output_offset = std::io::Cursor::new(0u32);

        comp.compress(
            values,
            u32::try_from(values.len())?,
            &mut std::io::Cursor::new(0u32),
```

HarelM:
A bit late to the party, but I suggest to use the synthetic-tools ts file in order to reuse some synthetic test code.
Also use the classify-rings from the style spec package instead of rewriting it here?

PR #1063, opened 2026-03-06
fix(ts): revert double as float

HarelM:
Fix unit tests, synthetic tests and the relevant parts of the code relate to double as double instead of double as float.

Mostly reverts 
- #965


HarelM:
Only 4 more skipped tests to fix! yay! :-)

PR #1065, opened 2026-03-06
feat(java): Optimize buffer handling

TimSylvester:
Use collections of individual buffers to avoid copying data when returning a subset or concatenation of data at each stage.  Previously each small part of the stream was concatenated into a new, larger array many times before the final byte array was formed, and each varint encoding involved an array copy.

~35% faster in my tests

Before:
	Command being timed: "java -Xmx96G -Xms48G -Dcom.google.protobuf.use_unsafe_pre22_gencode --enable-native-access=ALL-UNNAMED -jar mlt-cli/build/libs/encode.jar --pmtiles ../../../../Downloads/planet-20260112.pmtiles --mlt planet-20260112.mlt.gzip.max10.pmtiles --outlines ALL --colmap-delim []name/[:_]/ --sort-ids --coerce-mismatch --maxzoom 10 --parallel --verbose debug"
	User time (seconds): 799.56
	System time (seconds): 1835.72
	Percent of CPU this job got: 771%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 5:41.48

After:
	Command being timed: "java -Xmx96G -Xms48G -Dcom.google.protobuf.use_unsafe_pre22_gencode --enable-native-access=ALL-UNNAMED -jar mlt-cli/build/libs/encode.jar --pmtiles ../../../../Downloads/planet-20260112.pmtiles --mlt planet-20260112.mlt.gzip.max10.pmtiles --outlines ALL --colmap-delim []name/[:_]/ --sort-ids --coerce-mismatch --maxzoom 10 --parallel --verbose debug"
	User time (seconds): 652.14
	System time (seconds): 1106.63
	Percent of CPU this job got: 808%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 3:37.64

### DIFF
--- a/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/Encode.kt
+++ b/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/Encode.kt
@@ -359,7 +359,7 @@ object Encode {
                 if (result.isPresent) {
                     logger.warn("Tiles do not match: {}", result)
                 } else {
-                    logger.debug("Tiles match")
+                    logger.debug("Tiles match: {}:{},{}", z, x, y)
                 }
             }
         }
--- a/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/PMTiles.kt
+++ b/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/PMTiles.kt
@@ -205,7 +205,7 @@ private fun encodePMTiles(
             }
 
             if (state.success.get()) {
-                logger.debug("Finalizing MBTiles file")
+                logger.debug("Finalizing PMTiles file")
                 writer.finish(newMetadata)
             }
 


java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/DoubleEncoder.java
```diff
@@ -35,10 +38,9 @@ public static byte[] encodeDoubleStream(
             .encode();
 
     streamObserver.observeStream(
```
CommanderStorm: Already discussed on slack, but can be removed.

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/EncodingUtils.java
```diff
@@ -66,7 +69,7 @@ public static byte[] encodeDoublesLE(double[] values) {
 
   // Source:
   // https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/util/VarInt.java
-  public static byte[] encodeVarints(int[] values, boolean zigZagEncode, boolean deltaEncode)
+  public static ByteBuffer encodeVarints(int[] values, boolean zigZagEncode, boolean deltaEncode)
```
CommanderStorm: doing both varint, zigzag and delta is a bit hard on the JIT..
If the goal is perf, this might (or might not) have an effect

java/mlt-core/src/main/java/org/maplibre/mlt/util/ByteBufferUtil.java
```diff
@@ -0,0 +1,52 @@
...
+            b -> {
+              // Use `duplicate` to avoid changing the position of the input buffers
+              result.put(b.duplicate());
+            });
+    return result.flip();
+  }
+
+  public static int totalLength(Iterable<ByteBuffer> buffers) {
+    return streamOf(buffers).mapToInt(ByteBuffer::remaining).sum();
+  }
```
CommanderStorm: the `totalLength` is the sum of [`remaining`](https://docs.oracle.com/javase/8/docs/api/java/nio/Buffer.html#remaining--)?

Can we name it `totalRemainingLength` instead to avoid the mistake-potential that we get remaining vs total wrong?

Also the usage of this does not seem entirely above board as far as I looked.
Sometimes, this remaining seems intentional, sometimes I am not sure.
=> Please look at this change if this is what you intended again

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/DoubleEncoder.java
TimSylvester: I was planning to do that in a separate PR.  Do you think it's better to include it here?

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/EncodingUtils.java
TimSylvester: It might be a good idea to separate those, but the idea here is just to eliminate the huge number of array copies.

java/mlt-core/src/main/java/org/maplibre/mlt/util/ByteBufferUtil.java
TimSylvester: Good call on the naming.  Implicit here is that the buffers are not consumed at all, only accumulated for concatenation.

I started at the low level, using `ByteBuffer` to return a "view" of the result buffer from varint encoding instead of making a copy into a new buffer.  But that turned out to be the only place it's actually useful, so I'm considering going back to byte arrays.

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/DoubleEncoder.java
CommanderStorm: Fine by me

java/mlt-core/src/main/java/org/maplibre/mlt/util/ByteBufferUtil.java
CommanderStorm: It is not naming that I am concerned about. How this method is USED seems kind of buggy.

If you have checked and you think it is not buggy, I am fine with it.

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/DoubleEncoder.java
TimSylvester: If I'm' going back to byte[] it makes sense to remove it.

PR #1066, opened 2026-03-06
refactor(rust): major rework of the shared dict

nyurik:
there are a few fixme-s in the code - could be done as part of this PR, or possibly separate

### DIFF
--- a/rust/mlt-core/tests/property_optimizer.rs
+++ b/rust/mlt-core/tests/property_optimizer.rs
@@ -32,38 +32,36 @@ fn similar_strings_grouped_into_shared_dict() {
     assert_eq!(items[1].suffix, "de");
 
     // Encoder should be SharedDict
-    assert_debug_snapshot!(enc, @r#"
-    MultiPropertyEncoder {
-        properties: [
-            SharedDict(
-                SharedDictEncoder {
-                    dict_encoder: Plain {
-                        string_lengths: IntEncoder {
-                            logical: None,
+    assert_debug_snapshot!(enc, @"
+    [
+        SharedDict(
+            SharedDictEncoder {
+                dict_encoder: Plain {
+                    string_lengths: IntEncoder {
+                        logical: None,
+                        physical: VarInt,
+                    },
+                },
+                items: [
+                    SharedDictItemEncoder {
+                        optional: Absent,
+                        offset: IntEncoder {
+                            logical: Delta,
                             physical: VarInt,
                         },
                     },
-                    items: [
-                        SharedDictItemEncoder {
-                            optional: Absent,
-                            offset: IntEncoder {
-                                logical: Delta,
-                                physical: VarInt,
-                            },
+                    SharedDictItemEncoder {
+                        optional: Absent,
+                        offset: IntEncoder {
+                            logical: Delta,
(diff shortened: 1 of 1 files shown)


rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -381,7 +395,10 @@ impl Analyze for PropValue {
             Self::F32(v) => v.collect_statistic(stat),
             Self::F64(v) => v.collect_statistic(stat),
             Self::Str(v) => v.collect_statistic(stat),
-            Self::SharedDict => 0,
+            Self::SharedDict(items) => items
+                .iter()
+                .map(|item| item.values.collect_statistic(stat))
+                .sum(),
```
CommanderStorm, line 401: I think this is not quite corect in terms of the accounting for the different types of statistics.

rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -597,19 +593,12 @@ pub enum PresenceStream {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct MultiPropertyEncoder {
     pub(crate) properties: Vec<PropertyEncoder>,
```
CommanderStorm: should likely be removed

rust/mlt-core/src/layer/v01/property/optimizer.rs
```diff
@@ -597,16 +526,56 @@ mod tests {
...
+            make_prop(
+                "city:de",
+                PropValue::Str(vec![
+                    Some("Munich".to_string()),
+                    Some("Manheim".to_string()),
+                    Some("Garching".to_string()),
+                ]),
+            ),
+            make_prop(
+                "city:Colorado",
```
CommanderStorm, line 561: I like that you kept the dad joke 😉

rust/mlt-synthetics/src/main.rs
```diff
@@ -877,62 +877,59 @@ fn generate_shared_dictionaries(w: &SynthWriter) {
...
+            SharedDict::new("name", SE::plain(E::varint()))
+                .column(":de", O::Present, E::varint(), [Some(long_string_value())])
+                .column(":en", O::Present, E::varint(), [Some(long_string_value())]),
+        )
+        .add_shared_dict(
+            SharedDict::new("name", SE::plain(E::varint()))
+                .column(":fr", O::Present, E::varint(), [Some(long_string_value())])
+                .column(":he", O::Present, E::varint(), [Some(long_string_value())]),
+        )
+        .write("props_shared_dict_2_same_prefix-rust");
```
CommanderStorm, line 917: Makes optimising a lot harder..
Will see what I can do.

rust/mlt-core/src/layer/v01/property/mod.rs
CommanderStorm: removed

rust/mlt-core/src/layer/v01/property/mod.rs
CommanderStorm: actually, this is correct..
But only because of where it is called and how..

I don't like how hard this system is to keep correct

PR #1070, opened 2026-03-08
fix(ts): allow parsing of unsigned int32 and int64 values.

HarelM:
I'm not 100% sure this is the correct way to fix this, but I implemented things similar to how I fixed the int64 tests.
zigzag is trnasforming unsigned to signed and vice-versa.
Other are keeping the relevant sign and using the right array (I hope).


HarelM:
@DoFabien can you take a look at the changes I made related to fastPfor?

CommanderStorm:
(assuming you fix the argument passing issue CI is complaining about)

HarelM:
I've fixed the CI, I'll give @DoFabien a few days to comment, and if everything is well I'll merge this and publish a new version.

PR #1073, opened 2026-03-08
feat(rust): Integrate mlt-wasm release into our release pipeline

CommanderStorm:
https://www.npmjs.com/package/@maplibre/mlt-wasm

### DIFF
--- a/rust/mlt-wasm/src/properties.rs
+++ b/rust/mlt-wasm/src/properties.rs
@@ -1,7 +1,4 @@
-use js_sys::{
-    Array, BigInt64Array, BigUint64Array, Float32Array, Float64Array, Int8Array, Int32Array,
-    Uint8Array, Uint32Array,
-};
+use js_sys::{Array, Float32Array, Float64Array, Int8Array, Int32Array, Uint8Array, Uint32Array};
 use mlt_core::v01::{OwnedProperty, PropValue};
 use wasm_bindgen::prelude::*;
 
@@ -137,8 +134,13 @@ pub(crate) fn prop_values_to_js_column(pv: &PropValue, n: u32) -> JsValue {
                 }
                 arr.into()
             } else {
-                let buf = v.iter().flatten().copied().collect::<Vec<_>>();
-                BigInt64Array::from(buf.as_slice()).into()
+                let buf = v
+                    .iter()
+                    .flatten()
+                    .copied()
+                    .map(|n| n as f64)
+                    .collect::<Vec<_>>();
+                Float64Array::from(buf.as_slice()).into()
             }
         }
         PropValue::U64(v) => {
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -204,3 +252,63 @@ jobs:
...
+      - name: Release (GitHub) - Create Draft
+        uses: softprops/action-gh-release@a06a81a03ee405af7f2048a818ed3f03bbf83c7b # v2.5.0
+        with:
+          name: mlt-wasm-js-v${{ env.version }}
+          tag_name: mlt-wasm-js-v${{ env.version }}
+          prerelease: ${{ env.prerelease }}
+          draft: true
+
+      - name: Publish npm package
+        run: npm publish mlt-wasm-pkg/maplibre-mlt-wasm-*.tgz --access=public
```
HarelM, line 307: [Optional] Usually running this in the relevant folder avoids the need to specify the package name.

.github/workflows/rust.yml
```diff
@@ -59,27 +59,27 @@ jobs:
...
-#  test-msrv:
-#    name: Test MSRV
-#    runs-on: ubuntu-latest
-#    steps:
-#      - uses: actions/checkout@v6.0.2
-#      - uses: ./.github/actions/mlt-setup-java  # generate MLT test files
-#      - uses: ./.github/actions/mlt-setup-rust
-#        with: { toolchain: msrv }
-#      - run: just -v rust::ci-test-msrv
+  #  MSRV Testing is not really needed at this stage of code completeness
```
HarelM, line 62: Remove these comments?

rust/mlt-wasm/package.json
```diff
@@ -1,51 +1,51 @@
...
+    "build:wasm-gz": "gzip --best --no-name pkg/mlt_wasm_bg.wasm --keep --force && du -sh pkg/*.wasm*",
+    "build:ts": "tsc",
+    "build": "npm run build:wasm && npm run build:ts && npm run build:wasm-gz",
+    "test": "vitest --coverage --coverage.reportOnFailure --run",
+    "test:watch": "vitest --coverage --coverage.reportOnFailure --ui",
+    "bench": "vitest bench"
+  },
+  "dependencies": {
+    "@mapbox/point-geometry": "^1.1.0",
+    "@maplibre/vt-pbf": "4.3.0"
```
HarelM: vt-pbf package is only used for the types, maybe they can be used as devDependecny, or completely remove them?
Add style-spec to devDependencies and use that in the synthetic tests maybe?

CommanderStorm: `@maplibre/vt-pbf` is the API that gl-js would currently require.

Not sure if we will do that still in a few releases (given the bucketing work), but for now this is the correct package.

.github/workflows/rust.yml
CommanderStorm: does the tarball not also contain this?

Issue #1075, opened 2026-03-08
OMT fixture ID mismatch on large feature IDs


DoFabien:
While working on the TS unsigned fixes, I ran into a separate issue on some OMT fixtures: the MLT-decoded feature ID and the MVT feature ID do not line up reliably.

One concrete example is:

- tile: `test/expected/tag0x01/omt/10_530_682.mlt`
- layer: `aerodrome_label`
- embedded MLT metadata: `longID = false`
- ID stream: `FLAT`, `NONE`, `VARINT`
- decoded MLT ID: `4293407208`
- MVT ID seen from `@mapbox/vector-tile`: `18446744073707991000`
- raw MVT ID read directly : `18446744073707991528`

Two important points:

- `longID = false` means `uint32`, not signed `int32`, so `4293407208` is a valid MLT value
- the JS MVT ID is above `Number.MAX_SAFE_INTEGER`, so it is not reliable enough on its own

What makes this case interesting is that the raw MVT varint can be interpreted in different ways:

- as unsigned 64-bit: `18446744073707991528`
- as signed 64-bit: `-1560088`
- as unsigned 32-bit low bits: `4293407208`

So the current mismatch looks like an interpretation problem on large IDs.

I also saw similar mismatches on other OMT layers (`place`, `building`, `aerodrome_label`), so this does not look isolated to one tile

I think  the generated MLT fixture likely does not preserve the original ID semantics for this case

CommanderStorm:
MVT defines IDs as `uint64` => our IDs are also unsigned.

The rust decoder/encoder has this in order, the Java encoder might be a bit weird when it comes to this. Let me check

PR #1076, opened 2026-03-08
test(synthetic): More id synthetics

CommanderStorm:
reproducer for https://github.com/maplibre/maplibre-tile-spec/issues/1075

CC @DoFabien

### DIFF
--- a/java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
+++ b/java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
@@ -195,7 +195,7 @@ public class SyntheticMltGenerator {
     write(layer("ids64_opt", optIds64), cfg().ids());
     write(layer("ids64_opt_delta", optIds64), cfg(DELTA).ids());
 
-    // java doe not generate this as an ID64 if none of them are above the U32::Max threshold
+    // Java does not generate this as an ID64 if none of them are above the U32::Max threshold
     // FIXME: serialises as i64
     // var ids64MinMax = array(idFeat(0L), idFeat(0xFFFFFFFFFFFFFFFFL), idFeat(0L),
     // idFeat(0xFFFFFFFFFFFFFFFFL));


java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -205,9 +162,13 @@ private static void generateExtent() throws IOException {
   }
 
   private static void generateIds() throws IOException {
-    write("id0", idFeat(0), cfg().ids());
     write("id", idFeat(100), cfg().ids());
+    write("id_min", idFeat(0), cfg().ids());
+    // FIXME: serialises as i32
+    // write("id_max", idFeat(0xFFFF), cfg().ids());
     write("id64", idFeat(9_234_567_890L), cfg().ids());
+    // FIXME: serialises as i64
```
HarelM: What's the point of this comment?

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
```diff
@@ -233,13 +194,16 @@ private static void generateIds() throws IOException {
     var optIds64 = array(idFeat(), idFeat(9_234_567_890L), idFeat(101), idFeat(105), idFeat(106));
     write(layer("ids64_opt", optIds64), cfg().ids());
     write(layer("ids64_opt_delta", optIds64), cfg(DELTA).ids());
+
+    // java doe not generate this as an ID64 if none of them are above the U32::Max threshold
```
HarelM: Not sure I understand the point of adding these comments, maybe opening an issue or linking to one would be more appropriate here.

test/synthetic/0x01-rust/id_max-rust.json
HarelM, line 1: id32_max for the file name?

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
CommanderStorm: The point is to mark the the synthetics that should work but currently don't due to shortcomings.

For this, i64 means a different varInt encoding, since that does zigzag.

java/mlt-tools/src/main/java/org/maplibre/mlt/tools/SyntheticMltGenerator.java
CommanderStorm: Please see the rest of the file. The java encoder has many holes and irregularities, opening an issue for reach is too much work.

test/synthetic/0x01-rust/id_max-rust.json
CommanderStorm: Our naming scheme is Id and id64 for the other files.
So to be consistent it is id_max

PR #1077, opened 2026-03-08
fix(ts): preserve unsigned flat values

DoFabien:
Follow-up to #1070 .

There were still a few TS decoder paths where `UINT_32` / `UINT_64` flat values ended up in signed storage. In practice, values like `0xffffffff` and `0xffffffffffffffffn` could still come back as `-1` / `-1n`.

This keeps those paths unsigned:

- keep `UINT_32` flat values in `Uint32Array`
- keep `UINT_64` flat values in `BigUint64Array`
- fix the unsigned `CONST` ID path so const ID vectors do not fall back to signed storage

The signed decode paths stay untouched. This is limited to the remaining unsigned property / ID flow.

I also added regression tests for the `UINT_32` and `UINT_64` flat max-value cases.
The unsigned ID paths touched here are still covered by the existing `mltDecoder.spec.ts` tests.

There is also a small test-only guard in `mltDecoder.spec.ts` for MVT IDs that are not safe JS integers. That is only there to keep this PR focused; the separate fixture / ID issue is tracked in `#1075`.

### DIFF
--- a/ts/src/decoding/integerStreamDecoder.ts
+++ b/ts/src/decoding/integerStreamDecoder.ts
@@ -421,11 +421,7 @@ export function getVectorType(
     return streamMetadata.numValues === 1 ? VectorType.CONST : VectorType.FLAT;
 }
 
-function decodeRleInt32(
-    data: Int32Array,
-    streamMetadata: RleEncodedStreamMetadata,
-    isSigned: boolean,
-): Int32Array {
+function decodeRleInt32(data: Int32Array, streamMetadata: RleEncodedStreamMetadata, isSigned: boolean): Int32Array {
     return isSigned
         ? decodeZigZagRleInt32(data, streamMetadata.runs, streamMetadata.numRleValues)
         : new Int32Array(decodeUnsignedRleInt32(data, streamMetadata.runs, streamMetadata.numRleValues));
--- a/ts/src/mltDecoder.ts
+++ b/ts/src/mltDecoder.ts
@@ -8,9 +8,7 @@ import BitVector from "./vector/flat/bitVector";
 import {
     decodeConstIntStream,
     decodeConstLongStream,
-    decodeIntStream,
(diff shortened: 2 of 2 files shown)


ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -401,3 +446,15 @@ function decodeRleFloat64(
         ? decodeZigZagRleFloat64(data, streamMetadata.runs, streamMetadata.numRleValues)
         : decodeUnsignedRleFloat64(data, streamMetadata.runs, streamMetadata.numRleValues);
 }
+
+function reinterpretAsUint32(values: Int32Array): Uint32Array {
```
HarelM, line 450: These methods are redundant given that they are only one line of code and hide the actual implementation, which in this case doesn't help redability.
It's better to simply use `new XArray(...)`.

HarelM:
Should this PR be against my branch or am I missing something?

HarelM:
I'm confused with this PR, sorry, some of the things here are part of what I already did, but do not leverage my changes, moreover, some are incorrect form a possibility perspective (encoding singed int32 using varint will not work for negative numbders etc).

DoFabien:
Oh, my bad. 
Sorry, I got the base wrong.

I thought the work from `min-i32` had already landed on `main`,  so I opened this against `main`. That was my mistake, the right base here is your branch.
I’ll close this PR and reopen the follow-up on top of `min-i32` instead. I’ll pick this up again tomorrow.

HarelM:
I can merge the other one, I just want to make sure that what I did there is correct from fastPFor perspective.
We can iterate on the rest on main if the other PR is good to go.

DoFabien:
Yes, from what I checked, your PR looks fine from the `fastPFor` side

HarelM:
I've merged the other PR, feel free to reopen/recreate/redo this one.

PR #1078, opened 2026-03-08
fix(rust): migrate part of our lengths from usize to u32

CommanderStorm:
We are using usize in a few places.
For all types interacting with memory allocations (vec, ...), that is fine.

For the lenths we parse from varints? That seems buggy.


nyurik:
Actually, come to think of it, I don't think we should allow u64 sizes. I don't see any point of allowing layers larger than 4gb,  so perhaps u32?

PR #1079, opened 2026-03-08
fix(ts): preserve unsigned flat values

DoFabien:
Follow-up to #1070 .

There were still a few TS decoder paths where `UINT_32` / `UINT_64` flat values ended up in signed storage. In practice, values like `0xffffffff` and `0xffffffffffffffffn` could still come back as `-1` / `-1n`.

This keeps those paths unsigned:

- keep `UINT_32` flat values in `Uint32Array`
- keep `UINT_64` flat values in `BigUint64Array`
- fix the unsigned `CONST` ID path so const ID vectors do not fall back to signed storage

The signed decode paths stay untouched. This is limited to the remaining unsigned property / ID flow.

I also added regression tests for the `UINT_32` and `UINT_64` flat max-value cases.
The unsigned ID paths touched here are still covered by the existing `mltDecoder.spec.ts` tests.

There is also a small test-only guard in `mltDecoder.spec.ts` for MVT IDs that are not safe JS integers. That is only there to keep this PR focused; the separate fixture / ID issue is tracked in  #1075 .

### DIFF
--- a/ts/src/decoding/integerStreamDecoder.ts
+++ b/ts/src/decoding/integerStreamDecoder.ts
@@ -10,8 +10,6 @@ import {
     decodeUnsignedComponentwiseDeltaVec2Scaled,
     decodeUnsignedConstRleInt32,
     decodeUnsignedConstRleInt64,
-    decodeUnsignedDeltaRleInt32,
-    decodeUnsignedDeltaRleInt64,
     decodeUnsignedRleInt32,
     decodeUnsignedRleInt64,
     decodeUnsignedRleFloat64,
@@ -288,9 +286,6 @@ function decodeUnsignedInt32(
         case LogicalLevelTechnique.DELTA:
             if (streamMetadata.logicalLevelTechnique2 === LogicalLevelTechnique.RLE) {
                 const rleMetadata = streamMetadata as RleEncodedStreamMetadata;
-                if (!nullabilityBuffer) {
-                    return decodeUnsignedDeltaRleInt32(values, rleMetadata.runs, rleMetadata.numRleValues);
-                }
                 const deltaValues = decodeUnsignedRleInt32(values, rleMetadata.runs, rleMetadata.numRleValues);
                 decodedValues = decodeUnsignedZigZagDeltaInt32(deltaValues);
             } else {
@@ -386,9 +381,6 @@ function decodeUnsignedInt64(
         case LogicalLevelTechnique.DELTA:
             if (streamMetadata.logicalLevelTechnique2 === LogicalLevelTechnique.RLE) {
                 const rleMetadata = streamMetadata as RleEncodedStreamMetadata;
-                if (!nullabilityBuffer) {
-                    return decodeUnsignedDeltaRleInt64(values, rleMetadata.runs, rleMetadata.numRleValues);
-                }
                 const deltaValues = decodeUnsignedRleInt64(values, rleMetadata.runs, rleMetadata.numRleValues);
                 decodedValues = decodeUnsignedZigZagDeltaInt64(deltaValues);
             } else {
--- a/ts/src/decoding/integerDecodingUtils.ts
+++ b/ts/src/decoding/integerDecodingUtils.ts
@@ -604,25 +604,6 @@ export function decodeDeltaRleInt32(data: Uint32Array, numRuns: number, numValue
     return result;
 }
 
-export function decodeUnsignedDeltaRleInt32(data: Uint32Array, numRuns: number, numValues: number): Uint32Array {
-    const result = new Uint32Array(numValues);
-    let outPos = 0;
-    let previousValue = 0;
-
-    for (let i = 0; i < numRuns; i++) {
-        const runLength = data[i];
-        const zigZagDelta = data[i + numRuns];
-        const delta = decodeZigZagInt32Value(zigZagDelta);
-
-        for (let j = 0; j < runLength; j++) {
-            previousValue = (previousValue + delta) >>> 0;
-            result[outPos++] = previousValue;
-        }
-    }
-
-    return result;
-}
-
 /**
  * Decode Delta-RLE with multiple runs for 64-bit integers.
  */
@@ -638,25 +619,6 @@ export function decodeDeltaRleInt64(data: BigUint64Array, numRuns: number, numVa
 
         for (let j = 0; j < runLength; j++) {
             previousValue += delta;
-            result[outPos++] = previousValue;
-        }
-    }
-
-    return result;
-}
-
-export function decodeUnsignedDeltaRleInt64(data: BigUint64Array, numRuns: number, numValues: number): BigUint64Array {
-    const result = new BigUint64Array(numValues);
-    let outPos = 0;
-    let previousValue = 0n;
-
-    for (let i = 0; i < numRuns; i++) {
-        const runLength = Number(data[i]);
-        const zigZagDelta = data[i + numRuns];
-        const delta = decodeZigZagInt64Value(zigZagDelta);
-
-        for (let j = 0; j < runLength; j++) {
-            previousValue = BigInt.asUintN(64, previousValue + delta);
             result[outPos++] = previousValue;
         }
     }


ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -95,11 +105,10 @@ export function decodeConstIntStream(
     const values = decodePhysicalLevelTechnique(data, offset, streamMetadata);
 
     if (values.length === 1) {
-        const value = values[0];
-        return isSigned ? decodeZigZagInt32Value(value) : value;
+        return isSigned ? decodeZigZagInt32Value(values[0]) : values[0];
     }
 
-    return isSigned ? decodeZigZagConstRleInt32(values) : decodeUnsignedConstRleInt32(values);
+    return isSigned ? decodeZigZagConstRleInt32(values) : values[1];
```
HarelM: Are you sure about `values[1]`? This doesn't look correct...

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -54,6 +53,17 @@ export function decodeIntStream(
     return decodeInt32(values, streamMetadata, isSigned, scalingData, nullabilityBuffer);
 }
 
+export function decodeUnsignedIntStream(
```
HarelM: `decodeIntStream` is passed a `isSigned`, if you want to split these two I suggest to remove this parameter from the method input.

ts/src/decoding/integerStreamDecoder.ts
DoFabien: You're right on `values[1]`, that was too implicit.
I switched it back to `decodeUnsignedConstRleInt32(values)` for readability and consistency with the Int64 path.

ts/src/decoding/integerStreamDecoder.ts
DoFabien: I agree the API is a bit awkward here.

I kept `isSigned` on `decodeIntStream` because it is still used by the existing geometry/string paths, and I did not want to turn this PR into a broader decoder API refactor.

HarelM: I think it's worth improving the API and not keep it weird.

DoFabien: OK, it’s definitely better, but it will significantly increase the scope of the PR. I will work on it

DoFabien:
In commit 8942a3335ca3885dbc485eb6ec96102ac465a792, I expanded the scope of the original PR to clean up the signed/unsigned split in the decoder API.

The main change is that the integer stream decoder no longer relies on a generic API with an `isSigned` boolean at the call site. Instead, it now exposes explicit signed/unsigned entry points such as `decodeSignedIntStream`, `decodeUnsignedIntStream`, `decodeSignedConstIntStream`, `decodeUnsignedConstIntStream`, and the corresponding long / float64 variants.

This is mostly a readability change rather than a behavior change: the old API worked, but calls like `decodeIntStream(..., false)` were harder to read and easier to misinterpret. With the split API, the intent is visible directly at each call site.

While making that change, I also aligned a few TS types with the semantics that were already present at runtime:

- offsets / indices / geometry type buffers now use unsigned typed arrays where appropriate
- optional topology buffers are typed as `null` when absent, which matches the runtime behavior in `geometryDecoder`
- `vertexOffsets` is also nullable for the same reason

While touching those paths, I also renamed a few local variables to make the code a bit more explicit.

HarelM:
I think it's worthwhile to rename long to int64 in all the places and int to int32.
This way we align all the names and the types related to these names will be easier to read and understand.
What do you think?

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -61,12 +74,12 @@ describe("getVectorType", () => {
     });
 });
 
-describe("decodeIntStream", () => {
+describe("decodeSignedIntStream / decodeUnsignedIntStream", () => {
```
HarelM: Probably better to split those two describes, don't you think?

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -61,12 +74,12 @@ describe("getVectorType", () => {
...
     it("should decode with PhysicalLevelTechnique.NONE", () => {
-        const expectedValues = new Int32Array([10, 20, 30]);
+        const expectedValues = new Uint32Array([10, 20, 30]);
         const metadata = createStreamMetadata(LogicalLevelTechnique.NONE);
-        const data = encodeIntStream(expectedValues, metadata, false);
-        const result = decodeIntStream(data, new IntWrapper(0), metadata, false);
+        const data = encodeIntStream(new Int32Array(expectedValues), metadata, false);
```
HarelM: Let's do the same for `encodeIntStream` -> split it into `encodeUnsignedInt32Stream` and `emcodeUnsignedInt32Stream` to match the decoding names.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -42,16 +42,26 @@ import { VectorType } from "../vector/vectorType";
...
     data: Uint8Array,
     offset: IntWrapper,
     streamMetadata: StreamMetadata,
-    isSigned: boolean,
     scalingData?: GeometryScaling,
     nullabilityBuffer?: BitVector,
 ): Int32Array {
     const values = decodePhysicalLevelTechnique(data, offset, streamMetadata);
-    return decodeInt32(values, streamMetadata, isSigned, scalingData, nullabilityBuffer);
+    return decodeInt32(values, streamMetadata, true, scalingData, nullabilityBuffer);
```
HarelM: Is it worth splitting `decodeInt32` into two methods as well or should we keep it as is?

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -221,6 +272,16 @@ function decodeInt32(
...
 }
 
+function decodeUnsignedInt32(
+    values: Uint32Array,
+    streamMetadata: StreamMetadata,
+    scalingData?: GeometryScaling,
+    nullabilityBuffer?: BitVector,
+): Uint32Array {
+    const decodedValues = decodeInt32(values, streamMetadata, false, scalingData, nullabilityBuffer);
+    return new Uint32Array(decodedValues.buffer, decodedValues.byteOffset, decodedValues.length);
```
HarelM: That's probably a bad design outcome - the entire code path for unsigned should not convert from signed to unsigned, but instead keep the original array.
So I think it's worth splitting `decodeInt32` to avoid this.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -263,7 +324,16 @@ function decodeInt64(
...
 }
 
-export function decodeFloat64(values: Float64Array, streamMetadata: StreamMetadata, isSigned: boolean): Float64Array {
+function decodeUnsignedInt64(
+    values: BigUint64Array,
+    streamMetadata: StreamMetadata,
+    nullabilityBuffer?: BitVector,
+): BigUint64Array {
+    const decodedValues = decodeInt64(values, streamMetadata, false, nullabilityBuffer);
+    return new BigUint64Array(decodedValues.buffer, decodedValues.byteOffset, decodedValues.length);
```
HarelM: Same as above...

ts/src/mltDecoder.ts
```diff
@@ -187,19 +182,13 @@ function decodeIdColumn(
     }
 
     const idDataType = scalarTypeMetadata.longID ? ScalarType.UINT_64 : ScalarType.UINT_32;
+    const nullabilityBuffer = typeof sizeOrNullabilityBuffer === "number" ? undefined : sizeOrNullabilityBuffer;
```
HarelM, line 185: Great! Thanks for this!

ts/src/mltDecoder.spec.ts
```diff
@@ -147,6 +147,12 @@ function compareId(mltFeature: Feature, mvtFeature: VectorTileFeature, idWithinM
             return;
         }
 
+        // TODO(#1075): some OMT fixtures have large MVT IDs that are not safe JS integers,
```
HarelM: Is this related to this PR?

ts/src/vector/geometry/topologyVector.ts
```diff
@@ -1,5 +1,5 @@
 export type TopologyVector = {
-    readonly geometryOffsets: Uint32Array;
-    readonly partOffsets: Uint32Array;
-    readonly ringOffsets: Uint32Array;
+    readonly geometryOffsets: Uint32Array | null;
```
HarelM: I think it should be `| undefined` instead of `| null`, or simply make the property optional using `?`. or am I missing anything? Why was this added? Did typescript complained?

ts/src/vector/geometry/geometryVectorConverter.spec.ts
```diff
@@ -634,11 +634,11 @@ describe("Edge cases", () => {
             GEOMETRY_TYPE.POINT,
             VertexBufferType.VEC_2,
             {
-                geometryOffsets: undefined,
-                partOffsets: undefined,
-                ringOffsets: undefined,
+                geometryOffsets: null,
```
HarelM: Why use null instead of undefined?

ts/src/vector/geometry/geometryVector.ts
```diff
@@ -21,8 +21,8 @@ export abstract class GeometryVector {
     protected constructor(
         private readonly _vertexBufferType: VertexBufferType,
         private readonly _topologyVector: TopologyVector,
-        private readonly _vertexOffsets: Int32Array,
-        private readonly _vertexBuffer: Int32Array,
+        private readonly _vertexOffsets: Uint32Array | null,
```
HarelM: Same here, why use null and not undefined?

ts/src/vector/geometry/flatGeometryVector.ts
```diff
@@ -33,11 +32,11 @@ export function createFlatGeometryVectorMortonEncoded(
 export class FlatGeometryVector extends GeometryVector {
     constructor(
         vertexBufferType: VertexBufferType,
-        //TODO: refactor -> use UInt8Array
-        private readonly _geometryTypes: Int32Array,
+        // TODO: geometry type codes fit in Uint8Array, but the current decoding pipeline still produces Uint32Array.
```
HarelM: I don't see anyone picking up this TODO, better to simply remove it.

HarelM:
I've added some minor comments, overall, this looks great now.
The split between the signed and unsigned is a lot more readable now.

ts/src/vector/geometry/geometryVectorConverter.spec.ts
DoFabien: I used `null` initially because the decoder was already using `null` for missing geometry buffers, so I aligned the types with the existing runtime behavior. But I agree `undefined` / optional is cleaner here, so I switche it in the follow-up patch.

ts/src/mltDecoder.spec.ts
DoFabien: It is not really caused by this PR, but this PR makes the existing issue visible.

Before the unsigned fix, some `uint32` IDs were still flowing through a signed path on the TS side, so large unsigned values could appear as negative values and get skipped by the existing `mltFeatureId < 0` guard in the test.

After this PR, the TS decoder now preserves the correct unsigned value, so the comparison reaches the MVT-side value, where large IDs are already problematic / unreliable. So this is a pre-existing separate issue that became visible once the unsigned decoding was fixed.

DoFabien:
> I think it's worthwhile to rename long to int64 in all the places and int to int32. This way we align all the names and the types related to these names will be easier to read and understand. What do you think?

Yes, I think is definitly better.

But, could you clarify the scope you want for this PR?

- 1: only the stream encoder/decoder APIs, ( and their internal helpers, and the related tests)
- 2: option 1 + the TS vector classes/types/files such as `Int*` / `Long*`
- 3: a broader TS-wide rename of all `int` / `long` names

HarelM:
I'm good with 3, but other options are fine too. Renames are mostly easy.
We can do it in a different PR though to keep the scope of this one to only signed/unsigned...

HarelM:
I'll let you decide on the scope of this PR, I'm ok with however you prefer to scope it.
Regardless of the scope, in order to merge this there's a need to address the current review comments.

DoFabien:
Yes, of course. I’ve almost finished. I went with option 3, the largest scope. I’ll push the updates later today.

DoFabien:
Hi, 

The first new 2 commits address review comments:
  - split the signed and unsigned integer stream paths so the unsigned path no longer goes through signed decoding and reinterpretation;
  - clean up optional geometry buffer handling by consistently using `undefined` instead of `null`.

The last commit is the rename pass from `int` / `long` to `int32` / `int64`.

This rename goes a bit beyond pure search/replace in a few places:

- `IntVector` became `IdVector`, because that type is really the ID-column union and can hold `Int32`, `Int64`, or `Float64`-materialized IDs, so `IntVector` was misleading.
- `longSequenceVektor.spec.ts` was renamed to `int64SequenceVector.spec.ts`, which also fixes the old `Vektor` typo.
- `IntWrapper` was intentionally left unchanged. It is a pre-existing mutable offset helper, not an int32-specific domain type.
- `integerStreamDecoder.ts` / `integerStreamEncoder.ts` kept their filenames, since they still host both `int32` and `int64` stream logic.

ts/src/vector/featureTable.ts
```diff
@@ -89,11 +89,11 @@ export default class FeatureTable {
...
     }
 
-    private containsMaxSaveIntegerValues(intVector: IntVector) {
+    private containsMaxSafeIntegerValues(idVector: IdVector) {
         return (
-            intVector instanceof IntFlatVector ||
-            (intVector instanceof IntConstVector && intVector instanceof IntSequenceVector) ||
-            intVector instanceof DoubleFlatVector
+            idVector instanceof Int32FlatVector ||
+            (idVector instanceof Int32ConstVector && idVector instanceof Int32SequenceVector) ||
```
HarelM: I know this is only a rename, but does this condition makes sense? can a vector be both `Int32ConstVector` and `Int32SequenceVector`?

ts/src/vector/geometry/geometryVector.ts
```diff
@@ -34,11 +34,11 @@ export abstract class GeometryVector {
         return this._topologyVector;
     }
 
-    get vertexOffsets(): Int32Array {
+    get vertexOffsets(): Uint32Array | undefined {
```
HarelM, line 37: Is `| undefined` needed here? Does typescript complains?

ts/src/mltDecoder.ts
```diff
@@ -187,63 +182,51 @@ function decodeIdColumn(
...
                 id[0],
                 id[1],
                 (idDataStreamMetadata as RleEncodedStreamMetadata).numRleValues,
             );
         }
         case VectorType.CONST: {
-            const id = decodeConstLongStream(tile, offset, idDataStreamMetadata, false);
-            return new LongConstVector(columnName, id, sizeOrNullabilityBuffer, true);
+            const id = decodeUnsignedConstInt64Stream(tile, offset, idDataStreamMetadata);
+            return new Int64ConstVector(columnName, id, sizeOrNullabilityBuffer, false);
```
HarelM, line 229: This was previously `true` for `isSigned` I think, are you sure about this change?

ts/src/encoding/integerStreamEncoder.ts
```diff
@@ -89,11 +93,48 @@ function encodeInt32(
...
+    }
+}
+
+function encodeUnsignedInt32(
+    values: Uint32Array,
+    streamMetadata: StreamMetadata,
+    bitVector?: BitVector,
+    scalingData?: GeometryScaling,
+): { data: Uint32Array; runs?: number } {
+    values = bitVector ? packNullable(values, bitVector) : new Uint32Array(values);
```
HarelM, line 111: nit picking: Is there a need to create a copy here in case bitVector is not provided?

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -78,52 +92,54 @@ describe("decodeIntStream", () => {
...
+        const data = encodeUnsignedInt32Stream(expectedValues, metadata, bitVector);
 
-        const result = decodeIntStream(data, new IntWrapper(0), metadata, false, undefined, bitVector);
+        const result = decodeUnsignedInt32Stream(data, new IntWrapper(0), metadata, undefined, bitVector);
 
         expect(result).toEqual(expectedValues);
     });
+});
 
+describe("decodeSignedInt32Stream / decodeSignedConstInt32Stream", () => {
```
HarelM: Split describes if it's important enough to highlight it in this describe.

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -192,64 +208,87 @@ describe("decodeIntStream", () => {
...
+        const metadata = createStreamMetadata(LogicalLevelTechnique.NONE, LogicalLevelTechnique.NONE, 1);
+        const data = encodeSignedInt32Stream(new Int32Array([-8]), metadata);
+
+        const result = decodeSignedConstInt32Stream(data, new IntWrapper(0), metadata);
+
+        expect(result).toBe(-8);
+    });
+});
+
+describe("decodeUnsignedInt32Stream / decodeUnsignedConstInt32Stream", () => {
```
HarelM: Split describes if it's important enough to highlight it in this describe.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -203,12 +253,63 @@ function decodeInt32(
...
+    streamMetadata: StreamMetadata,
+    scalingData?: GeometryScaling,
+    nullabilityBuffer?: BitVector,
+): Uint32Array {
+    let decodedValues: Uint32Array;
+    switch (streamMetadata.logicalLevelTechnique1) {
+        case LogicalLevelTechnique.DELTA:
+            if (streamMetadata.logicalLevelTechnique2 === LogicalLevelTechnique.RLE) {
+                const rleMetadata = streamMetadata as RleEncodedStreamMetadata;
+                const deltaValues = decodeUnsignedRleInt32(values, rleMetadata.runs, rleMetadata.numRleValues);
```
HarelM, line 289: Is there a missing `if (!nullabilityBuffer) {` here? I think it was this was when both singed and unsigned used the same code...?

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -242,15 +342,49 @@ function decodeInt64(
...
+    values: BigUint64Array,
+    streamMetadata: StreamMetadata,
+    nullabilityBuffer?: BitVector,
+): BigUint64Array {
+    let decodedValues: BigUint64Array;
+    switch (streamMetadata.logicalLevelTechnique1) {
+        case LogicalLevelTechnique.DELTA:
+            if (streamMetadata.logicalLevelTechnique2 === LogicalLevelTechnique.RLE) {
+                const rleMetadata = streamMetadata as RleEncodedStreamMetadata;
+                const deltaValues = decodeUnsignedRleInt64(values, rleMetadata.runs, rleMetadata.numRleValues);
```
HarelM, line 384: Same as above with the nullability vector...?

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -379,20 +513,12 @@ export function getVectorType(
     return streamMetadata.numValues === 1 ? VectorType.CONST : VectorType.FLAT;
 }
 
-function decodeRleInt32(data: Uint32Array, streamMetadata: RleEncodedStreamMetadata, isSigned: boolean): Int32Array {
-    return isSigned
-        ? decodeZigZagRleInt32(data, streamMetadata.runs, streamMetadata.numRleValues)
-        : new Int32Array(decodeUnsignedRleInt32(data, streamMetadata.runs, streamMetadata.numRleValues));
+function decodeSignedRleInt32(data: Uint32Array, streamMetadata: RleEncodedStreamMetadata): Int32Array {
```
HarelM: If this is only a one liner method I would consider removing it.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -379,20 +513,12 @@ export function getVectorType(
...
-function decodeRleInt64(
-    data: BigUint64Array,
-    streamMetadata: RleEncodedStreamMetadata,
-    isSigned: boolean,
-): BigInt64Array {
-    return isSigned
-        ? decodeZigZagRleInt64(data, streamMetadata.runs, streamMetadata.numRleValues)
-        : new BigInt64Array(decodeUnsignedRleInt64(data, streamMetadata.runs, streamMetadata.numRleValues));
+function decodeSignedRleInt64(data: BigUint64Array, streamMetadata: RleEncodedStreamMetadata): BigInt64Array {
+    return decodeZigZagRleInt64(data, streamMetadata.runs, streamMetadata.numRleValues);
```
HarelM: If this is only a one liner method I would consider removing it.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -404,3 +530,46 @@ function decodeRleFloat64(
         ? decodeZigZagRleFloat64(data, streamMetadata.runs, streamMetadata.numRleValues)
         : decodeUnsignedRleFloat64(data, streamMetadata.runs, streamMetadata.numRleValues);
 }
+
+function decodeUnsignedZigZagDeltaInt32(data: Uint32Array): Uint32Array {
```
HarelM: The methods below should probably be in `integerDecodingUtils` instead of in this file.
`decodeUnsignedZigZagDeltaInt32`
`decodeUnsignedZigZagDeltaInt64`
`decodeUnsignedComponentwiseDeltaVec2`
`decodeUnsignedComponentwiseDeltaVec2Scaled`

HarelM:
I've reviewed the files, overall looks great, thanks, I've added a few comments.
If you can link to the relevant commit as a reply to a comment that I added it would help me make sure all code review comments were addressed.
Thanks!

ts/src/vector/featureTable.ts
DoFabien: You're right, this condition does not make sense.

An `idVector` cannot be both `Int32ConstVector` and `Int32SequenceVector`, so this should not be `&&` here. This was not intentional in the rename.

I fixed it in 4d9a6436.

ts/src/vector/geometry/geometryVector.ts
DoFabien: Yes, this one is intentional.

`vertexOffsets` is genuinely absent on some geometry decoding paths, so the `| undefined` is meant to reflect the actual runtime contract, not just satisfy TypeScript.

ts/src/mltDecoder.ts
DoFabien: Yes, I’m pretty sure.

The hunk is a bit misleading because GitHub is showing the diff against the PR base. The `true -> false` change did not come from the `int/long -> int32/int64` rename itself.

Just before the rename, this `UINT_64` const ID path was already using the unsigned decoder and `isSigned = false`, and the current code keeps that same behavior.

ts/src/encoding/integerStreamEncoder.ts
DoFabien: Yes, this copy is intentional.
Some encoding paths mutate the working buffer in place, so keeping a defensive copy here avoids mutating the caller-owned `Uint32Array` when no `bitVector` is provided.
This mirrors the signed path as well.

ts/src/decoding/integerStreamDecoder.spec.ts
DoFabien: Agreed, this is better. I split them in 827770dd.

ts/src/decoding/integerStreamDecoder.spec.ts
DoFabien: I split the unsigned `describe`s as well in 827770dd

ts/src/decoding/integerStreamDecoder.ts
DoFabien: Yes, you were right to call this out.
The unsigned path was still functionally correct without that branch, but it had lost the previous non-nullable fast path when the signed/unsigned code was split.

I restored that fast path in 45c505bd.

ts/src/decoding/integerStreamDecoder.ts
DoFabien: Yes, same here. Restored the fast path in 45c505bd.

ts/src/decoding/integerStreamDecoder.ts
DoFabien: Yes, I removed it and inlined the call in ad16b2db

ts/src/decoding/integerStreamDecoder.ts
DoFabien: I inlined it too in ad16b2db

ts/src/decoding/integerStreamDecoder.ts
DoFabien: Agreed. These are decoding primitives, so `integerDecodingUtils` is a better place for them.
I moved them there in f2e36ddb.

ts/src/decoding/integerStreamDecoder.ts
HarelM: Sorry, I didn't understood that it's the same functionality.
If the functionality is the same then your new version is better than the old one.
I don't want to do premature optimizations when we don't even know how much time it would save.
Let's keep the more elegant version and not the faster version, I like readability better when I don't know what the performance gain is.

HarelM:
Amazing work! Thanks!
Added one last comment about better looking code you created which I would like to keep.

ts/src/decoding/integerStreamDecoder.ts
DoFabien: I benchmarked this again on the real `omt` fixtures with `perf stat`

Comparing 45c505bd (with the unsigned `DELTA + RLE` fast path) vs 634cb16c (simpler version without it):

|  | without fastpath | with fastpath| Delta |
| --- | ---: | ---: | ---: |
| Median decode time | `3880.01 ms` | `3922.93 ms` | `+1.11%` |
| `task-clock` | `5,944,013,413` | `6,114,580,738` | `+2.87%` |
| `cycles` | `34,259,187,933` | `33,837,089,781` | `-1.23%` |
| `instructions` | `84,868,269,670` | `84,522,462,780` | `-0.41%` |
| `branches` | `20,219,049,056` | `20,122,379,525` | `-0.48%` |
| `branch-misses` | `207,177,057` | `208,319,638` | `+0.55%` |

So the fast path still did not improve end-to-end decoding time. It slightly reduced raw CPU work, but not enough to justify the extra branching/complexity here, so I kept the simpler version.

PR #1081, opened 2026-03-09
refactor(rust): noop, only renames for future refactoring

nyurik:


rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -72,7 +72,9 @@ impl OwnedProperty {
 
     #[must_use]
     pub fn approx_type(&self) -> AproxPropertyType {
-        use {AproxPropertyType as T, OwnedEncodedPropValue as Enc, PropValue as Dec};
+        use AproxPropertyType as T;
+        use OwnedEncodedPropValue as Enc;
+        use PropValue as Dec;
```
CommanderStorm, line 77: weird that we keep flip-flopping on this formatting.
Different compiler version?

nyurik: hmm, weird - I used `just fmt` - and i have recently ran `rustup update`.  The fmt used here IS nightly

PR #1086, opened 2026-03-09
refacor(rust): optimise geometry type decoding

CommanderStorm:


nyurik (approved):

CommanderStorm:
Byte RLE needs a deeper look at... it is a bit weird

PR #1087, opened 2026-03-09
chore(rust): refactor name storage

nyurik:


rust/mlt-core/src/layer/v01/property/strings.rs
```diff
@@ -203,24 +219,27 @@ impl<'a> EncodedStrings<'a> {
     #[must_use]
     pub fn streams(&self) -> Vec<&Stream<'_>> {
         match self {
-            Self::Plain { lengths, data } => vec![lengths, data],
+            Self::Plain { lengths, data, .. } => vec![lengths, data],
             Self::Dictionary {
                 lengths,
                 offsets,
                 data,
+                ..
```
CommanderStorm: nit: I would prefer `name: _` since that raises a compilation warning if another struct field is added.

CommanderStorm (approved):

rust/mlt-core/src/layer/v01/property/strings.rs
nyurik: fixed

PR #1088, opened 2026-03-09
chore(rust): refactoring step 3

nyurik:


rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -134,30 +135,81 @@ impl Analyze for EncodedProperty<'_> {
...
-            OwnedEncodedPropValue::F32(Some(_), _) => ColumnType::OptF32.write_to(writer)?,
-            OwnedEncodedPropValue::F32(None, _) => ColumnType::F32.write_to(writer)?,
-            OwnedEncodedPropValue::F64(Some(_), _) => ColumnType::OptF64.write_to(writer)?,
-            OwnedEncodedPropValue::F64(None, _) => ColumnType::F64.write_to(writer)?,
-            OwnedEncodedPropValue::Str(Some(_), _) => ColumnType::OptStr.write_to(writer)?,
-            OwnedEncodedPropValue::Str(None, _) => ColumnType::Str.write_to(writer)?,
-            OwnedEncodedPropValue::SharedDict(_) => ColumnType::SharedDict.write_to(writer)?,
-        }
+        use OwnedEncodedPropValue as Val;
+        let col_type = match &self.value {
```
CommanderStorm, line 139: Lets move this to a separate method (with one match for `presence.is_some` and one without).
This is kind of hard to read this way

rust/mlt-core/src/layer/v01/property/strings.rs
```diff
@@ -215,73 +242,135 @@ impl<'a> EncodedStrings<'a> {
...
+            | Self::Dictionary { presence, .. }
+            | Self::FsstPlain { presence, .. }
+            | Self::FsstDictionary { presence, .. } => presence.as_ref(),
+        }
+    }
+
+    /// Streams in wire order: presence (if any) then content streams.
     #[must_use]
     pub fn streams(&self) -> Vec<&Stream<'_>> {
+        let mut v: Vec<&Stream<'_>> = self.presence().into_iter().collect();
```
CommanderStorm, line 259: Why not inline this below?

Below you are always ignoring `presence`..

rust/mlt-core/src/layer/v01/property/mod.rs
CommanderStorm: now I see why you did it this way

rust/mlt-core/src/layer/v01/property/strings.rs
CommanderStorm: now I see..

PR #1094, opened 2026-03-09
refactor(rust): move all structs to structs.rs

nyurik:
consolidating all type declarations - makes it much easier to reason about


CommanderStorm:
Having enums in structs.re us a bit weird, but I don't mind. It is internal API at best

PR #1096, opened 2026-03-10
fix(java): Improve PMTiles directory handling in Java CLI

TimSylvester:
- Filter zoom levels as part of directory reading
- Stop exploring the directory hierarchy once the max level is explored
- Stop using Planetiler base class, it wasn't adding anything
- Separate io.Tileverse dependencies from Planetiler-based PMTiles reader

This improves the time for converting small tilesets, avoiding the need to read most of the, e.g., ~200MB tile directories in a z14 archive.

### DIFF
--- a/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/ReadablePmtiles.kt
+++ b/java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/ReadablePmtiles.kt
@@ -32,21 +32,17 @@ class ReadablePmtiles(
                 throw IllegalArgumentException("ByteRange length must be positive")
             }
         }
-        override fun equals(other: Any?): Boolean {
-            if (this === other) return true
-            if (javaClass != other?.javaClass) return false
-            other as ByteRange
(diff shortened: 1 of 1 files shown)


java/mlt-cli/src/test/kotlin/org/maplibre/mlt/cli/ReadablePmtilesMapDirectoryTest.kt
```diff
@@ -43,6 +44,20 @@ class ReadablePmtilesMapDirectoryTest {
         assertTrue(result.toList().isEmpty())
     }
 
+    @Test
+    fun `directory reference entry does not call recurse when tileId not less than endTileIndex`() {
```
CommanderStorm, line 48: strings as function names works?? WTF?

TimSylvester: Yeah, that's how Copilot generated tests, so I went with it.

PR #1099, opened 2026-03-10
chore(rust): rework our internal data model

nyurik:


rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -60,178 +61,214 @@ impl OwnedProperty {
...
-                | Dec::U64(..) => T::Integer,
-                Dec::F32(..) | Dec::F64(..) => T::Float,
-                Dec::Str(..) => T::String,
-                Dec::SharedDict(_) => T::SharedDict,
-            },
+                | Enc::U64(..),
+            ) => T::Integer,
+            Self::Encoded(Enc::F32(..) | Enc::F64(..)) => T::Float,
+            Self::Encoded(Enc::Str(..)) => T::String,
+            Self::Encoded(Enc::SharedDict(..)) => T::SharedDict,
```
CommanderStorm: nit: Likely impl `approx_type` for `OwnedEncodedProperty` would be better code.

rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -256,163 +293,175 @@ impl OwnedEncodedProperty {
...
-            Self::F32(v) => v.iter().map(Option::is_some).collect(),
-            Self::F64(v) => v.iter().map(Option::is_some).collect(),
-            Self::Str(v) => v.iter().map(Option::is_some).collect(),
-            Self::SharedDict(_) => Err(NotImplemented("presence stream for shared dict"))?,
-        })
+impl DecodedPresence {
+    #[must_use]
+    pub fn bools(&self, non_null_count: usize) -> Vec<bool> {
+        self.0.clone().unwrap_or_else(|| vec![true; non_null_count])
     }
```
CommanderStorm, line 333: Seems a bit inefficent to alloc this even if we don't need to, but 🤷🏻
Would need to benchmark if the alloc vs less branchy code is more siginficant

rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -425,8 +474,175 @@ impl PropValue {
...
 }
 
-impl_decodable!(Property<'a>, EncodedProperty<'a>, DecodedProperty);
-impl_encodable!(OwnedProperty, DecodedProperty, OwnedEncodedProperty);
+impl DecodedProperty<'static> {
```
CommanderStorm, line 481: `'static`?
Why is that nesseary?

rust/mlt-core/src/layer/v01/property/mod.rs
```diff
@@ -425,8 +474,175 @@ impl PropValue {
...
+            PropValue::SharedDict(shared_dict, items) => Self::SharedDict(name, shared_dict, items),
+        }
+    }
+}
+
+impl Default for DecodedProperty<'_> {
+    fn default() -> Self {
+        Self::Bool(String::new(), Vec::new())
+    }
+}
```
CommanderStorm: unless we actually need it, this seems a bit of a footgun


```suggestion
```

nyurik: yep, i also wanted to get rid of it but forgot

rust/mlt-core/src/layer/v01/property/mod.rs
nyurik: the 'static here allows us not to have an Owned variant at all because the prop has a Cow, which means when we generate fully-owned variant, we can simply use 'static path and inside it will all be owned with the Cow

PR #1100, opened 2026-03-10
refactor(rust): rename structs to model

nyurik:


CommanderStorm:
makes the diff smaller

PR #1108, opened 2026-03-10
chore(rust): move name into DecodedStrings

nyurik:
also renamed ApproxPropertyType into PropertyKind - seems much cleaner


CommanderStorm:
merging to reduce the chanvce of having merge conflicts

PR #1116, opened 2026-03-11
feat(java): Detect and avoid re-converting duplicate input tiles

TimSylvester:
PMTiles combines duplicate tiles into a run-length range when they're sequential (in Hilbert order), but when they're not there's no easy way to detect them.  Here we track all the input tiles by their byte offset in the source file and skip converting any that we have already handled, submitting the same hash as the previous result which causes `WritablePmtiles` to write a reference to that previous result's location.

Adds more options to control and evaluate the PMTiles reader cache.

Adds the option to display cache statistics for tuning cache parameters.

Refactors command-line and environment handling, adds tests.


java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/EncodeCommandLine.kt
```diff
@@ -820,3 +793,56 @@ Add an explicit column mapping on the specified layers:
 
     private val logger: Logger = LoggerFactory.getLogger(EncodeCommandLine::class.java)
 }
+
+val CommandLine.minZoom get() = getParsedOptionValue<Long>(EncodeCommandLine.MIN_ZOOM_OPTION, 0L).toInt()
```
CommanderStorm, line 788: This is a bit of a weird way to write this since it is reaching into a foreign thing..

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/Environment.kt
```diff
@@ -40,6 +40,10 @@ const val ENV_CACHE_BLOCK_SIZE = "MLT_CACHE_BLOCK_SIZE"
 const val DEFAULT_CACHE_BLOCK_SIZE = 16 * 1024
 val cacheBlockSize by lazy { computeCacheBlockSize() }
 
+const val ENV_MAX_TILE_TRACK_SIZE = "MLT_MAX_TILE_TRACK_SIZE"
+const val DEFAULT_MAX_TILE_TRACK_SIZE = 1024
```
CommanderStorm, line 44: 1KB seems a bit small, but you likely have a reason for setting it to this size

CommanderStorm (approved):

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/EncodeCommandLine.kt
TimSylvester: The built-in numeric converter only returns `Double` or `Long`, so `getParsedOptionValue<Int>()` always fails.  To get rid of this little oddity, I'd have to configure the CLI option with a custom converter.

java/mlt-cli/src/main/kotlin/org/maplibre/mlt/cli/Environment.kt
TimSylvester: It's mostly just "empty" tiles that are repeated, one water or land quad.  The largest repeated tile I observed in the first few million tiles of the example I'm using was 416 bytes.    Even if there are a few larger tiles repeated, it's probably not worth tracking millions more smaller tiles to catch them.  I'll do a more complete run to check, though.

TimSylvester: This is the distribution of repeated tiles in the OMT data I'm using (log/log).

Top item is 3,948,453 repeats of 78 bytes, out of 5,995,170 total repeats.  They don't show up at this scale, but there's exactly one repeat at each of 672, 810, and 1258 bytes.

For this dataset, the threshold should be a little smaller, but I think it's a reasonable default.  The effect on memory use is not huge, so I think it's usually better to over-estimate.

image

PR #1128, opened 2026-03-12
chore: remove ALP everywhere - it was never implemented

nyurik:
It is pointless to have ALP as panicking code everywhere, so for now lets remove it fully everywhere to keep less code


cpp/include/mlt/metadata/stream.hpp
```diff
@@ -39,9 +39,7 @@ enum class PhysicalLevelTechnique : std::uint32_t {
     /// Can produce better results in combination with a heavyweight compression scheme like Gzip.
     /// Simple compression scheme where the decoder are easier to implement compared to FastPfor.
     VARINT = 2,
-    /// Adaptive Lossless floating-Point Compression
-    ALP = 3,
-    VALUE_COUNT = 4,
+    VALUE_COUNT = 3,
```
CommanderStorm, line 42: was `VALUE_COUNT` implemented?
If yes, this would be a pretty harsh breaking change.

Would it not be simpler to just keep it?
I mean we kind of want to impl it..

CommanderStorm (approved):
I think @TimSylvester wanted to move towards implmenenting this on the CPP side.

Not sure how prevalent floats are in real world datasets.
Seems that they appear in overture..

I am fine with moving this to V2

cpp/include/mlt/metadata/stream.hpp
nyurik: no, this is a standard way for C++ to implement bounds checking :)

CommanderStorm: Uff..

CommanderStorm:
So, I tried to impl ALP into the rust encoder and I noticed that it does not cleanly work.
Alp is a logical encoding, not physical. At least somtimes.
We should re-introduce it as such instead.

PR #1134, opened 2026-03-12
chore(rust): replace FromEncoded with Decode and decode_into

nyurik:


CommanderStorm:
merging to prevent more merge conflicts

PR #1136, opened 2026-03-12
fix(ts): support PhysicalLevelTechnique.NONE for integer streams

DoFabien:
This PR fixes TypeScript support for `PhysicalLevelTechnique.NONE` on integer streams.

The previous implementation did not read or write integer payloads as explicit little-endian words. In practice:

- the 32-bit `NONE` path decoded bytes incorrectly
- the 32-bit encoder did not emit plain little-endian integer payloads correctly
- the 64-bit integer stream decode paths did not support `NONE`

## Changes

- decode integer `NONE` payloads as little-endian integer words
- encode integer `NONE` payloads as little-endian integer words
- add `NONE` support to the TypeScript `int64` integer stream decode paths
- split the physical integer decode path into explicit `Int32` / `Int64` helpers
- update tests so they actually exercise `physicalLevelTechnique: NONE`

## Note

The repository is currently inconsistent on this topic:

- `docs/specification.md` and `docs/encodings.md` do not describe the same story
- Rust / C++ / Java do not currently behave the same way for integer `NONE`

I will open a separate issue in the next few hours to clarify the intended project-wide direction for the spec and the other implementations.

### DIFF
--- a/rust/mlt-core/src/decode.rs
+++ b/rust/mlt-core/src/decode.rs
@@ -1,15 +1,32 @@
 use crate::MltError;
 
-/// Trait for types that can be constructed from encoded data
-pub(crate) trait FromEncoded<'a>: Sized {
-    type Input: 'a;
-    fn from_encoded(input: Self::Input) -> Result<Self, MltError>;
+/// Decoding counterpart to [`TryFrom`], used as a trait bound on [`Decodable::DecodedType`].
+///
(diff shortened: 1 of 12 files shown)


DoFabien:
Superseded by #1138.

I accidentally force-pushed my fork's `main`, so this PR lost its diff and was closed. Reopened properly in #1138.

PR #1138, opened 2026-03-12
fix(ts): support PhysicalLevelTechnique.NONE for integer streams

DoFabien:
> replace #1136 

This PR fixes TypeScript support for `PhysicalLevelTechnique.NONE` on integer streams.

The previous implementation did not read or write integer payloads as explicit little-endian words. In practice:

- the 32-bit `NONE` path decoded bytes incorrectly
- the 32-bit encoder did not emit plain little-endian integer payloads correctly
- the 64-bit integer stream decode paths did not support `NONE`

## Changes

- decode integer `NONE` payloads as little-endian integer words
- encode integer `NONE` payloads as little-endian integer words
- add `NONE` support to the TypeScript `int64` integer stream decode paths
- split the physical integer decode path into explicit `Int32` / `Int64` helpers
- update tests so they actually exercise `physicalLevelTechnique: NONE`

## Note

The repository is currently inconsistent on this topic:

- `docs/specification.md` and `docs/encodings.md` do not describe the same story
- Rust / C++ / Java do not currently behave the same way for integer `NONE`

I will open a separate issue in the next few hours to clarify the intended project-wide direction for the spec and the other implementations.


ts/src/encoding/integerStreamEncoder.ts
```diff
@@ -228,8 +231,8 @@ export function encodeInt64SignedDeltaRle(runs: Array<[number, bigint]>): Uint8A
 }
 
 /**
- * Encodes unsigned BigInt64 values with varint compression (no zigzag)
+ * Encodes unsigned BigInt64 values as plain little-endian uint64 words.
```
HarelM: Probably better to remove single line methods as it doesn't add a lot of value I believe...

ts/src/encoding/encodingUtils.ts
```diff
@@ -20,6 +20,14 @@ export function encodeDoubleLE(values: Float64Array): Uint8Array {
     return buffer;
 }
 
+export function encodeUint32sLE(values: Uint32Array): Uint8Array {
```
HarelM, line 23: Is this a simple copy of the data? If so maybe it's better to rename the method as `copyArray`, or simply remove it and use the content of the method when needed.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -513,8 +525,8 @@ export function getVectorType(
     if (streamMetadata.physicalLevelTechnique === PhysicalLevelTechnique.VARINT) {
         values = new Int32Array(decodeVarintInt32(data, offset, 4));
     } else {
-        const byteOffset = offset.get();
-        values = new Int32Array(data.buffer, data.byteOffset + byteOffset, 4);
+        const decoded = decodeUint32sLE(data, offset, 4);
+        values = new Int32Array(decoded.buffer, decoded.byteOffset, decoded.length);
```
HarelM: Can this be written like in the previous if statement as `values = new Int32Array(decodeUint32sLE(data, offset, 4));`?
Or maybe even something like:
```ts
let values: Int32Array = streamMetadata.physicalLevelTechnique === PhysicalLevelTechnique.VARINT 
  ? decodeVarintInt32(data, offset, 4) 
  : decodeUint32sLE(data, offset, 4);
```

ts/src/encoding/encodingUtils.ts
```diff
@@ -20,6 +20,14 @@ export function encodeDoubleLE(values: Float64Array): Uint8Array {
...
+    return new Uint8Array(values.slice().buffer);
+}
+
+export function encodeUint64sLE(values: BigUint64Array): Uint8Array {
+    return new Uint8Array(values.slice().buffer);
```
HarelM: Same as above.

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -664,11 +706,29 @@ describe("decodeInt64Stream", () => {
 
             expect(result).toEqual(expectedValues);
         });
+
+        it("should decode little-endian signed Int64 words with PhysicalLevelTechnique.NONE", () => {
```
HarelM: This test is not clear how the data is transformed from `0x0f, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00` to `-8n`.
I advise to add an encoderMethod to bridge this gap.

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -664,11 +706,29 @@ describe("decodeInt64Stream", () => {
...
+            const data = new Uint8Array([0x0f, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]);
+
+            const result = decodeSignedConstInt64Stream(data, new IntWrapper(0), metadata);
+
+            expect(result).toBe(-8n);
+        });
+
+        it("should decode little-endian unsigned Int64 words with PhysicalLevelTechnique.NONE", () => {
+            const metadata = createInt64PhysicalNoneMetadata(1);
+            const data = new Uint8Array([0x88, 0x77, 0x66, 0x55, 0x44, 0x33, 0x22, 0x11]);
```
HarelM: Same as above.

ts/src/decoding/decodingUtils.spec.ts
```diff
@@ -36,6 +46,54 @@ describe("decodingUtils", () => {
...
+            const encoded = encodeUint64sLE(data);
+            const offset = new IntWrapper(0);
+
+            const result = decodeUint64sLE(encoded, offset, 2);
+
+            expect(result).toEqual(data);
+            expect(offset.get()).toBe(16);
+        });
+
+        it("should decode uint64 values from little-endian bytes", () => {
```
HarelM: Same here.

ts/src/decoding/decodingUtils.ts
```diff
@@ -98,6 +98,58 @@ export function decodeDoublesLE(
     return fb;
 }
 
+export function decodeUint32sLE(
```
HarelM: Other methods in this file has support for `nullabilityBuffer`, do the new methods require this support as well?

Also other methods are not throwing exceptions, so I think it's better to keep it without exception handling at this level.

HarelM:
Might be worth waiting for a decision about the issue you opened before continuing with the implementation.
In any case, I've added some comments, mostly about expected vs input values in the tests - the more straight forwards this is the better we will be able to review and make sure the tests are working as expected.

HarelM:
Thanks again for all the hard work on this!

DoFabien:
Thanks for the review. I came back to this branch now that Rust PR #1466 has been merged. The Rust encoder can now choose `PhysicalLevelTechnique.NONE` for singleton integer streams when the plain representation is smaller than VarInt, so TypeScript needs to understand those streams as well.

I merged the latest `main` into this branch and addressed the relevant review comments. The main change is that signed and unsigned 32-bit and 64-bit plain integer payloads are now read and written explicitly as little-endian. I also fixed the optimized path that materializes UInt64 IDs within JavaScript's safe integer range as numbers, which was still assuming VarInt, and made sure the new little-endian readers cannot read outside the supplied `Uint8Array` view. The tests now use readable values and assert the actual bytes.

I also checked the actual wire bytes in both directions for singleton `u32`, `i32`, `u64`, and `i64` streams. Streams produced by Rust decode correctly in TypeScript, and TypeScript produces the same bytes as Rust.

CommanderStorm (approved):
Thank you

HarelM (approved):

HarelM:
If you have a synthetic tile that can be added to the synthetic tests that might be a good idea to add it here.

Issue #1124, opened 2026-03-11
MLT pmtiles archive created with planetiler from Overture data not usable in MapLibre 5.20.0


neodescis:
I have created a small tile set using planetiler [v0.10.0](https://github.com/onthegomap/planetiler/releases/tag/v0.10.0), and I am seeing issues with the tiles. I originally filed an [issue](https://github.com/onthegomap/planetiler/issues/1491) with planetiler, but @msbarry investigated and believes the issue lies with the TypeScript tile decoder, so I'm filing an issue here. I've created a repo with the pmtiles archive produced, and a test harness using MapLibre 5.20.0 and PMTiles 4.4.0:  https://github.com/neodescis/overture-planetiler-profile

In short, when trying to load the tiles in MapLibre, I'm seeing tile parse errors:

1. Invalid typed array length: -125305058
2. FastPFOR: invalid encodedByteLength=46 at offset=9040...
3. Invalid typed array length: -32555056
4. Specified physicalLevelTechnique is not supported (yet)

Note that I created the archive without FastPFOR or pre-computed triangulations. Also, @msbarry said he set up a test harness to decode the tiles with the Java decoder and the tiles are decoding fine there.

I'm also seeing the error "Cannot read properties of null (reading 'slice')" for tiles that do not exist in the archive, which I've traced to this line in MapLibre GL:  https://github.com/maplibre/maplibre-gl-js/blob/e9478ae76328e7c1606d649d84b56e3c1b9ca148/src/source/vector_tile_worker_source.ts#L137. The `rawData` coming back from the tile worker is null here, and thus `.slice()` is crashing. This does not happen with MVT tiles though, so maybe the TypeScript MLT decoder is producing null instead of an empty array in this case?

HarelM:
Thanks for taking the time to report this issue! 
Can you extract a single tile that fails decoding? There's a mltDecoder.spec.ts that should allow you to reproduce the issue here. 
You can check version 5.19 with the previous version of mlt and see if this is a regression issue or something that never worked.

neodescis:
The archive did not work in 5.19.0 either, though I was seeing different errors, probably because FastPFOR wasn't there yet. I'll see if I can get a tile or two extracted out and reproduce the issue directly.

HarelM:
Cool, thanks, let me know if and when you need some help.

neodescis:
[...]

PR #1146, opened 2026-03-13
perf(ts): optimize and harden metadata decoding

DoFabien:
This PR focuses on the **TS metadata decode path**, which had clear room for improvement in both robustness and performance.

It does two things in the same place:

1. **Makes it faster** on the hot path
2. **Makes it fail earlier and more clearly** on truncated or corrupted inputs

Before this change, broken metadata could desynchronize offsets and then fail later with misleading errors such as:

```
FastPFOR: invalid encodedByteLength=...
Invalid typed array length: ...
```

### Hot-path optimization

- `streamMetadataDecoder` now uses **precomputed lookup tables** and **cached `LogicalStreamType` instances** instead of rebuilding values on the hot path
- Metadata varints are now decoded with a **scalar helper** (`decodeVarintInt32Value(...)`) without intermediate array allocation
- The same scalar varint helper is also used for nearby metadata/header reads in:
  - `mltDecoder.ts`
  - `embeddedTilesetMetadataDecoder.ts`
  - `stringDecoder.ts`

### Input validation & error reporting

- Malformed 5-byte varint32 values are now **rejected explicitly** instead of being partially accepted
- `decodeStreamMetadata(...)` is now **atomic on failure**: if metadata decode fails, the caller offset is reset
- Invalid logical stream subtypes for `DATA`, `OFFSET`, and `LENGTH` now **fail explicitly** instead of silently decoding as an empty/default logical stream type
- Embedded tileset metadata string reads now check that the string payload is fully present before advancing the offset
- Truncated metadata now fails with **explicit messages**:

  ```
  truncated stream metadata while reading stream_type
  truncated stream metadata while reading encodings_header
  ...
  ```

### Setup

| Parameter | Value |
|---|---|
| **Baseline** | `main` |
| **Comparison** | this branch |
| **Pinning** | `taskset -c 2` (single core) |
| **Profiler** | `perf stat -r 10` (10 runs, averaged) |
| **Tile sets** | all `.mlt` fixtures under `test/expected/tag0x01` (bing, omt, amazon, amazon_here, simple) |

### Results -- all tag0x01 fixtures

| Metric | Baseline | This branch | Change |
|---|---|---|---|
| task-clock | 8.706 s | 7.936 s | **-8.8%** |
| cycles | 31.22 B | 27.58 B | **-11.6%** |
| IPC | 2.76 | 3.27 | **+18.5%** |
| cache-misses | 165.95 M | 147.95 M | **-10.8%** |

### Results -- per fixture family

| Fixture | task-clock | cycles | IPC |
|---|---|---|---|
| **omt** | -13.9% (7.080 s -> 6.099 s) | -7.0% (22.91 B -> 21.30 B) | 2.69 -> 3.11 |
| **bing** | -11.0% (0.938 s -> 0.835 s) | -13.5% (3.21 B -> 2.78 B) | 2.91 -> 3.09 |

### New coverage for `streamMetadataDecoder`

- Empty input
- Truncation before `stream_type`
- Truncation before `encodings_header`
- Truncation while reading metadata varints
- Truncated MORTON / RLE metadata
- Invalid logical stream subtypes
- Valid NONE, MORTON, and RLE cases

### Other test changes

- Kept one lightweight integration test in `propertyDecoder.spec.ts` to ensure truncated metadata fails before deeper decoder errors
- Added coverage for truncated embedded tileset metadata strings
- Updated `integerStreamDecoder.spec.ts`: one nullable test fixture was incorrect -- it used the logical count where the encoded nullable stream actually stores the compacted physical count. This inconsistency is now surfaced by the stricter metadata decode path; the test now makes the physical vs logical count explicit

### DIFF
--- a/ts/src/decoding/integerStreamDecoder.ts
+++ b/ts/src/decoding/integerStreamDecoder.ts
@@ -528,11 +528,7 @@ export function getVectorType(
     return streamMetadata.numValues === 1 ? VectorType.CONST : VectorType.FLAT;
 }
 
-function isDeltaRleSequenceVarintWidth(
-    data: Uint8Array,
-    offset: IntWrapper,
-    varintWidth: "int32" | "int64",
-): boolean {
+function isDeltaRleSequenceVarintWidth(data: Uint8Array, offset: IntWrapper, varintWidth: "int32" | "int64"): boolean {
     const savedOffset = offset.get();
 
     if (varintWidth === "int64") {


ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
```diff
@@ -13,15 +13,23 @@ const SUPPORTED_FIELD_TYPES = "10-29(scalars), 30(STRUCT)";
...
-    if (length === 0) {
-        return "";
+    const startOffset = offset.get();
+    try {
+        const length = decodeVarintInt32Value(src, offset);
+        if (length === 0) {
+            return "";
+        }
+
+        ensureRemaining(src, offset, length, "string bytes");
```
HarelM, line 23: What does "string bytes" mean in this context?

HarelM:
Can we split this PR somehow? I feel like the extra error handling does not help the readability of the code and the package size.
I think we can safely introduce the performance optimization without the error handling as I think the error handling is a bit more controversial...

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
DoFabien: It is only used as error-context text for the thrown message.

For example, this becomes:
`truncated embedded tileset metadata while reading string bytes`

That said, I agree `string bytes` is not very clear here. `string payload` probably better

DoFabien:
Yes, I agree.

The hardening part did reveal at least two latent correctness issues on the TS side:

- `156ca74a` (`fix(ts): use explicit varint width in vector type detection`)
- one incorrect nullable fixture in `integerStreamDecoder.spec.ts` where the test used the logical count instead of the compacted physical count

So I agree the current PR is mixing concerns a bit too much.

What I propose is to split this into 3 PRs:

1. **Correctness fixes exposed by the hardening**
   - the explicit 64-bit varint-width fix in vector-type detection
   - the incorrect nullable test fixture fix

2. **Hot-path performance improvements**
   - the optimization work itself, which is the main value of this branch

3. **Decoder hardening / earlier error reporting**
   - the stricter validation and clearer failures
   - this part is more controversial in terms of readability / package size, but it is also useful to surface real decoder issues earlier and to make debugging easier, e.g. issue #1124

I grouped them initially because they touched the same decode path, but I agree the split would make review easier.

DoFabien:
Closing this PR in favor of two narrower PRs:

- #1150 for the correctness fixes that were surfaced while working on the metadata hardening path
- #1152 for the metadata decode hot-path optimization on its own

I think this split makes the review much clearer

If we want to go further in that direction, I can open a separate PR for the metadata hardening later this week.

PR #1149, opened 2026-03-13
chore(rust): introduce staging types in Rust layer implementation

nyurik:


rust/mlt-core/src/frames/v01/stream/model.rs
```diff
@@ -13,6 +13,10 @@ pub enum LogicalTechnique {
 }
 
 /// Metadata for RLE decoding
+/// TODO v2 optimizations:
+///   * runs is identical to half the size of the associated array
+///   * num_rle_values is identical to the size of the sum of the first half of the array.
```
CommanderStorm:
```suggestion
///   * `num_rle_values` is identical to the size of the sum of the first half of the array.
```

PR #1150, opened 2026-03-14
fix(ts): vector type sequence was not well read.

DoFabien:
While working on the TS metadata hardening in PR `#1146`, we surfaced two latent issues:

1. a real decoder bug in 64-bit vector-type detection
2. an incorrect nullable test fixture in `integerStreamDecoder.spec.ts`

### 1. 64-bit varint width in `getVectorType(...)`

`getVectorType(...)` uses a small `VARINT` read to determine whether a `DELTA + RLE` stream can be represented as a `SEQUENCE`.

That internal check was always using the 32-bit varint path, which is incorrect for `ID64` / `INT_64` / `UINT_64` streams.

This PR makes that check explicit and width-aware, so 64-bit callers use the 64-bit varint path during vector-type detection.

### 2. Incorrect nullable fixture in `integerStreamDecoder.spec.ts`

One nullable test fixture was incorrect: it used the logical count where the encoded nullable stream actually stores the compacted physical count.

The test now makes the physical vs logical count explicit.

### DIFF
--- a/ts/src/decoding/integerStreamDecoder.spec.ts
+++ b/ts/src/decoding/integerStreamDecoder.spec.ts
@@ -28,7 +28,13 @@ import {
     encodeInt64UnsignedNone,
     encodeUnsignedInt32Stream,
 } from "../encoding/integerStreamEncoder";
-import { encodeVarintFloat64, encodeVarintInt64, encodeZigZagInt64Value } from "../encoding/integerEncodingUtils";
+import {
+    encodeDeltaRleInt32,
+    encodeVarintFloat64,
+    encodeVarintInt64,
+    encodeZigZagInt32Value,
+    encodeZigZagInt64Value,
+} from "../encoding/integerEncodingUtils";
 
 describe("getVectorType", () => {
     it("should return FLAT for RLE with 0 runs", () => {
@@ -69,8 +75,8 @@ describe("getVectorType", () => {
 
     it("should return SEQUENCE for RLE run with 2 runs", () => {
         const metadata = createRleMetadata(LogicalLevelTechnique.DELTA, LogicalLevelTechnique.RLE, 2, 5);
-        const data = new Uint8Array([1, 4, 2, 2]); // Can't achieve this array using the encoding method...
-        const result = getVectorType(metadata, 5, data, new IntWrapper(0));
+        const twoRunUnitDeltaVarintPayload = new Uint8Array([1, 4, 2, 2]); // Can't achieve this array using the encoding method...
+        const result = getVectorType(metadata, 5, twoRunUnitDeltaVarintPayload, new IntWrapper(0));
         expect(result).toBe(VectorType.SEQUENCE);
     });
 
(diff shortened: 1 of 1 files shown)


ts/src/mltDecoder.ts
```diff
@@ -184,7 +184,13 @@ function decodeIdColumn(
     const idDataType = scalarTypeMetadata.longID ? ScalarType.UINT_64 : ScalarType.UINT_32;
     const nullabilityBuffer = typeof sizeOrNullabilityBuffer === "number" ? undefined : sizeOrNullabilityBuffer;
 
-    const vectorType = getVectorType(idDataStreamMetadata, sizeOrNullabilityBuffer, tile, offset);
+    const vectorType = getVectorType(
+        idDataStreamMetadata,
+        sizeOrNullabilityBuffer,
+        tile,
+        offset,
+        idDataType === ScalarType.UINT_64 ? { varintWidth: "int64" } : undefined,
```
HarelM: Maybe pass down the scalar type instaed of adding a string type?

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -479,6 +479,7 @@ export function getVectorType(
     sizeOrNullabilityBuffer: number | BitVector,
     data: Uint8Array,
     offset: IntWrapper,
+    options?: { varintWidth?: "int32" | "int64" },
```
HarelM: Maybe a boolean is enough instead of an option object? Seems a bit odd given the other input parameters.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -525,6 +528,20 @@ export function getVectorType(
     return streamMetadata.numValues === 1 ? VectorType.CONST : VectorType.FLAT;
 }
 
+function isDeltaRleSequenceVarintWidth(data: Uint8Array, offset: IntWrapper, varintWidth: "int32" | "int64"): boolean {
```
HarelM, line 531: Should this be in the integer decoding utils?

DoFabien: I kept it local on purpose.
It is not a generic integer decoding utility, but part of the `getVectorType(...)` heuristic for the specific `DELTA RLE` sequence detection path.

ts/src/decoding/integerStreamDecoder.ts
DoFabien: I updated this to use a direct `varintWidth: "int32" | "int64"` parameter instead of an option object.
That seemed like a better compromise here: it keeps the call site explicit, while staying simpler than the extra object wrapper.
Does that seem reasonable to you?

ts/src/decoding/integerStreamDecoder.ts
HarelM: Fair enough.

ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -509,13 +510,15 @@ export function getVectorType(
...
     const savedOffset = offset.get();
 
-    let values: Int32Array;
     if (streamMetadata.physicalLevelTechnique === PhysicalLevelTechnique.VARINT) {
-        values = new Int32Array(decodeVarintInt32(data, offset, 4));
-    } else {
-        const byteOffset = offset.get();
-        values = new Int32Array(data.buffer, data.byteOffset + byteOffset, 4);
+        if (isDeltaRleSequenceVarintWidth(data, offset, varintWidth)) {
+            return VectorType.SEQUENCE;
```
HarelM, line 515: Is there a need to progress the offset in this case? I saw that you are returning it back in the `isDeltaRleSequenceVarintWidth` method.

HarelM:
Can you check the coverage report to see if there's a need to add a test maybe?
Also regarding: 
> a real decoder bug in 64-bit vector-type detection

Where was this bug manifested?

ts/src/decoding/integerStreamDecoder.ts
HarelM: Good enough, thanks!

ts/src/decoding/integerStreamDecoder.ts
DoFabien: You are right, this is clearer that way.
This path is really just a peek, so I changed it to decode through a temporary local `IntWrapper` instead of mutating and restoring the caller offset.
That makes the intent more explicit.

DoFabien:
> a real decoder bug in 64-bit vector-type detection
> 
> Where was this bug manifested?

This was surfaced by the TS synthetic `ids64_delta_rle` case while working on `#1146`.

If you want to reproduce it, the failing state was on branch `perf/fix-metadata-hotpath-hardening`, just before the fix commit 156ca74a,  at commit 5d4467c3

The failing test was:
`npx vitest run src/synthetic.spec.ts -t ids64_delta_rle`
with:  the error:

`Error: invalid varint32 at offset=21 while decoding value index=2`

The underlying issue was that `getVectorType(...)` was probing the `DELTA + RLE` path as `varint32` even for a 64-bit stream, and the stricter varint32 decoding in `#1146` made that latent bug visible.

ts/src/decoding/integerStreamDecoder.ts
DoFabien: 1a4dfa6beecd49fc498c2173b29f2271b72c49d1

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -73,6 +73,46 @@ describe("getVectorType", () => {
...
+        expect(result).toBe(VectorType.FLAT);
+    });
+
+    it("should detect SEQUENCE for DELTA+RLE direct int32 payloads with unit deltas", () => {
+        const metadata = {
+            ...createRleMetadata(LogicalLevelTechnique.DELTA, LogicalLevelTechnique.RLE, 2, 5),
+            physicalLevelTechnique: PhysicalLevelTechnique.NONE,
+            byteLength: 16,
+        };
+        const encodedDeltaRleProbe = new Int32Array([1, 4, 2, 2]);
```
HarelM: Is it possible to use one of the encoding methods? It's not clear what is encoded here...

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -73,6 +73,46 @@ describe("getVectorType", () => {
...
+        expect(result).toBe(VectorType.SEQUENCE);
+    });
+
+    it("should return FLAT for DELTA+RLE direct int32 payloads with non-unit deltas", () => {
+        const metadata = {
+            ...createRleMetadata(LogicalLevelTechnique.DELTA, LogicalLevelTechnique.RLE, 2, 5),
+            physicalLevelTechnique: PhysicalLevelTechnique.NONE,
+            byteLength: 16,
+        };
+        const encodedDeltaRleProbe = new Int32Array([1, 4, 2, 4]);
```
HarelM: Same as above: Is it possible to use one of the encoding methods? It's not clear what is encoded here...

ts/src/decoding/integerStreamDecoder.spec.ts
DoFabien: I tried to make the test payload clearer by naming the encoded pieces more explicitly.

I could use the real encoder for the non-unit-delta case, but not for the unit-delta one: the current encoders collapse adjacent identical delta runs into a single run, so this specific 2-run payload cannot be produced directly through the normal encoding path.

That is probably also a sign that the branch being tested here is not reachable in real-world data with the current encoders.

PR #1151, opened 2026-03-14
chore(rust): remove unnecessary to_owned calls

nyurik:
some statics still remain


rust/mlt-core/src/frames/v01/property/strings.rs
```diff
@@ -68,41 +68,14 @@ impl ParsedStrings<'static> {
...
-                .iter()
-                .map(|item| ParsedSharedDictItem {
-                    suffix: Cow::Owned(item.suffix.as_ref().to_string()),
-                    ranges: item.ranges.clone(),
-                })
-                .collect(),
-        }
-    }
-}
+impl ParsedSharedDict<'_> {}
```
CommanderStorm:
```suggestion
```

CommanderStorm (approved):

PR #1152, opened 2026-03-14
perf(ts): optimize metadata decode hot path

DoFabien:
This PR focuses on the **TS metadata/header decode hot path**.

It is scoped to performance only. For context, this work was originally explored alongside metadata hardening in `#1146`, but this PR keeps only the fast-path optimization part.

### Hot-path optimization

- `streamMetadataDecoder` now uses **precomputed lookup tables** and **cached `LogicalStreamType` instances** instead of rebuilding values on the hot path
- Metadata varints are now decoded with a **scalar helper** (`decodeVarintInt32Value(...)`) without intermediate array allocation
- The same scalar varint helper is also used for nearby metadata/header reads in:
  - `mltDecoder.ts`
  - `embeddedTilesetMetadataDecoder.ts`
  - `stringDecoder.ts`

### Results -- `test/expected/tag0x01/`

| Metric        | Baseline (main) | This branch |     Change |
| ------------- | --------------: | ----------: | ---------: |
| task-clock    |         8.196 s |     6.669 s | **-18.6%** |
| cycles        |         32.31 B |     26.20 B | **-18.9%** |
| instructions  |         85.60 B |     68.60 B | **-19.9%** |
| branches      |         19.11 B |     15.88 B | **-16.9%** |
| branch-misses |        104.55 M |     99.84 M |  **-4.5%** |

### DIFF
--- a/ts/src/metadata/tile/streamMetadataDecoder.ts
+++ b/ts/src/metadata/tile/streamMetadataDecoder.ts
@@ -34,6 +34,71 @@ export type RleEncodedStreamMetadata = StreamMetadata & {
     readonly numRleValues: number;
 };
 
+const PHYSICAL_STREAM_TYPE_BY_ID: readonly PhysicalStreamType[] = [
+    PhysicalStreamType.PRESENT,
+    PhysicalStreamType.DATA,
+    PhysicalStreamType.OFFSET,
+    PhysicalStreamType.LENGTH,
+];
+
+const LOGICAL_LEVEL_TECHNIQUE_BY_ID: readonly LogicalLevelTechnique[] = [
+    LogicalLevelTechnique.NONE,
+    LogicalLevelTechnique.DELTA,
+    LogicalLevelTechnique.COMPONENTWISE_DELTA,
+    LogicalLevelTechnique.RLE,
+    LogicalLevelTechnique.MORTON,
+    LogicalLevelTechnique.PDE,
+];
+
+const PHYSICAL_LEVEL_TECHNIQUE_BY_ID: readonly PhysicalLevelTechnique[] = [
+    PhysicalLevelTechnique.NONE,
+    PhysicalLevelTechnique.FAST_PFOR,
+    PhysicalLevelTechnique.VARINT,
+];
+
+const DICTIONARY_TYPE_BY_ID: readonly DictionaryType[] = [
+    DictionaryType.NONE,
+    DictionaryType.SINGLE,
+    DictionaryType.SHARED,
+    DictionaryType.VERTEX,
+    DictionaryType.MORTON,
+    DictionaryType.FSST,
+];
+
+const OFFSET_TYPE_BY_ID: readonly OffsetType[] = [
+    OffsetType.VERTEX,
+    OffsetType.INDEX,
+    OffsetType.STRING,
+    OffsetType.KEY,
+];
+
+const LENGTH_TYPE_BY_ID: readonly LengthType[] = [
+    LengthType.VAR_BINARY,
+    LengthType.GEOMETRIES,
+    LengthType.PARTS,
+    LengthType.RINGS,
+    LengthType.TRIANGLES,
+    LengthType.SYMBOL,
+    LengthType.DICTIONARY,
+];
+
+const DEFAULT_LOGICAL_STREAM_TYPE: LogicalStreamType = {};
+const DATA_LOGICAL_STREAM_TYPES: readonly LogicalStreamType[] = DICTIONARY_TYPE_BY_ID.map((dictionaryType) =>
+    ({
+        dictionaryType,
+    }),
+);
+const OFFSET_LOGICAL_STREAM_TYPES: readonly LogicalStreamType[] = OFFSET_TYPE_BY_ID.map((offsetType) =>
+    ({
+        offsetType,
+    }),
+);
+const LENGTH_LOGICAL_STREAM_TYPES: readonly LogicalStreamType[] = LENGTH_TYPE_BY_ID.map((lengthType) =>
+    ({
+        lengthType,
+    }),
+);
+
 export function decodeStreamMetadata(tile: Uint8Array, offset: IntWrapper): StreamMetadata {
     const streamMetadata = decodeStreamMetadataInternal(tile, offset);
     if (streamMetadata.logicalLevelTechnique1 === LogicalLevelTechnique.MORTON) {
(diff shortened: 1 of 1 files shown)


ts/src/decoding/integerDecodingUtils.ts
```diff
@@ -52,6 +52,34 @@ export function decodeVarintInt32(buf: Uint8Array, bufferOffset: IntWrapper, num
     return dst;
 }
 
+//based on https://github.com/mapbox/pbf/blob/main/index.js
+export function decodeVarintInt32Value(buf: Uint8Array, offset: IntWrapper): number {
```
HarelM: Does this really makes things faster? It looks almost the same as the other method...

DoFabien: Good question. I reran an A/B test for this change in isolation.
I was expecting a larger gain, but the delta is actually modest on its own: about `-1.5%`.

If you think the extra helper is not worth it for the code shape, I can drop it, but we would likely give back around `1.5%` of the overall decoder efficiency, which does not seem negligible to me.

HarelM: Can one method use the other somehow? I would like to reduce code duplication of possible...

DoFabien: I benchmarked the factorized version as well (91c2f669ebddaa1327e32dcb089d8c7a1bb6f3ee) , and it still seems surprisingly more efficient than going back to `decodeVarintInt32(..., 1)[0]` for these scalar reads.

The gap was larger than I expected, so there is probably some JIT/runtime behavior involved here as well.

HarelM: Can the array version use this new method then?

DoFabien: oh, wait, I'am wrong.

ts/src/metadata/tile/streamMetadataDecoder.ts
```diff
@@ -34,6 +34,66 @@ export type RleEncodedStreamMetadata = StreamMetadata & {
...
+    LengthType.PARTS,
+    LengthType.RINGS,
+    LengthType.TRIANGLES,
+    LengthType.SYMBOL,
+    LengthType.DICTIONARY,
+];
+
+const DEFAULT_LOGICAL_STREAM_TYPE = new LogicalStreamType();
+const DATA_LOGICAL_STREAM_TYPES: readonly LogicalStreamType[] = DICTIONARY_TYPE_BY_ID.map(
+    (type) => new LogicalStreamType(type),
```
HarelM: This is yet another example of a class that should be a type, can you replace the class with a type and avoid the need for this initialization?

ts/src/decoding/integerDecodingUtils.ts
DoFabien: I rechecked this more carefully, and the current scalar/factorized version is not a good tradeoff.

The earlier signal I reported was a benchmarking mistake on my side. In particular, one of my follow-up comparisons was not rebuilt correctly after the refactor, so I overstated the effect.

After rerunning this properly:
- the scalar-only change is at best a small/noisy signal in end-to-end decode
- the current factorized version is more complex and can even be slower
- the clearly dominant and stable gain comes from the metadata cache/table changes in `streamMetadataDecoder`

So I think the best path is to keep only that part for now. It is the main win, it is simpler to justify, and it benefits all tiles regardless of the compression technique used.

ts/src/metadata/tile/streamMetadataDecoder.ts
DoFabien: I agree, this is cleaner as a type.

I would treat it as a cleanup/modeling refactor rather than a performance change, since I do not expect a meaningful perf impact from that part itself.

If you want, I can widen the scope of this PR and include it.

HarelM: Probably better as a different PR...

ts/src/decoding/integerDecodingUtils.ts
HarelM: Not sure I fully understand, but feel free to let me know when the PR is ready for review.

DoFabien: I cleaned this up and removed the scalar optimization part.

After rechecking it more carefully, that piece may have had a small benefit in its duplicated form, but the signal was not strong enough and was too close to noise compared to the main optimization in this PR.

So I restored that part to its previous form and kept only the metadata cache/table changes in `streamMetadataDecoder`, which are the core of this PR and have a clear, stable impact on tile decode performance (roughly ~18% on my benchmark runs).

At this point the diff is focused on that part only, so it should be ready for review again.

ts/src/metadata/tile/streamMetadataDecoder.ts
```diff
@@ -93,32 +153,26 @@ function decodePartialRleEncodedStreamMetadata(
...
-            logicalStreamType = new LogicalStreamType(null, Object.values(OffsetType)[stream_type & 0xf] as OffsetType);
+            logicalStreamType = OFFSET_LOGICAL_STREAM_TYPES[stream_type & 0xf] ?? DEFAULT_LOGICAL_STREAM_TYPE;
             break;
         case PhysicalStreamType.LENGTH:
-            logicalStreamType = new LogicalStreamType(
-                null,
-                null,
-                Object.values(LengthType)[stream_type & 0xf] as LengthType,
-            );
+            logicalStreamType = LENGTH_LOGICAL_STREAM_TYPES[stream_type & 0xf] ?? DEFAULT_LOGICAL_STREAM_TYPE;
```
HarelM: It's worth mentioning that this is not exactly the same logic - the previous logic created a new instance every time this code path was hit as opposed to the new code which uses the same instance.
I don't know if this is problematic in any way, but it's worth mentioning.
This is why I mentioned the removal of the `LogicalStreamType` class, as it will make the code simpler in terms of returning an object that conforms to a type rather than creating an instance of a class or reusing the same instance.

So after looking at the changes again, I think it's worthwhile to change the `LogicalStreamType` class to a type as part of this PR to keep improved performance and also keep readability in a good state.

HarelM:
- See #1164

HarelM:
I've gone ahead and changed the logical type from a class to a type, you'll need to resolve conflicts, sorry, shouldn't be too hard though...

DoFabien:
Sorry for leaving this one aside for so long.

I refreshed it on top of the current `main` and kept the diff focused on the `streamMetadataDecoder` hot path

I reran the end-to-end decode benchmark on all `test/expected/tag0x01` fixtures:

| Metric | main | this PR | Change |
| --- | ---: | ---: | ---: |
| task-clock | 7.241 s | 5.979 s | -17.4% |
| cycles | 28.42 B | 23.53 B | -17.2% |
| instructions | 81.80 B | 66.92 B | -18.2% |
| branches | 18.21 B | 15.50 B | -14.9% |
| branch-misses | 103.97 M | 102.32 M | -1.6% |
| elapsed | 6.661 s | 5.377 s | -19.3% |
| median cpuNs/byte | 7.68 | 6.29 | -18.1% |

HarelM:
@DoFabien can you address the remaining comment above?
Retuning the object of the streamtype can cause issues if, for some reason, someone will mutate this object down the line, so I think it will be better to create a copy instead of using it as is.

DoFabien:
I agree this is safer.

`decodeStreamMetadata` now keeps the enum lookup tables, but creates a fresh `logicalStreamType` object for each decoded stream metadata instead of returning a shared cached object.

There is a small overhead compared with returning the cached object directly, but the main gain is still there. I reran the same benchmark on all `test/expected/tag0x01` fixtures:

| Metric | main | this PR | Change |
| --- | ---: | ---: | ---: |
| task-clock | 7.294 s | 6.152 s | -15.7% |
| cycles | 28.31 B | 22.67 B | -19.9% |
| instructions | 81.78 B | 66.57 B | -18.6% |
| branches | 18.21 B | 15.42 B | -15.3% |
| branch-misses | 103.82 M | 100.64 M | -3.1% |
| elapsed | 6.724 s | 5.586 s | -16.9% |
| median cpuNs/byte | 7.54 | 6.41 | -15.0% |

HarelM (approved):
Thanks!

HarelM:
A new version should be release momentarily. Thanks for working on this!

PR #1153, opened 2026-03-14
chore(rust): rework WASM code to use TileLayer01

nyurik:


rust/mlt-core/src/frames/v01/tile.rs
```diff
@@ -85,7 +85,10 @@ impl TryFrom<StagedLayer01> for TileLayer01 {
             match prop {
                 StagedProperty::SharedDict(sd) => {
                     for item in &sd.items {
-                        property_names.push(format!("{}:{}", sd.prefix, item.suffix));
+                        // The wire-format prefix may already end with ':' (e.g. "name:"),
+                        // so concatenate directly rather than adding a separator.
+                        let name = format!("{}{}", sd.prefix, item.suffix);
+                        property_names.push(name);
```
CommanderStorm:
```suggestion
                        property_names.push(format!("{prefix}{suffix}", prefix=sd.prefix, suffix=item.suffix));
```

PR #1163, opened 2026-03-15
chore(rust): add stateful decoder

nyurik:
based on #1162

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -64,9 +64,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.13"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
 
 [[package]]
 name = "anstyle-parse"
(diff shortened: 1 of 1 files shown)


rust/mlt-core/src/decoder.rs
```diff
@@ -0,0 +1,79 @@
...
+    #[inline]
+    pub fn consume(&mut self, size: u32) -> Result<(), MltError> {
+        Self::add(&mut self.bytes_used, size, self.max_bytes)
+    }
+
+    /// Take `size` bytes from the reservation budget.
+    /// Used when we know we will need to allocate in a later pass, but don't want to charge the budget yet.
+    #[inline]
+    pub fn reserve(&mut self, size: u32) -> Result<(), MltError> {
+        Self::add(&mut self.bytes_reserved, size, self.max_bytes)
```
CommanderStorm: a bit weird that we have to do the follow on every memory op.
```
dec.reserve(i)
dec.consume(i)
```

I think a safer api would be

```suggestion
    /// Take `size` bytes from the allocation budget. Call this before the actual allocation.
    #[inline]
    pub fn consume(&mut self, size: u32) -> Result<(), MltError> {
        Self::add(&mut self.bytes_used, size, self.max_bytes)
        if self.bytes_used > self.bytes_reserved {
            self.bytes_reserved = self.bytes_used;
        }
    }

    /// Take `size` bytes from the reservation budget.
    /// Used when we know we will need to allocate in a later pass, but don't want to charge the budget yet.
    #[inline]
    pub fn reserve(&mut self, size: u32) -> Result<(), MltError> {
        Self::add(&mut self.bytes_reserved, size, self.max_bytes)
```

CommanderStorm (approved):
I am not sure of the usage is solid across the board.
this two step reserving is very werid..

PR #1168, opened 2026-03-15
chore(rust): memory allocation budgeting

nyurik:


rust/mlt-core/benches/decoding_e2e.rs
```diff
@@ -1,7 +1,14 @@
...
 use criterion::{BatchSize, BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
-use mlt_core::{Decoder, parse_layers};
+use mlt_core::{Decoder, Parser, parse_layers};
+
+fn parser() -> Parser {
+    Parser::default()
+}
+fn dec() -> Decoder {
+    Decoder::default()
+}
```
CommanderStorm: why not inlined?

nyurik: far too many of these all over the place in testing - makes it harder to read.  I may revisit this though, will see

PR #1176, opened 2026-03-17
chore(java): Generalize data input for MLT encoding

TimSylvester:
This PR abstracts the Feature record into an interface and implements it in terms of MVT along with a preliminary MLT version for decoding MLT tiles, with the beginnings of support for Parquet-style nested properties.

The Layer/Feature hierarchy was based on MVT, and relied on type inspection to interpret the contents of the properties.  The properties are promoted from a generic map to a class with type information.  Lists of layers, features, and properties are changed from lists to methods that can get or generate the data as necessary.

The expectation is that a new implementation can allow for direct conversion from other data sources without imposing limitations of the MVT format on them.

Also:
- Moves `ColumnMapping` support out of the `mvt` package, as it's not specific to MVT
- Separates `FieldType` from `Field` for reuse in properties
- Makes `Feature`, `Field`, `Column` types immutable, with builder pattern
- Sorts the column items at the end of building metadata instead of sorting the properties of every feature to ensure they are encountered in name order
- Avoid creating a new property object for every property of every feature, instead allowing an indexed lookup per property
- Reuses `CompareHelper` in `TestUtils` comparison methods
- Allows loose tiles to be compressed in `Encode`
- Eliminates unused code and adds test coverage


java/mlt-core/src/main/java/org/maplibre/mlt/data/MLTFeature.java
```diff
@@ -0,0 +1,158 @@
...
+        + ", properties="
+        + properties
+        + "]";
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static Builder builder(@Nullable MLTFeature feature) {
```
CommanderStorm: I think the ideomatic way to do builders in java was via these `@Builder` annotations, but I might be wrong.

CommanderStorm (commented):
For a 5k line change, I won't give a full review.
I looked at the parts I care about (syntehtics, MLT data structures) but skipped the rest.

I just don't know what is ideomatic..
Maybe @louwers has more depth in java.

java/mlt-core/src/main/java/org/maplibre/mlt/data/MLTFeature.java
TimSylvester: That's a new dependency, I can pull it in if people want.  Looks like it does support F-bounded polymorphic builders, which is what makes these ugly.

TimSylvester: Here's what that looks like:
https://github.com/TimSylvester/maplibre-tile-spec/compare/encode-split...TimSylvester:maplibre-tile-spec:encode-split-2?expand=1
@louwers Thoughts on the dependency?  I was able to make it `compileOnly` so I guess that means it doesn't affect users of the library at all, right?
`SuperBuilder` is "experimental"...

TimSylvester: @louwers I switched everything over and it's working well, but Javadoc is complaining about not finding the generated builder classes.  I added a config file that makes the annotation processor add a generated attribute to them, but I think Javadoc might need to be told about `lombok.Generated` and I can't see how to configure it.  Any ideas?

TimSylvester: I found a workaround that seems to work: explicitly declaring the empty builder classes.

TimSylvester:
@louwers I think this is ready to go

java/mlt-core/src/main/java/org/maplibre/mlt/data/Feature.java
```diff
@@ -1,28 +1,80 @@
...
+import lombok.Getter;
+import lombok.experimental.Accessors;
+import lombok.experimental.SuperBuilder;
+import org.jetbrains.annotations.NotNull;
 import org.locationtech.jts.geom.Geometry;
+import org.maplibre.mlt.metadata.tileset.MltMetadata;
 
-public record Feature(boolean hasId, long id, Geometry geometry, Map<String, Object> properties) {
+/** A base class for features */
+@SuperBuilder(toBuilder = true)
```
louwers, line 17: This is an experimental Lombok feature. (It has been for 7 years).

I wonder how @msbarry feels about Lombok?

https://projectlombok.org/features/experimental/SuperBuilder

msbarry: Lombok isn't bad overall, but a few of it's features are no longer necessary since java has improved (`@var` -> `var`, records). I'd try to stick with records as much as possible, I often use lombok to add builder/toBuildler to  a java record until java [supports that natively](https://github.com/openjdk/amber-docs/blob/master/eg-drafts/reconstruction-records-and-classes.md). As a consumer of the library it should be transparent since it all gets compiled away so no concerns from planetiler side.

louwers: >  As a consumer of the library it should be transparent since it all gets compiled away so no concerns from planetiler side.

Alright that is good to know. Whatever @TimSylvester is comfortable with then.

TimSylvester: It did reduce the amount of explicit code but it was also a little awkward in some cases, including not being able to use `record`.  

The constrained generics needed for builders on nested classes are pretty ugly, so I'm happy to have all that hidden.  At the same time, I wouldn't object to bringing them back, I had to jump through some hoops to make it work that might be confusing to readers.  If the builders could go in separate files I would say they should be explicit, but apparently that's impossible.

Another approach would be to drop nested builders and always build the most-derived classes, duplicating setters in derived classes, and effectively upcast them on use.  Or perhaps require explicit type declarations on setters.  But either makes using them more confusing.

Or we could turn all the inheritance into composition, eliminating any need for `SuperBuilder`.

TimSylvester: I tried that last idea in the Field/Column hierarchy, and in the process I decided that each level has few enough fields that builders really make sense for them.  I think the result is an improvement.  @louwers ?

[https://github.com/TimSylvester/maplibre-tile-spec/compare/encode-split...TimSylvester:maplibre-tile-spec:encode-split-3?expand=1](https://github.com/TimSylvester/maplibre-tile-spec/compare/encode-split...TimSylvester:maplibre-tile-spec:encode-split-3?expand=1)

That doesn't help with the Feature hierarchy, though.

louwers (approved):

PR #1190, opened 2026-03-23
chore(rust): migrate to Rust fastpfor

nyurik:


rust/mlt-core/src/codecs/fastpfor.rs
```diff
@@ -1,55 +1,26 @@
+use crate::utils::AsUsize;
```
CommanderStorm: Grumble.. still don't think this is actually better code for the noise it creates ^^

```suggestion
use crate::utils::AsUsize as _;
```

PR #1192, opened 2026-03-24
chore(rust): cleanup Results, minor styling

nyurik:


rust/mlt-core/src/codecs/fsst.rs
```diff
@@ -181,27 +175,20 @@ mod tests {
...
-        let mut offset = 0;
-        for (s, &len) in values.iter().zip(&lengths) {
-            let len = len as usize;
-            assert_eq!(&corpus[offset..offset + len], *s);
-            offset += len;
+        let cases: &[&[&str]] = &[
+            &["hello world", "hello rust", "hello fsst", "world"],
+            &["hello"],
+        ];
+        for case in cases {
```
CommanderStorm: please use rstest for property testing instead since it is simpler to add new properties this way.

PR #1210, opened 2026-03-26
chore(rust): improve synthetics, geojson

nyurik:
* now synthetics do not duplicate mlts, only unique are placed to the -rust
* when running, it validates binary identity with /0x01/
* added generation of the fastpfor synthetics
* consolidate geojson comparison as part of core
* consolidate geotype conversion


rust/mod.just
```diff
@@ -367,91 +362,6 @@ wasm-test: wasm-build
 validate-synthetic:
     {{just}} mlt ls --validate-to-json --details basic ../test/synthetic/
```
CommanderStorm, line 363: I would still like the previous code more.
Debugging this LS command is not very user-friendly..

nyurik: the errors would still print - just not the passing ones.  Besides, validation is now fully done by the synthetics generator . I think we should use ls to validate all files, not just synthetics, e.g. after generating everything that is not checked in

PR #1211, opened 2026-03-26
fix(java): shared dict stream count

nyurik:
Java encoder was incorrectly adding 1 to the shared dictionaries stream count.

Note that all parsers seem to have safely been ignoring it -- possibly only using it for memory reservation only

### DIFF
--- a/java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/StringEncoder.java
+++ b/java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/StringEncoder.java
@@ -117,6 +117,7 @@ public class StringEncoder {
       // TODO: make present stream optional
       final var encodedPresentStream =
           BooleanEncoder.encodeBooleanStream(presentStream, PhysicalStreamType.PRESENT);
+      numStreams += 2;
 
       final var encodedDataStream =
           IntegerEncoder.encodeIntStream(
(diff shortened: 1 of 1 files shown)


nyurik:
> > Note that all parsers seem to have safely been ignoring it -- possibly only using it for memory reservation only
> 
> I think the rust decoder has a fixme where I said that we should error out here. Adding this error now that it is possible would be a nice touch.

I already have a fix for this in Rust - i'll merge it once this gets in

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/StringEncoder.java
```diff
@@ -106,8 +113,11 @@ public static Pair<Integer, ArrayList<byte[]>> encodeSharedDictionary(
       }
 
       final var encodedFieldMetadata = EncodingUtils.encodeVarints(new int[] {2}, false, false);
+
+      // TODO: make present stream optional
```
HarelM, line 117: Probably worth addressing this or removing it.

nyurik: This is clearly a separate PR optimization - make java code optimize away the presence stream when it is not needed. v1 should support it even if Java encoder never generates it.

PR #1225, opened 2026-03-29
feat(rust): enhance property iteration

nyurik:
* migrate towards iterator-based access as the primary way to access data
* some data has been marked as `pub(crate)` - more to be done later

### DIFF
--- a/rust/mlt-core/src/frames/v01/iterators.rs
+++ b/rust/mlt-core/src/frames/v01/iterators.rs
@@ -452,8 +452,6 @@ impl<'a> Iterator for FeatPropertyIter<'a> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


rust/mlt-core/src/frames/v01/iterators.rs
```diff
@@ -223,6 +452,8 @@ impl<'a> Iterator for FeatPropertyIter<'a> {
     }
 }
 
+// ── Feature iterator ──────────────────────────────────────────────────────────
+
```
CommanderStorm: LLM artefact :)

```suggestion
```

CommanderStorm (approved):
Thanks

PR #1230, opened 2026-03-30
chore(rust): dedup a lot of scalar code

nyurik:


rust/mlt-core/src/frames/v01/property/scalars.rs
```diff
@@ -97,3 +102,303 @@ impl From<ScalarEncoder> for PropertyEncoder {
...
+/// ```ignore
+/// scalar_match!(some_scalar, label, v => f.debug_tuple(label).field(v).finish())
+/// ```
+///
+/// **Form 4** – zip-match two scalars of different families; `$diff` is evaluated when
+/// the variants do not match:
+/// ```ignore
+/// scalar_match!(left, right, a, b => *a == b, else false)
+/// ```
+macro_rules! scalar_match {
```
CommanderStorm, line 199: I am not sure if this improves legibilitiy.
I think most of the times, custom maros make code less legible in the long run, this is one of them.

I would appreciate splitting or inlining, a 4 way macro is hard to reasona about

rust/mlt-core/src/frames/v01/root.rs
```diff
@@ -42,12 +42,20 @@ impl<'a> Layer01<'a, Lazy> {
...
 
+            macro_rules! push_scalar {
+                ($variant:ident, $parse_stream:expr) => {{
+                    (input, opt) = parse_optional(column.typ, input, parser)?;
+                    (input, value) = $parse_stream;
+                    properties.push(Raw(RawProperty::Scalar(
+                        Scalar::<RawScalarFam<'a>>::$variant(scalar(name, opt, value)),
+                    )));
+                }};
+            }
```
CommanderStorm, line 57: Previous code was more duplicated, but this is not better, imo this is more towards the obvuscation side since reading either this or below code is much harder now.

I would prefer the previous code

rust/mlt-core/src/frames/v01/property/scalars.rs
```diff
@@ -97,3 +102,303 @@ impl From<ScalarEncoder> for PropertyEncoder {
...
+    Bool(F::Of<bool>),
+    I8(F::Of<i8>),
+    U8(F::Of<u8>),
+    I32(F::Of<i32>),
+    U32(F::Of<u32>),
+    I64(F::Of<i64>),
+    U64(F::Of<u64>),
+    F32(F::Of<f32>),
+    F64(F::Of<f64>),
+}
```
CommanderStorm, line 138: This is fairly advanced code concepts and also not trivially understabdable what is going on for other engineers.

I would prefer the simple solution that we previously had since that spelled things out with less derive + trait + GAT, so in effect a bit clearer.

Yes I can still understand this, but this is more complicated than it needs to and I would like the simpele solutoon here

CommanderStorm:
this needs a re-think a bit

PR #1233, opened 2026-03-31
chore(rust): add tests with no presence streams

nyurik:
A lot of new tests for when all values are present in a column, thus not needing a presence stream.

* #1237
* #1238

Note that JS version passes a lot more of them except for fsst (not supported?)

### DIFF
--- a/cpp/tool/synthetic.test.ts
+++ b/cpp/tool/synthetic.test.ts
@@ -11,7 +11,28 @@ import { describe, expect, it } from "vitest";
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const binary = resolve(__dirname, "../build/tool/mlt-cpp-json");
 
-const SKIPPED_TESTS = [];
+const SKIPPED_TESTS = [
+  "0x01-rust/prop_str_ascii_np",
+  "0x01-rust/prop_str_empty_np",
+  "0x01-rust/prop_str_escape_np",
+  "0x01-rust/prop_str_special_np",
+  "0x01-rust/prop_str_unicode_np",
+  "0x01-rust/props_mixed_np",
+  "0x01-rust/props_no_shared_dict_np",
+  "0x01-rust/props_offset_str_fsst_np",
+  "0x01-rust/props_offset_str_np",
+  "0x01-rust/props_shared_dict_2_same_prefix_np",
+  "0x01-rust/props_shared_dict_fsst_np",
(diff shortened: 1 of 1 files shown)


ts/src/synthetic.spec.ts
```diff
@@ -10,8 +10,17 @@ import type FeatureTable from "./vector/featureTable";
 const EARCUT_MAX_RINGS = 500;
 
 const UNIMPLEMENTED_SYNTHETICS: string[] = [
-    "0x01-rust/poly_multi_morton_ring_morton",
-    "0x01-rust/poly_multi_morton_ring_no_morton",
+    "0x01-rust/props_offset_str_fsst_np",
```
HarelM, line 13: I'm not sure I understand the concept of `0x01-rust` folder, if we have folder based on the encoder then I would expect `0x01` to be renamed to `0x01-java`.
Another option is to make sure that when using the same input for the encoders we get the same output and avoid this cumbersome folder structure.

nyurik: I recently simplified/deduped -rust dir code, so now:
* `0x01/*.{mlt,json} ` files are re-generated by `just java::generate-synthetic-mlts`.  Most importantly the directory is emptied before regenerating files - so Java is the source of truth for which files are and are not in it.
* `just rust::generate-synthetic-mlts` has a bigger list of tests which do something differently:
  * if the file exists in `0x01/`, Rust validates file content is identical to Rust's synthetic
  * if the file does not exist, it will be placed into `0x01-rust/` so that Java does not delete-recreate these files
  * if the name of the test contains `fsst`, we know that encoder is not binary-identical, but decoded version must match. So it gets written to `0x01-rust`, but its decoded `.json` version is compared with the one in `0x01/`.  Same happens for any tests with `-rust` suffix (suffix is removed).
* lastly, we ensure that all .mlt files in all synthetics subdirs are unique - so that we don't accidentally have dups in `0x01/` and `0x01-rust`

As the result of all this, we do not have any duplicates, and we allow java-generated synthetics not to collide with the rust-generated ones.

Does it make sense? Any suggestions to improve it?

CommanderStorm: The reason why there is this is that fsst has multiple ways of being compressed.
Rust just chooses sligtly different ways of doing this.

You also would not expect gzip to behave the same way, but you would expect it to decompress to the same thing.
For the tests where there is the -rust, we can move the java version to a different folder.

That is a different piece of work from this PR though.

PR #1234, opened 2026-03-31
fix(ts): handle fsst dict with non-optional values

nyurik:
tiny vibed fix for #1233 -- needs an expert to review


HarelM:
If this fixes the failing tests in #1233 I would recommend adding it there since it will have the relevant tests to make sure it passes and doesn't break other things.
It would also be great to have a unit test to cover this scenario if possible. While the synthetic tests are great, I prefer to have a unit test as well to make sure this is covered and easy to understand if and when it fails.
Having said that, if you are in a hurry, this can be merged as is.

nyurik:
@HarelM I do not want to fix unrelated code as part of test generation PR.  #1233 identified new bugs/missing features in multiple languages. These issues are already in the main branch. In 1233, i added the new failing tests to typescript ignore list - this way you can fix it at a different timeline than adding new test cases to the main branch.  Moreover, I do not know enough about the typescript codebase - here I just wanted to mention the **partial** fix to just a few of the new failing test cases.

HarelM:
The only failing tests in typescript as far as I know are from invalid geometries, which I don't plan to fix in the near future because of how maplibre-gl works.
In any case, I'll wait for the other PR to be merged before merging this fix as this PR should have tests to cover it, and the other PR is what will add those tests.

nyurik:
I accidentally pushed the wrong changes to this branch. Now reverted and cleaned up.  Feel free to make any additional modifications to this as needed

HarelM:
- Superseded by #1331

PR #1252, opened 2026-04-02
Revert "chore(rust): try to remove java install for rust builds"

CommanderStorm:
Reverts maplibre/maplibre-tile-spec#1249

I will need this for the JNI PR that I have not yet rebased onto main


nyurik:
if you need it, please add it together with the PR - this way it only gets merged as part of it, and it is assocciated with that commit - so we can trace why we need it. Lastly, this way it all gets reviewed together

PR #1285, opened 2026-04-10
perf(rust): Don't allocate for calculating the hilbert params and resulting parameters

CommanderStorm:
Previous version was doing unnesary work, can be rustified


louwers (approved):
LGTM

PR #1292, opened 2026-04-13
test(java): Extend synthetic tests to include rings

TimSylvester:
The existing `poly_morton_ring...` cases get the decoder into the second, "indexed" branches of the poly/multi decoder, but do not exercise the rings loop within that relevant to #442.

This uses the same coordinates as rings within the same polygon in order to reach that loop, and makes them un-equal lengths, in order to make it an error if the ring size is not used appropriately instead of the shell size.

It might be simpler/better to change the split from half to quarter in the existing cases instead of rebuilding the rings, but I didn't know if I should avoid changing those cases.


CommanderStorm (approved):
I need to investigate where this breaks the rust encoder.
If you want to add this a simple skip for this test is fine, otherwise I will try to get to this over the course of the week

PR #1297, opened 2026-04-14
fix(rust): geo builds on wasm, remove unnecessary feature gates

michaelkirk:
I was reviewing the `geo` usage and noticed some comments about geo not building on wasm. 

> Note that geo crate is not WASM compatible until next release, so had to workaround it

The next release will happen soon, but the current version of `geo` also builds on wasm. 

I wonder if you were hitting https://github.com/georust/geo/pull/1476 ?

If so, the issue is that the maintainers of `rand` expect wasm clients using the `rand` crate to explicitly depend on `getrandom` for their leaf (binary/application) crates. 

For context, see:
https://github.com/rust-random/rand/issues/886?issue=rust-random%7Crand%7C1694

I assume (correct me if Im wrong) that this build issue was why the `mlt-core/tessellate` and `mlt-core/sort-coords-iter` feature flags were introduced, so I've also removed those and made `geo` mandatory. 

Let me know if I'm misunderstanding the situation, or if there's some other way to see the specific error message you're getting when trying to build for wasm.

### DIFF
--- a/rust/mlt-core/src/encoder/optimizer.rs
+++ b/rust/mlt-core/src/encoder/optimizer.rs
@@ -65,19 +65,6 @@ impl TileLayer01 {
         }
 
         let mut sort_by = vec![SortStrategy::Unsorted];
-        {
-            let try_spatial_sort = cfg.try_spatial_morton_sort || cfg.try_spatial_hilbert_sort;
-            if try_spatial_sort
-                && (self.features.len() < SORT_TRIAL_THRESHOLD
-                    || spatial_sort_likely_to_help(&self))
-            {
-                if cfg.try_spatial_morton_sort {
-                    sort_by.push(SortStrategy::SpatialMorton);
-                }
-                if cfg.try_spatial_hilbert_sort {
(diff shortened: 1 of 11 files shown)


rust/mlt-core/src/encoder/optimizer.rs
CommanderStorm, line 78: The scope here only existed because of the cfg (github mangled whitespace in the suggestion, but rustfmt will fix that ^^)

```suggestion
let try_spatial_sort = cfg.try_spatial_morton_sort || cfg.try_spatial_hilbert_sort;
if try_spatial_sort
&& (self.features.len() < SORT_TRIAL_THRESHOLD
|| spatial_sort_likely_to_help(&self))
{
if cfg.try_spatial_morton_sort {
sort_by.push(SortStrategy::SpatialMorton);
}
if cfg.try_spatial_hilbert_sort {
sort_by.push(SortStrategy::SpatialHilbert);
}
}
```

CommanderStorm (approved):
You are the best! ❤️ 
If we ever meet at one of the foss4g/sotm events, your beers go on me.

PR #1305, opened 2026-04-15
refactor(rust): update geo, simplify tessellation

michaelkirk:
- adapt to deprecations in geo::Triangle accessors
- simplify tessellation now that geo triangulation omits the redundant "closing" coordintae
  - See https://github.com/georust/geo/pull/1507
  - See https://github.com/georust/geo/pull/1508
- getrandom no longer required by geo

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1032,12 +1032,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d0881ea181b1df73ff77ffaaf9c7544ecc11e82fba9b5f27b262a3c73a332555"
 
 [[package]]
-name = "earcutr"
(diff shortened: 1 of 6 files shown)


michaelkirk:
I forgot to mention: triangulation should be a bit faster with this version of geo. I have a couple ideas for further work here, but haven't had a chance to measure the impact yet.

See https://github.com/georust/geo/pull/1508

nyurik (approved):
thanks!  looks great, i have a follow up cleanup that we can do separately

nyurik:
@michaelkirk thx for the update! Would it make sense to port earcut dependency to be an algo in the `geo` crate that uses geo-types directly, thus allowing tighter integration -- removing the need to go from i32->f64->i32?

PR #1306, opened 2026-04-15
feat(rust): mlt UI to browse .mbtiles

nyurik:
Just browse the .mbtiles file the way you browse a map in a browser.

Note:  All of the `ui` code at this point has been vibed, including this new .mbtiles browser. Eventually I plan to clean it up, but at this stage this is a testing tool that does not need to be as carefully planned as the core code.


CommanderStorm (approved):

CommanderStorm:
(fix CI though)

PR #1307, opened 2026-04-16
feat(rust): flamegraph and a minor optimization

nyurik:


rust/mlt/src/convert.rs
```diff
@@ -35,6 +32,41 @@ const SAVE_EVERY: Duration = Duration::from_secs(60);
...
+/// Pick an internal checkpoint strategy.
+///
+/// Large datasets prefer PASSIVE to avoid expensive final WAL truncation.
+fn choose_checkpoint_mode(tile_count: u64) -> CheckpointMode {
+    if tile_count > 100_000 {
+        CheckpointMode::Passive
+    } else {
+        CheckpointMode::Truncate
+    }
+}
```
CommanderStorm, line 68: I would prefer we default to truncating.
Being smaller is more important than <<4% faster we still have those 20% perf gains that can be made

rust/mlt/src/convert.rs
```diff
@@ -35,6 +32,41 @@ const SAVE_EVERY: Duration = Duration::from_secs(60);
...
+/// Pick an internal mbtiles batch size based on total tile count.
+///
+/// Larger datasets benefit from larger batches (fewer transactions), while
+/// smaller datasets keep memory usage lower with modest batch sizes.
+fn choose_batch_size(tile_count: u64) -> usize {
+    match tile_count {
+        0..=10_000 => 500,
+        10_001..=100_000 => 2_000,
+        _ => 5_000,
+    }
```
CommanderStorm, line 56: You have benchmarked this, right?

The wisdom from the DB realm is that approx. Every 100ms  a morsel should finish.
https://db.in.tum.de/~leis/papers/morsels.pdf

I chose 500 because only above time goes up.
Likely should have made this a power of two

CommanderStorm (approved):
Assuming the three comments are addressed

CommanderStorm:
code has diverged too much, if we do this it needs a re-attempt

PR #1310, opened 2026-04-17
chore(rust): cleanup Morton structs

nyurik:


rust/mlt-core/src/codecs/hilbert.rs
```diff
@@ -18,14 +18,14 @@ pub fn hilbert_xy_to_index(level: u32, x: u32, y: u32) -> u32 {
...
-/// per-axis).  `num_bits` is the grid level — both shifted components must
-/// fit in `[0, 2^num_bits)`.
+/// per-axis).  `bits` is the grid level — both shifted components must
+/// fit in `[0, 2^bits)`.
 ///
-/// Use [`hilbert_curve_params_from_bounds`] to compute `shift` and `num_bits`
+/// Use [`hilbert_curve_params_from_bounds`] to compute `shift` and `bits`
 /// from global min/max coordinates.
 #[must_use]
-pub fn hilbert_sort_key(c: Coord32, shift: u32, num_bits: u32) -> u32 {
```
michaelkirk, line 27: The new name is shorter, but it's more ambiguous. The previous name was better IMO.

rust/mlt-core/src/decoder/stream/logical.rs
```diff
@@ -85,8 +84,8 @@ impl LogicalValue {
                 let expanded = rle.decode(data, dec)?;
                 decode_zigzag_delta::<i32, _>(&expanded, dec)
             }
-            LogicalEncoding::Morton(meta) => decode_morton_codes(data, meta, dec),
-            LogicalEncoding::MortonDelta(meta) => decode_morton_delta(data, meta, dec),
+            LogicalEncoding::Morton(meta) => meta.decode_codes(data, dec),
```
michaelkirk:
```suggestion
            LogicalEncoding::Morton(morton) => morton.decode_codes(data, dec),
```

Now that `Meta` isn't part of the type, `meta` makes less sense as a variable name.

rust/mlt-core/src/codecs/morton.rs
```diff
@@ -61,51 +63,50 @@ pub fn morton_sort_key(c: Coord32, shift: u32, num_bits: u32) -> u32 {
 }
 
 // ── Encoder ─────────────────────────────────────────────────────────────────
+impl Morton {
+    /// Compute `ZOrderCurve` parameters from the vertex value range.
+    ///
+    /// Returns a [`Morton`] whose `bits` and `shift` match Java's
+    /// `SpaceFillingCurve` implementation.
+    pub fn from_vertices(vertices: &[i32]) -> MltResult<Self> {
```
michaelkirk, line 73: This diff is convoluted, but I'm assuming you strictly moved existing code from free functions to instance methods, without changing any of the functionality. Or did you change any impls?

nyurik:
@michaelkirk thanks for the review!  I have been wondering if this is a good approach overall, or if #1311 would make more sense - it seems we have nearly identical `bits + shift` pattern, so consolidating it this way may be better?  I think @CommanderStorm is also more in favour of it, but i might be wrong

rust/mlt-core/src/codecs/morton.rs
nyurik: no, it was just a minor reorg, no code changes

CommanderStorm (approved):
I am fine with all three, the design is solid either way, not as huge of an opinion from me than you think

PR #1313, opened 2026-04-17
chore(rust): cleanup dict ranges

nyurik:


rust/mlt-core/src/decoder/property/model.rs
```diff
@@ -111,18 +111,34 @@ pub struct ParsedScalar<'a, T: Copy + PartialEq> {
...
+/// Equal `start` and `end` indicate an empty string.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+pub struct DictRange {
+    pub start: i32,
+    pub end: i32,
+}
+impl DictRange {
+    /// Sentinel value indicating a NULL (absent) entry.
+    pub const NULL: Self = Self { start: -1, end: -1 };
```
CommanderStorm, line 127: I somewhat dislike sentinals.
Can we use an enum or option<Range<i32> instead?

rust/mlt-core/src/encoder/property/shared_dict.rs
```diff
@@ -315,7 +293,7 @@ impl StagedSharedDict {
                             ranges.push(encode_shared_dict_range(start, end)?);
                             present_items += 1;
                         }
-                        None => ranges.push((-1, -1)),
+                        None => ranges.push(DictRange::NULL),
```
CommanderStorm, line 296: I don't quite understand this code path. Seems kind of weird,but if we specified it this way..

CommanderStorm (approved):
Commented some things below

rust/mlt-core/src/decoder/property/model.rs
nyurik: I have considered that, but it would instantly blow up the memory usage (u32+u32+bool = 12 bytes) - and we still need to analyze each value, so might as well continue with this

PR #1315, opened 2026-04-18
ci: add collector workflow with path-based filtering

CommanderStorm:
Replace per-language workflow triggers (push/pull_request) with workflow_call, and add a single ci.yml orchestrator that uses dorny/paths-filter to conditionally fan out to each language CI. The ci-passed job serves as the sole required status check, treating skipped jobs as OK and failing only on failure/cancelled.

Kind of vibed, but
- it passes CI on my path, 
- the require steps are the same and
- **now has a consistent job that we can require for automerge/..**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,94 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  changes:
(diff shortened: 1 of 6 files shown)


CommanderStorm:
Also, java CI sometimes seems to need a bit of a retry.
Might be worth to add a retry job for this that automatically fixes this flakyness.

Different issue though..

nyurik (approved):

PR #1317, opened 2026-04-18
chore(java): add fpf-align and prop-type synthetics

nyurik:
just a few more java-generated synthetics for completeness


CommanderStorm (approved):
LGTM, but CI needs a little change.

That CI change is coming in handy already, love it (yes, self-praise.. sue me)

PR #1329, opened 2026-04-22
chore(rust): rework presence to use bitvec

nyurik:
* keep presence as uncompressed bit vector if possible
* use less space even if decode is needed

### DIFF
--- a/rust/mlt-core/src/decoder/stream/decode.rs
+++ b/rust/mlt-core/src/decoder/stream/decode.rs
@@ -10,42 +10,32 @@ use crate::codecs::rle::decode_byte_rle;
 use crate::codecs::varint::parse_varint_vec;
 use crate::decoder::{LogicalEncoding, LogicalValue, PhysicalEncoding, RawStream};
 use crate::errors::{AsMltError as _, fail_if_invalid_stream_size};
-use crate::utils::{AsUsize as _, Presence};
+use crate::utils::AsUsize as _;
 use crate::{Decoder, MltError, MltResult};
 
 impl<'a> RawStream<'a> {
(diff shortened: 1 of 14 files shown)


rust/mlt-core/src/decoder/stream/decode.rs
```diff
@@ -1,16 +1,54 @@
...
+    /// The returned `Presence` is always truncated to exactly `num_values` bits.
+    ///
+    /// Note: currently all presence streams produced by this encoder use
+    /// `LogicalEncoding::Rle`, so the zero-copy path is reserved for future
+    /// uncompressed presence streams.
+    pub fn decode_presence(self, dec: &mut Decoder) -> MltResult<Presence<'a>> {
+        let num_values = self.meta.num_values.as_usize();
+        if self.meta.encoding.physical == PhysicalEncoding::VarInt {
+            return Err(MltError::NotImplemented("varint presence decoding"));
+        }
```
CommanderStorm, line 26: 🤔 maybe something we can tighten up in v2.
We really only need 1 bit here..

rust/mlt-core/src/encoder/property/compare.rs
```diff
@@ -24,27 +21,31 @@ impl<T: Copy + PartialEq> PartialEq<ParsedScalar<'_, T>> for StagedScalar<T> {
     }
 }
 
-// ── StagedOptScalar: expand dense values using presence ──────────────────────
+// ── StagedOptScalar: compare dense values using presence ─────────────────────
```
CommanderStorm, line 25: vibing artefact

```suggestion
```

CommanderStorm (approved):

PR #1333, opened 2026-04-23
fix(rust): terrible fsst compression ratio due to reusing compressor results for unrelated shared dicts

CommanderStorm:
For example the latin one for cyrilic alphabet.


rust/mlt-core/src/encoder/property/shared_dict.rs
```diff
@@ -349,9 +349,16 @@ pub(crate) fn write_shared_dict(
         Some(StrEncoding::Plain | StrEncoding::Dict) => None,
         None => {
             // Populate cache on first sort trial, reuse on subsequent.
+            // Key includes the suffix as otherwise multiple groups could share the same prefix
+            // (e.g. two "name:" groups for Arabic vs Cyrillic scripts).
+            // Since the grouping is done only once, the order inside the items is deterministic, so we can just take the first suffix for the cache key.
+            let first_suffix = shared_dict.items.first().map_or("", |i| &i.suffix);
             enc.fsst_cache
                 .entry(shared_dict.prefix.clone())
```
CommanderStorm:
```suggestion
```

rust/mlt-core/src/encoder/property/shared_dict.rs
```diff
@@ -349,9 +349,16 @@ pub(crate) fn write_shared_dict(
...
-                .or_insert_with(|| fsst_try_train(&dict))
+                .entry(format!(
+                    "{prefix}{first_suffix}",
+                    prefix = shared_dict.prefix
+                ))
```
CommanderStorm, line 360:
```suggestion
                ))
                .or_insert_with(|| fsst_try_train(&dict))
```

nyurik (approved):
for a moment i was considering to use some other non-alloc, e.g. an xxhash that's computed over name+first column - this would avoid constant cloning, but I am not certain how much of an impact that would be... perhaps just for clarity we could still do it if its only a few lines?

PR #1342, opened 2026-04-24
chore(java): Refactor

TimSylvester:
Just refactoring to remove redundancy and improve efficiency, no intended behavior changes.

- Merge `encodeGeometryColumn` and `encodePretessellatedGeometryColumn` which were largely identical
- Same with `encodePolygonPretessellationStreams` and `encodePolygonPretessellationStreamsWithOutlines`
- ... `reverseMap` ...
- Improve some geometry encoding by eliminating unnecessary boxing, using arrays where possible
- Produce fewer buffers to be concatenated when encoding metadata
- Call out several bugs, to be fixed separately:
  - Geometry encoding encodes stuff that never ends up being written
  - It also bases decisions about what to write on the sizes of things other than what's written, leading to incorrect choices
  - Geometry streams are written even if they're empty
  - Duplicates are not removed from Hilbert vertex dictionaries


java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/GeometryEncoder.java
```diff
@@ -91,422 +73,234 @@ public static EncodedGeometryColumn encodePretessellatedGeometryColumn(
     final var numGeometries = new ArrayList<Integer>();
     final var numParts = new ArrayList<Integer>();
     final var numRings = new ArrayList<Integer>();
-    final var numTriangles = new ArrayList<Integer>();
-    final var indexBuffer = new ArrayList<Integer>();
+    final var numTriangles = enableTessellation ? new ArrayList<Integer>() : null;
```
louwers, line 76: Java also has an Optional type. Consider using that.

louwers (approved):

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/GeometryEncoder.java
TimSylvester: I don't think it provides any benefit in this case.

PR #1346, opened 2026-04-25
chore(rust): new encoding methods

nyurik:


CommanderStorm (approved):

CommanderStorm:
since you are already touching the enc part, maybe fix the overflow the fuzzing found

CommanderStorm (approved):

PR #1348, opened 2026-04-26
refactor(rust): compute more stats before sort to enable deduplicating presence in v2

nyurik:
The goal is to compute these in addition to the existing string grouping stats:
* presence statistics
* min/max for integers
* in v2 - duplicate presence

### DIFF
--- a/rust/mlt-core/src/decoder/model.rs
+++ b/rust/mlt-core/src/decoder/model.rs
@@ -203,3 +203,33 @@ pub enum PropValue {
     F64(Option<f64>),
     Str(Option<String>),
 }
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PropKind {
+    Bool,
+    I8,
+    U8,
+    I32,
+    U32,
+    I64,
+    U64,
+    F32,
+    F64,
+    Str,
+}
+impl From<&PropValue> for PropKind {
+    fn from(prop: &PropValue) -> Self {
+        match prop {
+            PropValue::Bool(_) => Self::Bool,
+            PropValue::I8(_) => Self::I8,
+            PropValue::U8(_) => Self::U8,
+            PropValue::I32(_) => Self::I32,
+            PropValue::U32(_) => Self::U32,
+            PropValue::I64(_) => Self::I64,
+            PropValue::U64(_) => Self::U64,
+            PropValue::F32(_) => Self::F32,
+            PropValue::F64(_) => Self::F64,
+            PropValue::Str(_) => Self::Str,
+        }
+    }
+}
--- a/rust/mlt-core/src/decoder/model.rs
+++ b/rust/mlt-core/src/decoder/model.rs
@@ -204,7 +204,8 @@ pub enum PropValue {
     Str(Option<String>),
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::IntoStaticStr)]
+#[strum(serialize_all = "lowercase")]
 pub enum PropKind {
     Bool,
     I8,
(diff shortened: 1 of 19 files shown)


rust/mlt-core/src/encoder/optimizer.rs
```diff
@@ -104,3 +101,290 @@ impl TileLayer {
...
+                *self = Self::Unsigned {
+                    min: value,
+                    max: value,
+                };
+            }
+            Self::Unsigned { min, max } => {
+                *min = (*min).min(value);
+                *max = (*max).max(value);
+            }
+            _ => panic!("mixed property types are not allowed"),
```
CommanderStorm: I would rather we raise an error

rust/mlt-core/src/encoder/optimizer.rs
```diff
@@ -104,3 +101,290 @@ impl TileLayer {
...
+        max: u64,
+    },
+    Float,
+    String {
+        total_bytes: usize,
+        max_bytes: usize,
+        exact_hashes: Vec<u64>,
+        trigram_hashes: Vec<u64>,
+        shared_dict: SharedDictRole,
+    },
```
CommanderStorm, line 157: this is fairly weird modeling.
- trigram_hashes, exact_hashes should not be nessary after we have determined the grouping
- the properties like max_bytes, total_bytes seem to only be used in tests

CommanderStorm (commented):
This seems to be a bit undercooked and I am not entirely sure if this is working as intended

nyurik:
yep, sorry, still hacking on it

CommanderStorm:
It is a good idea 👍🏻

nyurik:
ok, i think i got it cleaned up a bit, sorry for the noise

rust/mlt-core/src/encoder/optimizer.rs
nyurik: interestingly enough, we were never validating that entire column is of the same type, and silently ignoring mismatches. Now there is an err

rust/mlt-core/src/encoder/optimizer.rs
nyurik: oops, thx

rust/mlt-core/src/encoder/optimizer.rs
```diff
@@ -104,3 +98,302 @@ impl TileLayer {
...
+        match self {
+            Self::String { shared_dict, .. } => shared_dict.clone(),
+            _ => SharedDictRole::None,
+        }
+    }
+
+    pub(crate) fn set_shared_dict(&mut self, role: SharedDictRole) {
+        match self {
+            Self::String { shared_dict, .. } => *shared_dict = role,
+            _ => debug_assert_eq!(role, SharedDictRole::None),
```
CommanderStorm, line 181: seems a bit weird, not sure

rust/mlt-core/src/decoder/model.rs
```diff
@@ -203,3 +203,34 @@ pub enum PropValue {
...
+    Bool,
+    I8,
+    U8,
+    I32,
+    U32,
+    I64,
+    U64,
+    F32,
+    F64,
+    Str,
```
CommanderStorm, line 219: supprising that this is not a dup..

CommanderStorm (approved):
LGTM.
Took me a bit to understand that this is not actually making the code faster, but only refactoring 👍🏻

rust/mlt-core/src/decoder/model.rs
nyurik: Yes, this is used to reason about the type of each column in terms of mvt, and to track if expansion or conversion is needed

rust/mlt-core/src/encoder/optimizer.rs
nyurik: Yeah, I'm also not happy about this part, will think about it in a bit

rust/mlt-core/src/decoder/model.rs
nyurik: Technically it is a bit faster because it should avoid a few extra sequence iterations, but likely not significant. I'm more interested in the v2 presence dedup

Issue #1353, opened 2026-04-28
Add .pmtiles output format to converter (Rust)


thomfuhrmann:
Hi,

I don't know about your plans to support the .pmtiles format in this repository. However, I wrote a small addition to the exisiting converter to support .pmtiles, since I needed it. I will link a PR to this issue.

nyurik:
this is definitely on our roadmap, thanks for the PR!

PR #1354, opened 2026-04-28
feat(rust): add .pmtiles output format to convert

thomfuhrmann:
Adds .pmtiles as another output format to the convert CLI.

Ref #1353


rust/mlt/src/convert/mod.rs
```diff
@@ -93,10 +116,12 @@ pub fn convert(args: &ConvertArgs) -> AnyResult<()> {
         ..Default::default()
     };
 
-    if is_mbtiles_extension(&args.input) {
-        if !is_mbtiles_extension(&args.output) {
+    if args.input_format() == Some(TileFormat::Mbtiles) {
+        if args.output_format() != Some(TileFormat::Mbtiles)
+            && args.output_format() != Some(TileFormat::Pmtiles)
+        {
```
CommanderStorm: this is a very confusing way of writing this

```suggestion
    if args.input_format() == Some(TileFormat::Mbtiles) && args.output_format() == None {
```

rust/mlt/src/convert/tileset.rs
```diff
@@ -31,11 +31,41 @@ pub async fn convert_mbtiles(
...
+
+pub async fn convert_tiles(args: &ConvertArgs) -> AnyResult<()> {
+    let cfg = EncoderConfig {
+        tessellate: args.tessellate,
+        try_spatial_morton_sort: matches!(args.sort, SortMode::Auto | SortMode::Morton),
+        try_spatial_hilbert_sort: matches!(args.sort, SortMode::Auto | SortMode::Hilbert),
+        try_id_sort: matches!(args.sort, SortMode::Auto | SortMode::Id),
+        allow_shared_dict: !args.no_shared_dict,
+        ..Default::default()
+    };
```
CommanderStorm: we now have this code twice.

Can we move this code to be `impl From<ConvertArgs> for EncoderConfig` instead?
Alternatively, passing it down is also fine, but I want to avoid bugs like `allow_shared_dict` not being passed..

rust/mlt/src/convert/tileset.rs
```diff
@@ -31,11 +31,41 @@ pub async fn convert_mbtiles(
...
-    drop(src_conn);
+    match (args.input_format(), args.output_format()) {
+        (Some(TileFormat::Mbtiles), Some(TileFormat::Mbtiles)) => {
+            convert_mbtiles_to_mbtiles(&args.input, &args.output, args.mbtiles_format.clone(), cfg)
+                .await?;
+        }
+        (Some(TileFormat::Mbtiles), Some(TileFormat::Pmtiles)) => {
+            convert_mbtiles_to_pmtiles(&args.input, &args.output, cfg).await?;
+        }
+        _ => bail!("Conversion formats not supported yet"),
```
CommanderStorm: lets be more explicit here

```suggestion
        (from, to) => bail!("Conversion from {from} to {to} is not currently implemented"),
```

rust/mlt/src/convert/mod.rs
```diff
@@ -22,6 +22,12 @@ fn whole_rate_per_sec(state: &ProgressState, w: &mut dyn std::fmt::Write) {
     let _ = w.write_fmt(format_args!("{}/s", state.per_sec() as u64));
 }
 
+#[derive(ValueEnum, Clone, Debug, PartialEq)]
+pub enum TileFormat {
+    Mbtiles,
+    Pmtiles,
+}
```
CommanderStorm: the value enum is not used.

If we change this from Option<TileFormat> to TileFormat, this would be the result.
I think it is likely a bit simpler to thread and reason about, or?

```suggestion
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum TileFormat {
    Mbtiles,
    Pmtiles,
    Files,
}
```

rust/mlt/src/convert/tileset.rs
```diff
@@ -67,3 +97,76 @@ pub async fn convert_mbtiles(
...
+    cfg: EncoderConfig,
+) -> AnyResult<()> {
+    let (raw_tx, raw_rx) = tokio::sync::mpsc::channel::<(TileCoord, Vec<u8>)>(500);
+    let (final_tx, mut final_rx) = tokio::sync::mpsc::channel::<(TileCoord, Bytes)>(500);
+
+    // Read from .mbtiles
+    let mbt = Mbtiles::new(input)?;
+    let mut conn = mbt.open_readonly().await?;
+    tokio::spawn(async move {
+        let mut stream = mbt.stream_tiles(&mut conn);
```
CommanderStorm:
```suggestion
        // FIXME: If the input is a normalised tile, we can gain a lot of performance by exploiting this fact as this tile then needs only be converted once
        let mut stream = mbt.stream_tiles(&mut conn);
```

rust/mlt/src/convert/tileset.rs
```diff
@@ -67,3 +97,76 @@ pub async fn convert_mbtiles(
...
+        .unwrap_or(1);
+    for _ in 0..num_cpus {
+        let input = raw_rx.clone();
+        let output = final_tx.clone();
+
+        tokio::spawn(async move {
+            while let Some((coord, data)) = {
+                let mut lock = input.lock().await;
+                lock.recv().await
+            } {
```
CommanderStorm: how about we use the mpmc from crossbeam here instead of the mutex, or is the mutex not a big performance hog?

rust/mlt/src/convert/tileset.rs
```diff
@@ -67,3 +97,76 @@ pub async fn convert_mbtiles(
...
+        serde_json::Value::String(Format::Mlt.metadata_format_value().into()),
+    );
+    let file = std::fs::File::create(output)?;
+    let metadata_str = serde_json::to_string(&metadata.tilejson)?;
+    let mut stream_writer = PmTilesWriter::new(TileType::Mlt)
+        .metadata(&metadata_str)
+        .create(file)?;
+    while let Some((coord, data)) = final_rx.recv().await {
+        stream_writer.add_tile(coord, &data)?;
+    }
```
CommanderStorm: does this correctly consume all tiles or do we need to await threads joining?

CommanderStorm (changes requested):
Mostly LGTM, I have a few cases where there might be gotchas or (easy) performance improvemnts.
I am fine with documenting those if you don't want to invest the time to fix them.

The main ask is to make sure that CI is happy with this change.
Otherwise, @nyurik has also asked for this, so exciting to see this getting added 🎉

thomfuhrmann:
Thanks for adding the improvements @CommanderStorm, I am fine with that.

CommanderStorm (approved):
Lets merge 😉

PR #1355, opened 2026-04-28
chore(rust): refactor low level encoding

nyurik:


rust/mlt-core/src/encoder/geometry/encode.rs
```diff
@@ -500,15 +370,17 @@ fn write_geo_precomputed_stream(
...
             let meta = StreamMeta::new2(stream_type, logical, phys, data.len())?;
             write_stream_payload(&mut enc.data, meta, false, vals)?;
+        } else if data.is_empty() {
+            let meta = StreamMeta::new2(ctx.stream_type, logical, PhysicalEncoding::None, 0)?;
+            write_stream_payload(&mut enc.data, meta, false, &[])?;
         } else {
-            let candidates = DataProfile::prune_candidates::<i32>(data);
             let mut alt = enc.try_alternatives();
-            for cand in candidates {
+            for physical_enc in [PhysicalEncoder::FastPFOR, PhysicalEncoder::VarInt] {
```
CommanderStorm, line 513: 🤔 
So just encoding FPF and varint is faster than doing a pruning pass first?

CommanderStorm (approved):

CommanderStorm:
+1 conditional on https://github.com/maplibre/maplibre-tile-spec/pull/1355#discussion_r3156533779 being a "yes, was intentional"

rust/mlt-core/src/encoder/geometry/encode.rs
nyurik: as explained - we were already doing just these two, but we did them several times, and now will do them just once each

PR #1356, opened 2026-04-28
chore(java): minor API improvements

TimSylvester:
Adds some convenience overloads and allows for nulls instead of empty lists for column mappings on non-struct columns.  I noticed the need for these while integrating #1176 in Planetiler.

### DIFF
--- a/java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/PropertyEncoder.java
+++ b/java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/PropertyEncoder.java
@@ -46,7 +46,7 @@ public class PropertyEncoder {
     final var estimatedBuffers = propertyColumns.size() * 5;
     final var featureScopedPropertyColumns = new ArrayList<byte[]>(estimatedBuffers);
 
-    final var columnMappingsIterator = (columnMappings != null) ? columnMappings.iterator() : null;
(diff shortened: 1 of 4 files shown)


CommanderStorm (approved):

louwers (approved):

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/PropertyEncoder.java
```diff
@@ -46,7 +46,7 @@ public static ArrayList<byte[]> encodePropertyColumns(
     final var estimatedBuffers = propertyColumns.size() * 5;
     final var featureScopedPropertyColumns = new ArrayList<byte[]>(estimatedBuffers);
 
-    final var columnMappingsIterator = columnMappings.iterator();
+    final var columnMappingsIterator = (columnMappings != null) ? columnMappings.iterator() : null;
```
louwers: Consider using the Optional type. https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Optional.html

TimSylvester: Yeah, I'm always torn on this.  It often produces clearer semantics, but also warnings recommending against it, in part because the optional itself can be null.

PR #1359, opened 2026-04-29
chore(rust): presence detection

nyurik:
* consolidate detection of identical presence across id/property space
* a few minor cleanups

### DIFF
--- a/rust/mlt-core/src/encoder/optimizer.rs
+++ b/rust/mlt-core/src/encoder/optimizer.rs
@@ -1,4 +1,4 @@
-use bitvec::prelude::{BitSlice, BitVec};
+use bitvec::vec::BitVec;
 
 use crate::decoder::{Morton, PropKind, TileLayer};
 use crate::encoder::model::{CurveParams, StagedLayer};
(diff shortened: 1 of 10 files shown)


rust/mlt-core/src/encoder/optimizer.rs
```diff
@@ -134,6 +142,10 @@ pub enum Presence {
     AllPresent,
     /// Some, but not all, features have a value for this logical column.
     Mixed,
+    /// Mixed presence with the same per-feature mask as the ID column.
+    SameAsId,
```
CommanderStorm: Just to talk motivation:
do we really want to allow this?

Would it not be simpler to just say "same as Prop N" on the Id stream?
I.e. a clear, one-way direction for pointers? currently, circles are a bit simpler to impl if not carefull..
Also, what about geometrys, can they not also have presence?

CommanderStorm (approved):
seems a bit messy with all of those unrelated changes, but sure.
I would like to know what the underlying idea is though, to not be left out..

rust/mlt-core/src/encoder/optimizer.rs
nyurik: * geometries (in v1) have no presence.  It has been requested on slack, but I still don't have a clear understanding of the use case, so not certain yet if it should be part of v2. It should probably be mentioned in #1201 to discuss
* Agree, `SameAsProp(idx)` is simpler here.  Note that at this point it is only used as hints to construct staged data
* The underlying idea is the realization that presence is frequently duplicated across properties, and thus we should introduce a shared presence dictionary, and reference it whenever more than one column shares it - see #1201 too. The simplest implementation is this:  each property and id columns in v2 can reference the shared presence table. We currently have 11 data types, so we can reserve 5 bits for the type (0..31), and use 3 bits for shared dictionaries - i.e. value 0 would mean "local" presence specific to current column, and 1..7 is one of the shared presence sequences that are used more than once.  Should be plenty, and we can easily pick out the most duplicated ones first if we exceed that number, and serialize others as if they are not duplicate.

nyurik: P.S. presence dedup may also help with v1 because it would be possible to just copy the output of the prior presence column.  The only problem is that we would need to move Id column to be AFTER properties.  Not sure how much it would save though.

PR #1360, opened 2026-04-30
fix(java): Only use space-filling curves in geometry encoding if the coordinate range is supported

TimSylvester:
- Add tests and max-level check for space-filling curves.
- Only consider Hilbert/Morton curves if the vertex span is within their supported range.


java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/GeometryEncoder.java
```diff
@@ -251,6 +279,21 @@ public static EncodedGeometryColumn encodeGeometryColumn(
         numStreams + 1, result, vertexLimits.max, geometryColumnSorted);
   }
 
+  private static <T extends Comparable<T>> boolean isLessThan(Optional<T> a, Optional<T> b) {
+    return isLessThan(a, b, Comparator.<T>naturalOrder());
+  }
+
+  /// Compare two Optional values, treating empty as greater than any present value
+  private static <T> boolean isLessThan(Optional<T> a, Optional<T> b, Comparator<T> comparator) {
```
louwers: Seems generic enough to live in its own util class.

louwers (approved):

PR #1363, opened 2026-05-01
ci(rust): per-tile size diff in the hotpath PR comment

CommanderStorm:
an experiment how we can report size changes in the hotpath based benches.


nyurik:
seems like it did not post anything?

CommanderStorm:
> seems like it did not post anything?

Yes, it is also not currently in main, so not expected to post something on second thought.
Let me open an PR against CommanderStorm/main and see what the output is of that

PR #1365, opened 2026-05-01
fix(java): Don't write empty streams

TimSylvester:
Stop writing empty streams, remove the corresponding Rust cases


rust/mlt-synthetics/src/layer.rs
```diff
@@ -192,6 +192,7 @@ impl Layer {
     /// Useful for producing byte-for-byte output that matches Java's encoder when a
     /// normally-empty stream must still appear in the wire format.
     #[must_use]
+    #[allow(dead_code)]
```
CommanderStorm:
```suggestion
    #[expect(dead_code, reason = "needs PR to impl that empty (but still present) property, gemetry and id streams don't cause issues")]
```

CommanderStorm (approved):

PR #1375, opened 2026-05-08
feat(rust): `mlt-core` ffi api for CPP, C, Kotlin

CommanderStorm:


nyurik (commented):
I have been thinking how we can mix two different ffi paradigms: zero cost vs marshalling. It sounds like Javascript and possibly Java do not allow foreign memory ownership,  where rust owns an object, and accessors are zero cost.  C/C++ on the other hand should allow this. Thus, it means how we structure c and c++ interface would be significantly different from the marshalled  ones - so it may make sense to separate two approaches entirely... what do you think?

CommanderStorm:
I removed the JS part, that needs a deeper rework upstream due to bugs. I think this works well.
Also I don't think that kotlin needs a different API..

release-plz.toml
```diff
@@ -31,6 +31,13 @@ name = "mlt"
...
 name = "mlt-core"
 
+[[package]]
+name = "mlt-ffi"
+
+[[package]]
+name = "mlt-core"
+
+
 [[package]]
```
CommanderStorm:
```suggestion
[[package]]
```

nyurik (approved):

rust/mlt-ffi/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+)]
+#[expect(
+    clippy::use_self,
+    reason = "Diplomat expands macros using the concrete type name rather than `Self`."
+)]
+#[diplomat::bridge]
+mod ffi {
+    use mlt_core::encoder::EncoderConfig;
+    use mlt_core::mvt::{mvt_to_tile_layers, tile_layers_to_mvt};
+    use mlt_core::{Decoder, Layer, MltError, Parser};
```
CommanderStorm:
```suggestion
    use mlt_core::{Decoder, Layer, Parser};
```

PR #1379, opened 2026-05-12
feat(java): Nested property values

TimSylvester:
- Adds support for JSON-style nested properties, including maps and lists with heterogeneous types.
- Adds a 0x02 type map, using that tag only when necessary.
- New synthetics, excluded from JS/Rust/C++ checks
- Refactors some encoding paths to reduce unnecessary boxing (more to do on that)
- Makes `Unsigned` classes `Comparable` for use in ordered containers
- Copilot-generated unit tests for un-covered property encoder cases, as a basis for ...
- Unit tests of round-trip encoding/decoding of various cases

The approach for encoding nested columns is:
- Find all the unique values in keys, values, lists, etc., grouped by encodable type (int, string, etc.)
- Store each key and value as an index into the combined list of sorted, unique values, flattening the structure with special values.
- Write a stream for each encodable type with values and integer streams for length and indexes

Multiple columns can share dictionaries, with the same number of streams.

Layers containing mixed properties are supported, with nested values for some features and not for others.

Potential improvements:
- It might make sense to separate integers into signed/unsigned and/or 32/64-bit, even at the cost of additional stream overhead, perhaps based on analysis of the data to be stored.  Currently they're all stored together, which will be inefficient in certain cases.
- Boolean values cost one byte.  They could theoretically be packed into the high bit of the previous value, at the cost of reducing the range limits of those values, which may or may not be a problem.  But use of varints makes this complex.
- This is not very efficient for small documents, so it might make sense to include a special "inline" encoding variant for when, e.g., a value stream would contain a single value.


java/mlt-core/src/main/java/org/maplibre/mlt/converter/MltConverter.java
```diff
@@ -785,4 +813,42 @@ private static MltMetadata.Column createColumn(
...
+
+    // Remove the original columns that are now merged into the map column.
+    mapColumns.forEach(entry -> columnSchemas.remove(entry.getKey()));
+
+    // Insert the merged column with a unique name.  The name is not encoded.
+    IntStream.iterate(0, i -> i + 1)
+        .mapToObj(i -> resolved.getLeft() + i)
+        .filter(name -> !columnSchemas.containsKey(name))
+        .findFirst()
+        .ifPresent(name -> columnSchemas.put(name, mapColumn));
```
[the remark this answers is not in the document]
TimSylvester: nonsense

rust/mlt-wasm/js/synthetic.spec.ts
```diff
@@ -27,6 +27,14 @@ const UNIMPLEMENTED_SYNTHETICS = new Map([
   ["poly_self_intersect_fpf", "FastPFor not supported"],
   ["poly_self_intersect_fpf_tes", "FastPFor not supported"],
   ["poly_multi_morton_hole_morton", "Pending investigation"],
+  ["0x01/prop_nested_big", "not implemented yet"],
+  ["0x01/prop_nested_json", "not implemented yet"],
+  ["0x01/prop_nested_list_root", "not implemented yet"],
+  ["0x01/prop_nested_list", "not implemented yet"],
```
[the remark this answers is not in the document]
TimSylvester: wrong, it doesn't work without the prefix, no idea about the others

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/MapPropertyEncoder.java
```diff
@@ -0,0 +1,550 @@
...
+    if (!uniqueValues.strings().isEmpty()) {
+      final var encoded =
+          StringEncoder.encode(uniqueValues.strings().keySet(), physicalLevelTechnique, useFSST);
+      numStreams += encoded.numStreams();
+      encodedStreams.add(new byte[] {(byte) encoded.numStreams()});
+      encodedStreams.addAll(encoded.encodedData());
+    }
+
+    if (!uniqueValues.ints().isEmpty()) {
+      if (uniqueValues.ints.keySet().stream().allMatch(x -> x < Integer.MAX_VALUE)) {
```
CommanderStorm:
```suggestion
      if (uniqueValues.ints.keySet().stream().allMatch(x -> Integer.MIN_VALUE < x && x < Integer.MAX_VALUE)) {
```

test/synthetic/0x02/prop_nested_big.json
CommanderStorm, line 1: I was under the impression that nested properties are a v2 feature, or?

Can we open an new directory for this?

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/MltTypeMap.java
```diff
@@ -1,12 +1,61 @@
 package org.maplibre.mlt.converter.encodings;
 
 import jakarta.annotation.Nullable;
-import java.util.List;
 import java.util.Optional;
 import org.maplibre.mlt.metadata.tileset.MltMetadata;
 
 public class MltTypeMap {
-  public static class Tag0x01 {
+  public static final class Tag0x02 extends Tag0x01 {
```
CommanderStorm, line 8: a bit akward, but 🤷🏻

CommanderStorm (approved):
LGTM

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/MapPropertyEncoder.java
TimSylvester: yeah, the comparisons should be inclusive, too

test/synthetic/0x02/prop_nested_big.json
TimSylvester: ah, I thought I was forgetting something

java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/MltTypeMap.java
TimSylvester: yes, but I decided to keep them entirely separate for now, I fully expect the values to change

PR #1391, opened 2026-05-19
fix(ts): geometry int encoding mismatch

### LINKED ISSUE #7659: Error when handling mlt generated from martin
**maplibre-gl-js version**: 5.24.0

**browser**: Safari 26.3.1, Chrome 146.0.7680.155 

### Steps to Trigger Behavior

 1. Using planetiler, generate pmtiles with mvt tiles (default tile_format)
 2. Configure martin to convert mvt->mlt
 3. Pass `Accept` header for martin's content negotiation
```ts
transformRequest: (url, resourceType) => {
  if (resourceType === 'Tile' && url.includes('areamap')) {
    return {
      url,
      headers: { Accept: 'application/vnd.maplibre-tile' },
    };
  }
  return { url };
},
```
4. Load page

Instead of rendering the layer, a console error is printed.

### Link to Demonstration

As @HarelM requested in slack, here are some example tiles. 
 - original: ~~[js_decode_tile_errors.zip](https://github.com/user-attachments/files/27969587/js_decode_tile_errors.zip)~~
- updated with mlt dump: [js_decode_tile_errors.zip](https://github.com/user-attachments/files/27976612/js_decode_tile_errors.zip), including uncompressed tiles per https://github.com/maplibre/martin/pull/2813

The source for these tiles was a pmtiles archive created with planetiler from https://download.bbbike.org/osm/bbbike/Bogota/Bogota.osm.pbf

## basecase: areamap-7-37-62-planetiler.mlt

This tile decodes  without error. It's served by martin directly from a pmtiles archive created with `--tile_format=mlt` (no dynamic conversion by martin).

Martin settings:
```
    sources:
      areamap-mlt:
        path: ../../data/Bogota/Bogota-mlt.pmtiles
```

## areamap-7-37-62-default-martin-settings.mlt

This tile errors. It's served by martin directly from a pmtiles archive with mvt tiles (converted to mlt on the fly by martin).

Console error:

> Unable to parse the tile at http://localhost:9000/tileserver/areamap-mlt/6/18/31, got error: Currently only optional string fields are implemented for a struct.

Martin settings:

    sources:
      areamap-mlt:
        path: ../../data/Bogota/Bogota-mvt-uncompressed.pmtiles
        convert_to_mlt: auto

I thought maybe this was related to maplibre-tile-spec/ts/src/mltDecoder.spec.ts:115:

    //TODO: fix -> since a change in the java converter shared dictionary encoding empty strings are not

## areamap-7-37-62-converted-with-no-shared-dict.mlt

Based on the aforementioned TODO, I tried disabling `shared_dict`, and now I get a new error:

> Unable to parse the tile at http://localhost:9000/tileserver/areamap-mlt/7/37/62, got error: The specified geometry type is currently not supported.

This is what the martin-config.yaml looks like:

    sources:
      areamap-mlt:
        path: ../../data/Bogota/Bogota-mvt-uncompressed.pmtiles
        convert_to_mlt:
          allow_shared_dict: false

michaelkirk:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/7659

ts has an optimization for "const" streams - those where every element in the stream has the same value. The "const" stream concept seems unique to the ts decoder (though java uses it on the encoder side).

Java uses isConst to determine if it should output an RLE encoded stream. For longer streams, it's obvious why RLE will be the winning encoding stream.

However, ts assumes that const stream will *always* be RLE encoded, presumably because that's what the java encoder emits.

But that seems to be an unsafe assumption, and is broken by the rust encoder. Rust doesn't have special handling for const streams, and simply races the approach against eachother. It's actually more efficient to delta encode a single element (or equally: to do NO logical encoding).

The actual difference in byte size here is tiny, and probably not relevant. But, the point is that the encoder could reasonably pick any of these strategies and the decoder needs to do the right thing.

Though this fixes the error I encountered, I'm not entirely confident that there aren't other similar ones. I'm new to the ts codebase and especially given the unique "const" stream concept in ts, it's a bit harder to reconcile with the rust codebase which I'm a little more familiar with.

---

I'm going to follow up with another PR (_update: https://github.com/maplibre/maplibre-tile-spec/pull/1392_) that defaults rust to use NO encoding in the cases where it's just as space efficient as delta encoding, since it should be more cpu efficient, and presumably a less surprising choice. Such a change also appears sufficient to avoid this same TS decode error, so I'm motivated to be able to produce artifacts that the older ts libraries can handle. To be clear, I still think we should merge this.


nyurik (approved):
seems reasonable, but I only looked at the rust code.  Thx for adding a new synthetics!

HarelM:
Thanks for looking into this!
Can you add a small unit test to cover this functionality?
I know that the synthetic covers this code path, but I would like to have a unit test as well so more information can be added to the test if needed. Also there's the co-pilot comment about the negative value, so maybe a unit test to cover this negative value scenario might be interesting here, IDK.
Overall, this looks good.
There might be other issues as you mention and we will cover them as we continue to improve this library.

ts/src/decoding/integerStreamDecoder.spec.ts
```diff
@@ -303,6 +303,25 @@ describe("decodeUnsignedConstInt32Stream", () => {
...
+    it("should ZigZag-decode a single DELTA-encoded value", () => {
+        const metadata = createStreamMetadata(LogicalLevelTechnique.DELTA, LogicalLevelTechnique.NONE, 1);
+        const data = encodeUnsignedInt32Stream(new Uint32Array([4]), metadata);
+
+        const result = decodeUnsignedConstInt32Stream(data, new IntWrapper(0), metadata);
+
+        expect(result).toBe(4);
+    });
+
+    it("decodes negative value, though I'm not sure the current behavior makes sense", () => {
```
HarelM: Let's avoid adding this test if we don't think it's the right approach. Thanks for taking the time to write it!

HarelM (approved):

PR #1392, opened 2026-05-19
fix(rust): integer encoding alternative: tie goes to None

michaelkirk:
When encoding a single integer, Delta and None were equally efficient. Because we tried Delta first it would win the `Alternatives` race. 

I think it's more intuitive to default to `None` in this case. It should also be a tiny be more efficient. We're talking tiny streams here, so I doubt it's measurable. 

My real motivation is that this works around https://github.com/maplibre/maplibre-gl-js/issues/7659, but it seems like a slightly more reasonable behavior anyway.

(For a more principled fix to https://github.com/maplibre/maplibre-gl-js/issues/7659, see https://github.com/maplibre/maplibre-tile-spec/pull/1391)

### DIFF
--- a/rust/mlt-core/src/encoder/stream/codecs.rs
+++ b/rust/mlt-core/src/encoder/stream/codecs.rs
@@ -189,9 +189,7 @@ impl Codecs {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::decoder::{LogicalEncoding, RawStream};
-    use crate::encoder::Encoder;
-    use crate::encoder::model::StreamCtx;
+    use crate::decoder::RawStream;
     use crate::test_helpers::{assert_empty, parser};
 
     #[test]
--- a/rust/mlt-core/src/encoder/stream/codecs.rs
+++ b/rust/mlt-core/src/encoder/stream/codecs.rs
@@ -143,6 +143,13 @@ impl Codecs {
         let Self { logical, physical } = self;
         let mut alt = enc.try_alternatives();
 
+        physical.write_alternatives::<Output<T>>(
+            &mut alt,
+            logical.none(values),
+            LE::None,
+            ctx.stream_type,
+        )?;
+
         let sample = logical.none(DataProfile::take_sample(values));
         let profile = DataProfile::profile::<<[T] as LogicalIntStreamKind>::Profile>(sample);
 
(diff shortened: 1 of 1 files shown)


michaelkirk:
CI failure https://github.com/maplibre/maplibre-tile-spec/actions/runs/26132753533/job/76861219922#step:4:88 seems unrelated...

The proximate cause is a build error while building `rustix`... 

```
  Compiling rustix v0.36.5
   Compiling semver v1.0.14
   Compiling serde_json v1.0.108
   Compiling quote v1.0.33
   Compiling linux-raw-sys v0.1.3
   Compiling bitflags v1.3.2
   Compiling thiserror v1.0.50
   Compiling syn v2.0.39
   Compiling camino v1.1.6
   Compiling current_platform v0.2.0
error: attributes starting with `rustc` are reserved for use by the `rustc` compiler
  --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-0.36.5/src/backend/linux_raw/io/errno.rs:28:25
   |
28 | #[cfg_attr(rustc_attrs, rustc_layout_scalar_valid_range_start(0xf001))]
   |     
```

This step is using rust +nightly. So maybe we can just try again tomorrow. 😬

nyurik:
I don't think we should try None as the first approach in general.  Instead, we should have a heuristics that uses different order of trying different alternatives depending on whatever criteria we have

michaelkirk:
maybe I am misunderstanding. I thought that all of these alternatives are being tried, and the one that performed the best was picked. The issue was that None and Delta were tied from a performance size perspective, and because delta was first, it was picked. I’ve only re-ordered them so that no-op gets the tie breaker.

nyurik:
you are correct - they are all tried, but ideally we should try the best one first to avoid extra copying - in other words if alt 1 is shorter than alt 2, alt 1 is written first, followed by alt 2, and later the 2nd one is just discarded. If alt 1 is actually longer, than first we write alt1, followed by alt2, and later alt2 is moved in-place to the alt1 location.  So you are totally correct, but the fewer moves we have, the better.

michaelkirk:
I see. I think it's ultimately a special case (single item) that I'm trying to accommodate. It's probably not worth slowing down the normal path. I'll see if I can come up with something reasonable tomorrow.

michaelkirk:
After brainstorming a couple of approaches, I've decided it's not worth complicating the code, and am going to close this. Maybe someone else will come up with a better approach in the future.

PR #1411, opened 2026-06-03
feat(rust): pmtiles to pmtiles transcoding

CommanderStorm:


rust/mlt/src/convert/common.rs
```diff
@@ -0,0 +1,111 @@
...
+    data: &[u8],
+    encoding: Encoding,
+    cfg: EncoderConfig,
+) -> AnyResult<(Bytes, bool)> {
+    if data.len() > MAX_TILE_CACHE_TRACK_SIZE_BYTES {
+        return Ok((encode_one(data.to_vec(), encoding, cfg)?, false));
+    }
+    let mut hit = true;
+    let encoded = cache
+        .try_get_with(xxh3_64(data), || {
```
nyurik: do you rely on the hash of the tile data to be the key of that data and to be unique? This is never guaranteed, even though it has 64bits - there could still be a conflict with "relatively" small chances. Should we store some additional identity in the cache?

nyurik (approved):

PR #1413, opened 2026-06-04
feat(rust): fast-mvt crate

nyurik:
work in progress - a much faster MVT parser.  Gains 157% over mvt-reader


rust/fast-mvt/src/types.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct MvtTile {
+    pub layers: Vec<MvtLayer>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct MvtLayer {
+    pub name: String,
+    pub extent: u32,
```
CommanderStorm: how about this and we remove the error structurally?

```suggestion
    pub extent: NonZeroU32,
```

rust/fast-mvt/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+| Decoder      |    Time |     Compare |
+|--------------|--------:|------------:|
+| `fast-mvt`   |  453 ms |           - |
+| `mvt-reader` | 1165 ms | 157% slower |
+
+| Encoder    | Time | Compare |
+|------------|-----:|--------:|
+| `fast-mvt` |  TBD |         |
+| `mvt`      |  TBD |         |
```
CommanderStorm: since we have the bench for this, we just need to run the bench 😉

rust/fast-mvt/src/writer.rs
```diff
@@ -0,0 +1,220 @@
...
+
+impl TagsBuilder {
+    fn new() -> Self {
+        Self {
+            keys: DupIndexerRefs::new(),
+            values: DupIndexer::new(),
+        }
+    }
+
+    fn insert(&mut self, key: &str, value: MvtValue) -> MvtResult<(u32, u32)> {
```
CommanderStorm:
```suggestion
    /// Inserts the KV pair into the key and value dictionary and returns the index where it was inserted
    fn insert(&mut self, key: &str, value: MvtValue) -> MvtResult<(u32, u32)> {
```

rust/fast-mvt/justfile
CommanderStorm, line 1: a lot of duplication here compard to the main justfile.
Please de-dup

rust/fast-mvt/build.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(feature = "codegen")]
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    buffa_build::Config::new()
+        .files(&["src/vector_tile.proto"])
+        .includes(&["src"])
+        .out_dir("src/generated")
+        .include_file("mod.rs")
+        .generate_json(true)
+        .gate_impls_on_crate_features(true)
```
CommanderStorm, line 13: I think, we don't need to preserve unknown fields

```suggestion
        .gate_impls_on_crate_features(true)
        .preserve_unknown_fields(false)
```

rust/fast-mvt/build.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(feature = "codegen")]
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    buffa_build::Config::new()
+        .files(&["src/vector_tile.proto"])
+        .includes(&["src"])
+        .out_dir("src/generated")
+        .include_file("mod.rs")
+        .generate_json(true)
```
CommanderStorm, line 11: if we are already optimising for perf, we might as well do a very quick bench if german strings (https://cedardb.com/blog/german_strings/) speed up this.
We use them in a few places and anecdotally, they speed things along quite nicely.

```
.string_type(repr)	String	Use an alternative owned representation (SmolStr, EcoString, CompactString) for all string fields (see [String field representations](https://github.com/anthropics/buffa/blob/main/docs/guide.md#string-field-representations))
.string_type_in(repr, &[...])	—	Use an alternative string representation for matching string fields
```

rust/fast-mvt/README.md
CommanderStorm, line 1: please add a table where all the `cfg`-features are explained.
Otherwise not very obvious what this `view` cfg does.

rust/fast-mvt/build.rs
CommanderStorm, line 13: this and adding an `arbitrary` feature allows users to use fuzzing, which I think I have shown to be usefull

```suggestion
        .gate_impls_on_crate_features(true)
        .generate_arbitrary(true)
```

rust/fast-mvt/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+mod geometry;
+
+#[allow(
+    clippy::derivable_impls,
+    clippy::pedantic,
+    clippy::upper_case_acronyms,
+    clippy::use_self,
+    dead_code,
+    unused_qualifications
+)]
```
CommanderStorm, line 24: should we expect this?

```suggestion
#[expect(
    clippy::derivable_impls,
    clippy::pedantic,
    clippy::upper_case_acronyms,
    clippy::use_self,
    dead_code,
    unused_qualifications
)]
```

rust/fast-mvt/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+- `reader` (default): decode MVT bytes, inspect layers/features/tags values.
+  The reader API is view-backed: `MvtReaderRef<'a>` borrows from caller-provided
+  bytes, so the bytes must outlive the reader.
+- `writer` (default): encode owned `Tile`/`Layer`/`Feature` values and build
+  MVT key dictionaries with `dup-indexer`.
+- `json`: enable serde JSON support for the generated protobuf bindings.
+- `views`: internal features that should not be used by external callers.
+- `codegen`: regenerate protobuf bindings from `src/vector_tile.proto`.
+  The generated file is checked in, so normal builds do not require `protoc`.
```
CommanderStorm: Should we add an `autofix` that runs codegen?
This way we if an new version of buffa improves things, we get those in the respective PR.

rust/fast-mvt/src/proto.rs
```diff
@@ -0,0 +1,45 @@
...
+        tile
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::Tile;
+
+    #[test]
+    fn tile_deserializes_from_object_but_not_layers_array() {
```
CommanderStorm, line 23: I think this test needs to be cfg'ed on the `json` feature

rust/fast-mvt/src/reader.rs
```diff
@@ -0,0 +1,483 @@
...
+    pub fn to_proto(&self) -> crate::generated::vector_tile::Tile {
+        self.0.to_owned_message()
+    }
+}
+
+fn validate_layers(tile: &TileView<'_>) -> MvtResult<()> {
+    for layer in &tile.layers {
+        if layer.name.is_empty() {
+            return Err(MvtError::MissingLayerName);
+        }
```
CommanderStorm, line 49: I am 50-50 that this should be "Empty..", but protobuf I think does not differnetiate.. meh

CommanderStorm (approved):
I found a few cases where we might be able to increase perf, otherwise LGTM.

given the size of the PR and test suite, the geometry part did not get an as-thourough review as it could have from me.

michaelkirk (commented):
It looks like this is currently unused code. I assume the idea is that this new crate will eventually replace the usage of `mvt-reader`. Is that right?

nyurik:
correct - the goal is to use this crate for all MLT and Martin needs going forward

nyurik:
I ended up publishing it from my own repo, as adding it to this repo was a bit too complex with all the unsafe, workspace, test integration, etc.  I am adding @CommanderStorm as a maintainer to it, but eventually I do plan to move it to other home - either maplibre or geo orgs - TBD

https://crates.io/crates/fast-mvt

PR #1416, opened 2026-06-06
feat(py): simple, geojson based encoding

CommanderStorm:
`encode(layer) -> bytes` encodes a single layer to an MLT blob. Geometry is in tile-local coordinate space (no projection), matching tilezen/mapbox-vector-tile's default. Coordinates must be integer-valued and 2D.

Used claude for the codegen part.


rust/mlt-py/maplibre_tiles.pyi
```diff
@@ -57,6 +58,16 @@ def decode_mlt_to_geojson(data: bytes) -> builtins.str:
     Decode an MLT binary blob and return GeoJSON as a string.
     """
 
+def encode(geojson: typing.Mapping[builtins.str, builtins.object], name: builtins.str, extent: builtins.int = 4096) -> bytes:
```
tobi-stephen: suggestion: rename method name as `encode_geojson` as we discussed

rust/mlt-py/README.md
```diff
@@ -24,3 +24,42 @@ geojson_str = maplibre_tiles.decode_mlt_to_geojson(data)
...
+                "geometry": {"type": "Point", "coordinates": [2048, 1024]},
+                "properties": {"name": "main", "lanes": 3},
+            },
+        ],
+    },
+    name="roads",
+    extent=4096,
+)
+
+# Multi-layer tiles: encode each layer and concatenate the bytes
```
tobi-stephen: I'm curious to know the result of decoding the multi-layer tile blob.
like this: `maplibre_tiles.decode(multi_layer_mlt_blob)`

CommanderStorm: What are you curious about?
Is the API for that (see above in this file) good?

That API was mostly done for qgis, so may not be fitting your usecase entirely.
Without entirely knowing what your usecase is more precicely, that is challenging to know 😉

tobi-stephen: my usecase: I have mvt files(saved as pbf) that I want to transform to mlt files for evaluating it's benefit.
the mvts are used by a mobile app using MapLibre Native on IOS.
Note, I have used the Java 'encoding server' for on-the-fly MLT transform and it worked on the mobile app, although, I do have some pbf files that are not being transformed, but they are mostly very custom MVT tiles with unsupported types, I guess - not an issue right now

I'm just asking to see know if the MLT multi-tile-blob generated here be read successfully by MapLibre Native

rust/mlt-py/src/encode/geojson.rs
tobi-stephen, line 1: should the source file and test file also be named `encode_geojson` or can the present encode file be extended with the `encode_layers` or `encode_tiles` later

rust/mlt-py/README.md
CommanderStorm: Then a simpler (and more performant) API would be `encode_mvt(bytes, options)`.
This API is likely still something we eventually need to build out, given that it gives lower level control

tobi-stephen: Yeah, I agree and it will allow for easier portability of tiles from MVT-MLT with python.

rust/mlt-py/src/encode/geojson.rs
CommanderStorm: true, can be build a bit more "futureproof" ^^

nyurik (commented):
shouldn't we use geojson crate?

CommanderStorm:
For what? if we are going to decode into `mlt_core::geojson::FeatureCollection` anyhow, there does not really seem a good point..

nyurik (approved):

PR #1417, opened 2026-06-07
fix(rust): accept .pbf in convert

tobi-stephen:
Given that MVTs are saved as either MVT/PBFs, but the rust's command for conversion mlt convert <in_dir> <out_dir> only expects the dir to contain MVT file format alone, I propose updating the rust command to read PBFs too.


CommanderStorm (approved):

CommanderStorm:
Thanks 👍🏻

PR #1420, opened 2026-06-07
chore(deps): pin all ci versions

CommanderStorm:
We are currently not pinning all our ci versions.
We should, so that we know which software we are running in CI and so our supply chain checks can be effective.


nyurik (approved):
did you use some tool, or just ai?

CommanderStorm:
@nyurik do you really think I would trust my LLM of choice with this? That sounds tremendously stupid to do..

https://github.com/suzuki-shunsuke/pinact

PR #1423, opened 2026-06-08
chore(rust): enforce non-empty layer name

nyurik:
ai-gened enforcement that ensures that we do not have zero-length layer names.  MVT spec has it non-explicitly, but it seems some implementations actually enforce this, so easier to make it explicit


CommanderStorm (approved):

tobi-stephen (approved):

TimSylvester:
I would suggest that encoders should reject such layers with an error, but decoders should ignore them with at most a warning, allowing other layers in the same tile to work.

michaelkirk:
It looks like @TimSylvester was ignored - error on decode is still being thrown. Was that intentional @nyurik?

PR #1427, opened 2026-06-08
feat(py): make `mlt.encode_geojson` configurable

CommanderStorm:
Testing side is a bit weak, but idk how to improve here without being very brittle..


tobi-stephen (approved):
Thanks for adding this. This should unblock an issue I'm having with maplibre-native test app because of the `FastPFOR` option set to True and not configurable from the python lib.

MLT parsing was failing, so I needed to "turn on" FPF and recompile the maplibre-native app for the test glfw_app.

rust/mlt-py/maplibre_tiles.pyi
```diff
@@ -58,13 +58,19 @@ def decode_mlt_to_geojson(data: bytes) -> builtins.str:
...
 
     `geojson` is an RFC 7946 `FeatureCollection`.
     `name` and `extent` set the MLT layer metadata, since a `FeatureCollection` has no slot for them.
     Geometry is in tile-local coordinate space (no projection).
+
+    `tessellate` generates triangulation data for polygons and multi-polygons.
+    `sort` chooses which feature ordering(s) the encoder trials: `auto` tries all, a named curve (`morton`/`hilbert`/`id`) tries just that one, and `none` keeps the input order.
+    `shared_dict` allows grouping strings into shared dictionaries.
+    `fsst` allows FSST string compression.
+    `fpf` allows FastPFOR integer compression.
```
tobi-stephen: I looked through the Rust backend for the "FastPFOR" flag, and it seems like it will always be true, despite this configuration

Probably might need to make this config value flow to the Rust backend and apply it, my guess is it should be write_alternative() function in rust/mlt-core/src/encoder/stream/write.rs

rust/mlt-py/src/encode/shared.rs
```diff
@@ -0,0 +1,42 @@
...
+/// `sort` collapses the three sort-strategy toggles into one choice, mirroring the `mlt convert` CLI.
+pub(crate) fn encoder_config(
+    tessellate: bool,
+    sort: &str,
+    shared_dict: bool,
+    fsst: bool,
+    fpf: bool,
+) -> PyResult<EncoderConfig> {
+    let (morton, hilbert, id) = match sort {
+        "auto" => (true, true, true),
```
nyurik: i thought we didn't want some of these on auto because they are not producing meaningful savings but take a lot longer?

nyurik (approved):
check inline

rust/mlt-py/src/encode/shared.rs
CommanderStorm: true

```suggestion
        "auto" => (true, false, false),
```

rust/mlt-py/src/encode/shared.rs
```diff
@@ -0,0 +1,42 @@
...
+    shared_dict: bool,
+    fsst: bool,
+    fpf: bool,
+) -> PyResult<EncoderConfig> {
+    let (morton, hilbert, id) = match sort {
+        "auto" => (true, false, false),
+        "morton" => (true, false, false),
+        "hilbert" => (false, true, false),
+        "id" => (false, false, true),
+        "none" => (false, false, false),
```
CommanderStorm, line 26:
```suggestion
    let (morton, hilbert, id) = match sort {
        "all" => (true, true, true),
        "auto" | "morton" => (true, false, false),
        "hilbert" => (false, true, false),
        "id" => (false, false, true),
        "none" => (false, false, false),
```

rust/mlt-py/src/encode/geojson.rs
```diff
@@ -11,31 +11,49 @@
...
+)]
 pub fn encode_geojson(
     py: Python<'_>,
     #[gen_stub(override_type(type_repr = "typing.Mapping[builtins.str, builtins.object]"))]
     geojson: &Bound<'_, PyAny>,
     name: String,
     extent: u32,
+    tessellate: bool,
+    #[gen_stub(override_type(
+        type_repr = "typing.Literal['auto', 'morton', 'hilbert', 'id', 'none']"
```
CommanderStorm:
```suggestion
        type_repr = "typing.Literal['all', 'auto', 'morton', 'hilbert', 'id', 'none']"
```

rust/mlt-py/src/encode/geojson.rs
```diff
@@ -11,31 +11,49 @@
...
+use super::shared::{encoder_config, val_err};
+
 /// Encode a GeoJSON `FeatureCollection` into MLT bytes.
 ///
 /// `geojson` is an RFC 7946 `FeatureCollection`.
 /// `name` and `extent` set the MLT layer metadata, since a `FeatureCollection` has no slot for them.
 /// Geometry is in tile-local coordinate space (no projection).
+///
+/// `tessellate` generates triangulation data for polygons and multi-polygons.
+/// `sort` chooses which feature ordering(s) the encoder trials: `auto` tries all, a named curve (`morton`/`hilbert`/`id`) tries just that one, and `none` keeps the input order.
```
CommanderStorm:
```suggestion
/// `sort` chooses which feature ordering(s) the encoder trials: `all` tries all orderings, `auto` tries a subset with a good speed-size tradeoff, a named curve (`morton`/`hilbert`/`id`) tries just that one, and `none` keeps the input order.
```

rust/mlt-py/src/encode/mvt.rs
CommanderStorm:
```suggestion
                out.extend_from_slice(&tile.encode(cfg)?);
```

CommanderStorm:
```suggestion
```

rust/mlt-py/src/encode/mvt.rs
```diff
@@ -1,28 +1,39 @@
...
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 use pyo3::types::PyBytes;
 use pyo3_stub_gen::derive::gen_stub_pyfunction;
 
+use super::shared::encoder_config;
+
 /// Encode an entire MVT tile to MLT using default encoding options.
 ///
 /// `data` is a raw Mapbox Vector Tile (protobuf).
```
CommanderStorm, line 14:
```suggestion
/// `data` is a raw Mapbox Vector Tile (protobuf).
/// 
/// Options:
/// `tessellate` generates triangulation data for polygons and multi-polygons.
/// `sort` chooses which feature ordering(s) the encoder trials: `all` tries all orderings, `auto` tries a subset with a good speed-size tradeoff, a named curve (`morton`/`hilbert`/`id`) tries just that one, and `none` keeps the input order.
/// `shared_dict` allows grouping strings into shared dictionaries.
/// `fsst` allows FSST string compression.
/// `fastpfor` allows FastPFOR integer compression.
```

PR #1428, opened 2026-06-09
chore(rust): rm pub fields, builder pattern

nyurik:


rust/mlt-core/src/decoder/model.rs
```diff
@@ -1,9 +1,37 @@
...
 use crate::decoder::{Geometry, Id, Property};
-use crate::{DecodeState, Lazy, Parsed};
+use crate::{DecodeState, Lazy, MltError, MltResult, Parsed};
+
+/// Non-zero tile extent.
+///
+/// Use [`Extent::new`] to validate raw integer input before storing it in
+/// owned row or staged layer structures.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Extent(NonZeroU32);
```
CommanderStorm, line 14: We still have quite a few `extent: u32` laying around.
I would prefer if we fully do the migration here, but I can also do an PR

rust/mlt-core/src/encoder/model.rs
```diff
@@ -90,6 +179,85 @@ impl Default for EncoderConfig {
...
 }
 
+impl EncoderConfig {
+    #[must_use]
+    pub fn tessellate(self) -> bool {
+        self.tessellate
+    }
+
+    #[must_use]
+    pub fn try_spatial_morton_sort(self) -> bool {
```
CommanderStorm: nit: if we are going for fully pub api, lets not name it `try_..` since that usually implies faliability.


how about something synonymous like `attempt_sort_by_spatial_morton`

CommanderStorm (approved):

rust/mlt-core/src/decoder/model.rs
nyurik: @CommanderStorm i don't see any structs that have `extent: u32` - other than function params and error states.  Should we make Extent be an external param, or should we only use it internally for now?

rust/mlt-core/src/encoder/model.rs
nyurik: technically all EncoderConfig params are "attempt" - e.g. `attempt_fpf`... so maybe just remove the `attempt` everywhere?   How about this API:

```rust
use_fastpfor(self) -> bool
use_fsst(self) -> bool
use_id_sort(self) -> bool
use_shared_dict(self) -> bool
use_spatial_hilbert_sort(self) -> bool
use_spatial_morton_sort(self) -> bool
use_tessellation(self) -> bool

with_fastpfor(mut self, enabled: bool) -> Self
with_fsst(mut self, enabled: bool) -> Self
with_id_sort(mut self, enabled: bool) -> Self
with_shared_dict(mut self, enabled: bool) -> Self
with_spatial_hilbert_sort(mut self, enabled: bool) -> Self
with_spatial_morton_sort(mut self, enabled: bool) -> Self
with_tessellation(mut self, enabled: bool) -> Self
```

CommanderStorm: sure

rust/mlt-core/src/decoder/model.rs
CommanderStorm: Yes, function params and errror state.
I would like them to carry forward the `NonZeroU32` restriction so that our types are as tight as they can be 😉 

I can do an followup PR if you prefer this as said

rust/mlt-core/src/encoder/model.rs
```diff
@@ -186,17 +186,17 @@ impl EncoderConfig {
     }
 
     #[must_use]
-    pub fn try_spatial_morton_sort(self) -> bool {
+    pub fn attempt_spatial_morton_sort(self) -> bool {
         self.try_spatial_morton_sort
```
michaelkirk: Is there a reason to rename the outter but not the inner? (This comment applies to several places in this file)

CommanderStorm:
simpler to maintain in main, enabling automerge

PR #1429, opened 2026-06-10
fix(java): correct fastpfor config option to enable-fastpfor

tobi-stephen:
From the encoder-server, this fix adjusts the `fastpfor` option to map correctly to the expected `enable-fastpfor` cli flag.


CommanderStorm (approved):
Thanks

PR #1432, opened 2026-06-13
fix(rust): clear losing sort-trial buffers between encode trials

CommanderStorm:
Previously, this bug means we encode Plain way too often..

### DIFF
--- a/rust/mlt-core/src/encoder/writer.rs
+++ b/rust/mlt-core/src/encoder/writer.rs
@@ -265,6 +265,21 @@ impl Encoder {
         self.hdr.len() + self.meta.len() + self.data.len()
     }
 
+    /// Empty the output buffers (`hdr`/`meta`/`data`) so this encoder can be
+    /// reused for the next sort trial, keeping their allocated capacity and the
+    /// seeded curve/FSST caches.
+    ///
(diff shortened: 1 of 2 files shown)


rust/mlt-core/src/encoder/writer.rs
```diff
@@ -265,6 +265,21 @@ impl Encoder {
...
+    /// Empty the output buffers (`hdr`/`meta`/`data`) so this encoder can be
+    /// reused for the next sort trial, keeping their allocated capacity and the
+    /// seeded curve/FSST caches.
+    ///
+    /// Unlike [`Self::preserve_results`] (which moves the buffers out into the
+    /// kept "best" result), this is used when a trial loses: its bytes must be
+    /// discarded, otherwise the next trial's `encode_into` would append to them
+    /// and over-count its `total_len`.
+    pub(crate) fn clear_results(&mut self) {
+        debug_assert!(self.alt_stack.is_empty(), "Alternatives stack is not empty");
```
CommanderStorm, line 277: Can likely also be done using Phantomdata, but this is an well run part of the codebase, so I did not do this.

nyurik (approved):

PR #1433, opened 2026-06-13
perf(rust): use mimalloc as the global allocator for the mlt CLI

CommanderStorm:
gains us 1-1.5 cpu cores spend idling for a bit higher RSS.
Likely a good trade

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -3281,6 +3281,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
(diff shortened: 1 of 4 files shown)


nyurik (approved):
interesting... should we make this conditional?  i.e. have a feature that enables it only when a feature is specified?  Or since this is mlt bin only, its not worth it?  At least this could allow us an easier way to run benchmark comparisons in the future

CommanderStorm:
I don't think this is one of the main areas of churn. mimaloc is pretty good for now I think, overhead of alloc on cpu cycels is not the main thing now, I think.

Maybe in the future again, we will see.
On the few cases where the allocator mattered, I was IO/Memory bound.

nyurik:
I'm confused - if its about the same perf as the one built in, why both with a custom builder, which requires longer C-based compilation

PR #1434, opened 2026-06-13
refactor(rust): split convert into from_files/from_mbtiles/common and add --sort all

CommanderStorm:
This removes Hilbert and ID from the auto option as they together only gain <0.5% in size for >30% in runtime, so likely not in the auto set


nyurik (approved):
i was contemplating if https://crates.io/crates/bitflags would be a good fit, but probably not worth it

PR #1446, opened 2026-06-18
fix(rust): `allow_fastpfor` not consulted for precomputed geometry stream

michaelkirk:
Before 0110faf2 I saw the error:

 > [MapLibre] [event]:ParseTile [code]:-1 [message]:MLT parse failed: FastPFOR decoding is not enabled. Configure with MLT_WITH_FASTPFOR=ON                                                                              

After 0110faf2 it becomes:

 > [MapLibre] [event]:ParseTile [code]:-1 [message]:MLT parse failed: FastPfor encoding for geometries is not yet supported.    

This fixes the other place we were using fastpfor


CommanderStorm (approved):
I thought I already fixed this. ups

Issue #1387, opened 2026-05-16
List example datasets stored/exposed as MLT and ongoing integrations


jo-chemla:
Following the MLT announcement a few months back, I was curious to know how one can currently produce and consume MLT given the great perf/feature-set allowed over the established MVT. 

I saw that the [implementation status](https://maplibre.org/maplibre-tile-spec/implementation-status/) doc page lists 
 - existing implementations of encoders/decoders
 - as well integrations. Apart from the pure maplibre ecosystem with gl-js and native being able to render MLT, great to know `planetiler` can produce them, `pmtiles` can store them (although I'm not sure a tool exists to produce `MLT-into-pmtiles` dataset), and `martin` can watch them.

Apart from the quickfixes listed below, I wanted to reach out to know if it would make sense to list on that doc page, or add a meta tracker thread in a repo issue: 
 - some existing/upcoming datasets like:
   - [maplibre/demotiles](https://github.com/maplibre/demotiles) at `https://demotiles.maplibre.org/tiles-mlt/plain/{z}/{x}/{y}.mlt` or tilejson
   - Overture consideration for switching to MLT at https://github.com/OvertureMaps/overture-tiles/issues/50
   - [shademap](https://ted-piotrowski.github.io/mapbox-gl-shadow-simulator/examples/maplibre.html) buildings and planet `https://cfw.shademap.app/planet/{z}/{x}/{y}.mlt` and `https://cfw.shademap.app/buildings/{z}/{x}/{y}.mlt`
   - or OpenRailwayMap https://github.com/hiddewie/OpenRailwayMap-vector/issues/919 

 - as well as ongoing work to add support for MLT - for people to know these efforts are ongoing, and eventually take part. Could list 
   - `felt/tippecanoe` support soon landing thanks to the C++ encoder https://github.com/felt/tippecanoe/issues/380, 
   - `systemed/tilemaker` also probably soon getting support thanks to the C++ encoderat https://github.com/systemed/tilemaker/issues/856 
   - `georust/geozero` framework was awaiting rust encoder https://github.com/georust/geozero/issues/227 
   - `developmentseed/tipg` waiting for ST_AsMLT within PostGIS at https://github.com/developmentseed/tipg/issues/248
   - `versatiles-org/versatiles-rs` rust tiler tracking support in https://github.com/versatiles-org/versatiles-rs/issues/111 and https://github.com/versatiles-org/versatiles-spec/issues/11
[...]

PR #1448, opened 2026-06-18
implementation-status: add upcmoing integrations and open-data datasets

jo-chemla:
See discussion thread https://github.com/maplibre/maplibre-tile-spec/issues/1387#issuecomment-4741327752


CommanderStorm (approved):
thanks

PR #1450, opened 2026-06-19
fix(cpp,ts,syntetics): single value MLT integers in non-plain formats

CommanderStorm:
adds synthetics and makes sure they parse correctly.

Motivation is primarily that `ts` doubled my values, which was not so pleasant.
cpp also did this wrong, but threw an exepction instead.

CC @michaelkirk mostly because you might come across this bug too


ts/src/decoding/integerStreamDecoder.ts
```diff
@@ -209,6 +209,9 @@ export function decodeUnsignedConstInt64Stream(
     const values = decodeVarintInt64(data, offset, streamMetadata.numValues);
 
     if (values.length === 1) {
+        if (streamMetadata.logicalLevelTechnique1 === LogicalLevelTechnique.DELTA) {
+            return decodeZigZagInt64Value(values[0]);
+        }
```
CommanderStorm, line 214: this is nasty, because not having it doubles my values..

HarelM: Not sure what you mean by this comment, but I don't see an issue with this fix.
It would be great to have a unit test to cover this scenario as well, it shouldn't be hard to add one to the existing.

CommanderStorm: The syntetics that I added cover EXACTLY this (plus a few more) cases.

One item, DELTA logical enc makes 💥 in both deocoders.

CommanderStorm: and the java ancoder cannot encode all, because it chooses the "smart" option instead of what I ask him to encode, but that is not a new phenomenon.

CommanderStorm: If you want a non-synthetic test, here is a real world tile that gets decoded wrong

```bash
curl -s -H 'Accept: application/vnd.maplibre-tile' https://nav.tum.de/martin/events_upcoming/13/4359/2842 -o /tmp/ev.mlt
```

HarelM: Just a simple unit test, it's clearer what part it tests. Synthetic are more "end to end", which is important but it's easier to understand unit tests. 
If it's a big ask I can help add that test.

CommanderStorm: I guess, I added another unit test.

This is firmly tested already through the syntehtics already though, so I don't see the point of that test

HarelM (approved):
Typescript part is approved from my point of view. Thanks!
Feel free to create a bump version PR and release a new patch version of the typescript decoder, it will be automatically picked up by maplibre-gl-js dependabot process.

PR #1451, opened 2026-06-19
test: make sure that marked-failing synthetics actually fail

CommanderStorm:
Currently, when a synthetic test passes we would not know.
This happened to MLT-Wasm, so lets assert that this does not happen.


michaelkirk (approved):

michaelkirk:
I was confused by the PR title, but essentially: 

If a known failing synthetic starts newly passing, we want to move it from UNSUPPORTED to supported. Otherwise we might accidentally regress it later.

CommanderStorm:
Yea, reworded it

PR #1456, opened 2026-06-19
Enforce TypeScript formatting with Biome

### LINKED ISSUE #842: Enable automatic precommit-ci TypeScript code formatting
Current /ts code is not being auto-formatted, often resulting in unstably-formatted code to be merged. We have been using https://github.com/maplibre/maplibre-tile-spec/blob/main/.pre-commit-config.yaml to automate this in other languages. It seems pretier one has been archived, so we may need to either pick a different formatter, or use some other method to make sure the repo code is clean.

AlexChakmakian:
This builds on the existing Biome setup for the TypeScript code.

I found that 'just ts::fmt' was calling 'npm run format', but that script did not exist yet. Also, formatting was only handled through pre-commit, so it was still possible for unformatted TypeScript code to get merged if pre-commit was not installed locally.

This PR adds Biome as a TypeScript dev dependency, adds format scripts, and makes the TypeScript lint step check formatting too.

Closes #842.

**Testing I did**

Ran locally:

- 'npm ci'
- 'npm run format:check'
- 'npm run format'
- 'npm run lint'
- 'npm run test'
- 'npm run build'

I also temporarily added badly formatted TypeScript and confirmed 'npm run format:check' failed, then restored the file.


CommanderStorm:
Can we add either an [autofix](https://github.com/maplibre/maplibre-tile-spec/blob/main/.github/workflows/autofix.yml) or [pre-commit hook](https://github.com/maplibre/maplibre-tile-spec/blob/main/.pre-commit-config.yaml) so that this is done automatically?

This way, we fix it instead of requiring "not quite productive" work from an contributor ^^

CommanderStorm (approved):
thanks

PR #1462, opened 2026-06-23
chore(deps): adapt to typescript 6 and strict mode

CommanderStorm:
I asked Claude to update us to ts6, which also apparently means strict mode.

A few of the changes that were nessary seem not quite right, but I need to investigate further.

PR mostly to give notice that I am working on this


ts/src/decoding/stringDecoder.ts
```diff
@@ -82,57 +82,68 @@ export function decodeString(
...
 function decodeFsstDictionaryVector(
     name: string,
-    symbolTableStream: Uint8Array | null,
-    offsetStream: Uint32Array | null,
-    dictionaryLengthStream: Uint32Array | null,
-    dictionaryStream: Uint8Array | null,
-    symbolLengthStream: Uint32Array | null,
-    nullabilityBuffer: BitVector | null,
-): Vector | null {
+    symbolTableStream: Uint8Array | undefined,
```
CommanderStorm, line 91: not sure if undefined vs null is better here.

I just know that this being uneven between calle and caller makes for lots of akward conversion casts..

HarelM:
There's too many "as something" added, which I find odd like you did I think... It would be better to avoid those if possible.

ts/src/encoding/fastPforEncoder.ts
```diff
@@ -215,7 +215,7 @@ function recordBlockExceptions(
         if (value >>> bitWidth !== 0) {
             byteContainerPos = writeByte(workspace, byteContainerPos, k);
             if (exceptionBitWidth !== 1) {
-                const exceptionValues = dataToBePacked[exceptionBitWidth];
+                const exceptionValues = dataToBePacked[exceptionBitWidth] as Uint32Array;
```
CommanderStorm: array indexing can result in undefined, we know it cannot happen -> save

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
```diff
@@ -48,17 +48,20 @@ export function decodeField(src: Uint8Array, offset: IntWrapper): Field {
...
+    const base = decodeColumnType(typeCode);
+    if (!base) {
+        throw new Error(`Unsupported field type code ${typeCode}. Supported: ${SUPPORTED_FIELD_TYPES}`);
     }
 
+    // Field type codes (10-30) always carry an explicit name.
+    const column: Column = { ...base, name: decodeString(src, offset) };
+
     if (columnTypeHasChildren(typeCode)) {
+        const complexCol = column.complexType as ComplexColumn;
```
CommanderStorm: invariant of `columnTypeHasChildren`

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
```diff
@@ -70,27 +73,28 @@ export function decodeField(src: Uint8Array, offset: IntWrapper): Field {
...
+        name = "geometry";
     }
 
+    const column: Column = { ...base, name };
+
     if (columnTypeHasChildren(typeCode)) {
         // Only STRUCT (typeCode 30) has children
         const childCount = decodeVarintInt32(src, offset, 1)[0] >>> 0;
-        const complexCol = column.complexType;
+        const complexCol = column.complexType as ComplexColumn;
```
CommanderStorm: invariant of `columnTypeHasChildren`

CommanderStorm:
> There's too many "as something" added, which I find odd like you did I think... It would be better to avoid those if possible.

a lot more change, but sure. I updated the code to abort with better error messages on invariant violation

CommanderStorm:
> There's too many "as something" added, which I find odd like you did I think... It would be better to avoid those if possible.

Removed them, but that forced a lot of code churn, added aborts and other things.

HarelM:
This PR is a lot to review, any chance to slice it to smaller PRs?

CommanderStorm:
needs breaking apart. a bit too non-atomic.

HarelM:
While I don't fully stand behind this statement, I've checked locally and you can set `strcit: false` to update to typescript 6.
There's also a way to add more granular stricktiness check and avoid the `strict: true` as I find it a bit incorrect in some places and adds code just to satisfy the strict check without meaningful value.
At least that's my experience with it.
So I think a smaller change can be made here, but still leverage some valuable improvements strict is showing us.

PR #1463, opened 2026-06-23
spec: add missing extent field to FeatureTableMetadata schema

pltrz:
The FeatureTable binary format encodes an `extent varint` between the `name` string and the `columnCount` varint
[(see tileset.cpp)](https://raw.githubusercontent.com/maplibre/maplibre-tile-spec/refs/heads/main/cpp/src/mlt/metadata/tileset.cpp).

This field was absent from the FeatureTableSchema diagram and had no accompanying description, making any spec-only implementation produce an incorrect parser.

Add extent to the schema and document it's semantics: tile coordinate space size, default 4096.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -194,6 +194,7 @@ direction TB
 
     class FeatureTable {
       +String name
+      +VarInt extent
       +VarInt columnCount
       +Column[] columns
     }
@@ -210,6 +211,10 @@ direction TB
     TileMetadata --> FeatureTable : featureTables
     FeatureTable --> Column : columns
 ```
+
+The `extent` field defines the coordinate space size for the tile's geometry.
+Geometry coordinates in MapLibre Tiles are encoded as integers in the range `[0, extent)`.
+The default value is `4096`, matching the Mapbox Vector Tile convention.
 
 Strings are encoded as UTF-8 sequences of characters with a length header:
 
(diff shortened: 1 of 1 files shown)


pltrz:
related: https://github.com/maplibre/mvt-cpp/pull/3

CommanderStorm (approved):
Yes, thanks

docs/specification.md
```diff
@@ -211,6 +212,10 @@ direction TB
     FeatureTable --> Column : columns
 ```
 
+The `extent` field defines the coordinate space size for the tile's geometry.
+Geometry coordinates in MapLibre Tiles are encoded as integers in the range `[0, extent)`.
+The default value is `4096`, matching the Mapbox Vector Tile convention.
```
CommanderStorm: Except that because it is an required VarInt, we require it.

```suggestion
The required `extent` field defines the coordinate space size for the tile's geometry.
Geometry coordinates in MapLibre Tiles are encoded as signed integers in vector-tile grid coordinates (typically near the `0..=extent` range, but not restricted to it).
Values MAY be negative or exceed extent for geometry that crosses tile boundaries.
Encoders MAY default to `4096` when a user does not specify the extent.
Decoders MUST require it to diferente `extent` and `columnCount`.
```

CommanderStorm: and that it is not `[0..extent)`, but `(0..extent)`

PR #1465, opened 2026-06-24
feat(rust): narrow 64-bit property columns to 32-bit when values fit

CommanderStorm:
Currently when converting MVT-> MLT we don't auto-narrow.
This is problematic because it is quite easy to add and because on the web 64-bit numbers are a problem.


CommanderStorm:
Notably, this is an important thing to do, because gl-js interprets u64 as an bigint, which then gets interpreted as an object, which then breaks most of the things in the expression system.. 😓

CommanderStorm:
+1 from @nyurik

PR #1466, opened 2026-06-24
fix(rust): Skip encoding competition for single-value int streams; pick smaller of VarInt/plain

CommanderStorm:
- non-optimal strat
- doing work is slower than skipping work
- not a "high value" issue though, given occurance

### DIFF
--- a/rust/mlt-core/src/encoder/stream/codecs.rs
+++ b/rust/mlt-core/src/encoder/stream/codecs.rs
@@ -140,27 +140,30 @@ impl Codecs {
             return encoder::write_stream_payload(enc.data_mut(), meta, false, vals2);
         }
 
-        // One value has no deltas and no runs, and FastPFOR framing always
-        // costs more here. Only the physical layout varies: VarInt is shorter
-        // until the value needs all its bytes (>= 2^28 for u32, >= 2^56 for u64,
-        // post-zigzag for signed), where plain avoids VarInt's continuation
-        // byte. Pick the smaller; no competition needed.
-        if values.len() == 1 {
-            let Self { logical, physical } = self;
-            let vals1 = logical.none(values);
-            let plain: &[u8] = cast_slice(vals1);
-            let varint = physical.varint(vals1);
-            let (pe, payload) = if varint.len() <= plain.len() {
-                (PE::VarInt, varint)
+        // A single value has no deltas or runs, so Delta would only zigzag the
+        // lone value into something equal or larger, and FastPFOR never pays off
+        // for one value. Only the physical layout matters, so skip the
+        // competition. Matching the encoded stream binds the lone value with no
+        // indexing bounds check. VarInt is shorter until the value needs all its
+        // bytes (>= 2^28 for u32, >= 2^56 for u64, post-zigzag for signed),
+        // where fixed-width plain avoids VarInt's continuation byte.
+        let Self { logical, physical } = self;
+        if values.len() == 1
+            && let [value] = logical.none(values)
+        {
+            let n: u64 = (*value).into();
+            let varint_len = (u64::BITS - n.leading_zeros()).max(1).div_ceil(7) as usize;
+            let encoded = std::slice::from_ref(value);
+            let (pe, payload) = if varint_len <= size_of_val(encoded) {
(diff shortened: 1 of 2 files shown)


rust/mlt-core/src/encoder/stream/codecs.rs
```diff
@@ -140,8 +140,30 @@ impl Codecs {
...
-        let allow_fastpfor = enc.config().allow_fastpfor();
+        // A single value has no deltas or runs, so Delta would only zigzag the
+        // lone value into something equal or larger, and FastPFOR never pays off
+        // for one value. Only the physical layout matters, so skip the
+        // competition. Matching the encoded stream binds the lone value with no
+        // indexing bounds check. VarInt is shorter until the value needs all its
+        // bytes (>= 2^28 for u32, >= 2^56 for u64, post-zigzag for signed),
+        // where fixed-width plain avoids VarInt's continuation byte.
         let Self { logical, physical } = self;
+        if values.len() == 1
```
CommanderStorm, line 151: len and the destructuring since destructuring can mean zigzagging, so we only want to do this if we know this is nessary 😉

CommanderStorm:
+1 from @nyurik

michaelkirk:
It seems this isn't supported by cpp. 😭

```
MLN[warning] [event]:ParseTile [code]:-1 [message]:MLT parse failed: Specified physical level technique not yet supported 0 (virtual bool mbgl::MLNCoreLoggingObserver::onRecord(EventSeverity, Event, int64_t, const std::string &):29)
```

CommanderStorm:
Anoying.
We SHOULD have a test against this, but apparently we don't.

This might be interesting to you
- https://github.com/maplibre/maplibre-native/pull/4428

michaelkirk:
It seems like this isn't supported by Java, TS, or CPP, so only rust supports this. 😞 

I understand the motivation for the change, and it's allowed in the spec, but if I'm understanding it correctly, it broke interop with every other client for very little gain. In hindsight I would say this change was not worthwhile. 

I'm not here just to criticize though! I'm hoping to help make these kind of breaking changes happen less often, or at least not to be done accidentally.

My current effort is https://github.com/michaelkirk/martin-e2e which has a www (maplibre-gl-js) and a maplibre-native (cpp) app consuming martin (rust) tiles. It's caught several things not yet covered by the synthetics (I'll start upstreaming them, but it's a slow process).

> https://github.com/maplibre/maplibre-native/pull/4428

This is indeed interesting!

nyurik:
if none of the decoders support this, we can use it to v2

michaelkirk:
> if none of the decoders support this, we can use it to v2

That's a good point. We **could** revert this encoder change for now, unless people feel like this ship has already sailed.

This one is interesting because I think it's allowed by the v1 spec - it just wasn't implemented by any of the decoders (probably because it's not a very valuable optimization). I know MLT is still young software, I don't think full https://lkml.org/lkml/2012/12/23/75 mode is warranted yet, but I think it would make sense to have a higher bar for small encoder optimizations which could break interop until v2 (assuming v2 is certainly a thing).

I don't mean to over index on this particular change, it's just one of several interop breakages I've encountered while trying to get ts, cpp, and rust MLT to talk to each other. 

An alternative to waiting for a major version bump, would be to push out decoder support to all client libs for a while before enabling the new behavior in the encoder. I also acknowledge that being diligent about tracking feature state like that would be additional work, perhaps more work than maintainers are willing to accommodate.

CommanderStorm:
For plain, I think all patches are already in, reverting is more work. 
We originally specified it to exist..

michaelkirk:
#1544 isn't merged yet - and even if it were, theres's a big difference between in `main` vs released and deployed. Please also consider us poor souls trying to use your code which explodes - every time this is encountered, it's "more work" for us to figure out what's going on and what needs to be updated.

> We originally specified it to exist..

This is exactly the mindset I'm arguing against — "it's technically legal, so we should do it without regard for people actually using the software" is not the way to think about a deployed system.

nyurik:
I agree with @michaelkirk here - if the new, not widely used data rust MLT encoder generates is not supported by the majority of MLT decoders, lets only enable it in v2

PR #1468, opened 2026-06-24
refactor(ts): construct Column/Field as discriminated unions

CommanderStorm:
Split from #1462:
build Column/Field as immutable discriminated-union literals (adds ColumnWithoutName) instead of `{} as Column` plus mutation.

### DIFF
--- a/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
+++ b/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
@@ -77,10 +77,11 @@ function decodeColumn(src: Uint8Array, offset: IntWrapper): Column {
     if (columnTypeHasName(typeCode)) {
         name = decodeString(src, offset);
     } else if (typeCode <= 3) {
-        // ID and GEOMETRY columns have implicit names
         name = "id";
-    } else {
+    } else if (typeCode === 4) {
         name = "geometry";
+    } else {
+        throw new Error(`Unsupported column type code ${typeCode}. Supported: ${SUPPORTED_COLUMN_TYPES}`);
     }
 
     const column: Column = { ...base, name };
--- a/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
+++ b/ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
@@ -29,10 +29,11 @@ function decodeString(src: Uint8Array, offset: IntWrapper): string {
  * Used when decoding Field metadata which has the same format as Column.
  */
 function columnToField(column: Column): Field {
-    const base = { name: column.name, nullable: column.nullable };
+    const name = column.name;
+    const nullable = column.nullable;
     return column.type === "scalarType"
-        ? { ...base, type: "scalarField", scalarField: column.scalarType }
-        : { ...base, type: "complexField", complexField: column.complexType };
+        ? { type: "scalarField", scalarField: column.scalarType, name, nullable }
+        : { type: "complexField", complexField: column.complexType, name, nullable };
 }
 
 /**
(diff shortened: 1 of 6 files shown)


ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
```diff
@@ -29,13 +29,10 @@ function decodeString(src: Uint8Array, offset: IntWrapper): string {
...
  */
 function columnToField(column: Column): Field {
-    return {
-        name: column.name,
-        nullable: column.nullable,
-        scalarField: column.scalarType,
-        complexField: column.complexType,
-        type: column.type === "scalarType" ? "scalarField" : "complexField",
-    };
+    const base = { name: column.name, nullable: column.nullable };
```
HarelM: Is this the same logic? Because the previos version is more readable...

CommanderStorm: `complexField` does only make sense if `type: "complexField"`.

that is the change we are doing here to not have to cast it, but have this act as a discriminant.

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
```diff
@@ -29,13 +29,10 @@ function decodeString(src: Uint8Array, offset: IntWrapper): string {
...
+    return column.type === "scalarType"
+        ? { ...base, type: "scalarField", scalarField: column.scalarType }
+        : { ...base, type: "complexField", complexField: column.complexType };
```
CommanderStorm:
```suggestion
    const name = column.name;
    const nullable = column.nullable;
    return column.type === "scalarType"
        ? { type: "scalarField", scalarField: column.scalarType, name, nullable }
        : { type: "complexField", complexField: column.complexType, name, nullable };
```

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
CommanderStorm: cannot both be more type-save and less type save...

ts/src/metadata/tileset/tilesetMetadata.ts
```diff
@@ -32,62 +32,65 @@ export const LogicalComplexType = {
...
 }
 
-export interface Field {
-    name?: string | null;
-    nullable?: boolean | null;
-    scalarField?: ScalarField | null;
-    complexField?: ComplexField | null;
-    type?: "scalarField" | "complexField";
-}
+export type Field = {
```
HarelM, line 77: This was changed to a `type` while the rest are still interfaces. I think it would be better to change the interfaces here to types as I don't think there's much value to "inheritance" with those.

ts/src/metadata/tileset/embeddedTilesetMetadataDecoder.ts
```diff
@@ -70,24 +67,25 @@ export function decodeField(src: Uint8Array, offset: IntWrapper): Field {
...
+    } else if (typeCode <= 3) {
         // ID and GEOMETRY columns have implicit names
-        if (typeCode >= 0 && typeCode <= 3) {
-            column.name = "id";
-        } else if (typeCode === 4) {
-            column.name = "geometry";
-        }
+        name = "id";
+    } else {
+        name = "geometry";
```
HarelM, line 82: Would it make sense to add back the `typeCode === 4` for geometry and throw for the other column types?
As far as I understand 0-3 is ID, 4 is geometry and has name is above 10, so that leaves 5-9 and "unhandled". The previous code behaved similarly for those, but it didn't assign geometry for column name...

HarelM:
I've added some comments, but this looks good overall, thanks!

HarelM (approved):

PR #1470, opened 2026-06-24
fix(ts): Add defensive guards on geometry decode paths

CommanderStorm:
Split out from #1462

Adds named-error guards on the geometry decode paths.
Missing vertex/triangle/offset buffers/ morton settings fail with a clear message instead of a later errors.

### DIFF
--- a/ts/src/decoding/geometryDecoder.ts
+++ b/ts/src/decoding/geometryDecoder.ts
@@ -96,7 +96,14 @@ export function decodeGeometryColumn(
             }
         }
 
+        if (vertexBuffer === undefined) {
+            throw new Error("Geometry column is missing its vertex buffer.");
+        }
+
         if (indexBuffer) {
+            if (triangleOffsets === undefined) {
+                throw new Error("Tessellated geometry is missing its triangle offsets.");
+            }
             if (geometryOffsets !== undefined || partOffsets !== undefined) {
                 /* Case when the indices of a Polygon outline are encoded in the tile */
                 const topologyVector = { geometryOffsets, partOffsets, ringOffsets };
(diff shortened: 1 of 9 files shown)


ts/src/vector/geometry/gpuVector.ts
```diff
@@ -139,7 +144,6 @@ export abstract class GpuVector implements Iterable<CoordinatesArray> {
             yield geometries[index++];
         }*/
 
-        //throw new Error("Iterator on a GpuVector is not implemented yet.");
-        return null;
+        throw new Error("Iterator on a GpuVector is not implemented yet.");
```
HarelM: Can you remove the commented out code above? Or remove this method completely? I don't think it's in use...

HarelM:
Are these possible code flows? I'm not very keen on defensive programing...

ts/src/decoding/geometryDecoder.ts
```diff
@@ -208,13 +215,19 @@ export function decodeGeometryColumn(
...
-    if (indexBuffer && !partOffsets) {
-        /* Case when the indices of a Polygon outline are not encoded in the data so no
-         *  topology data are present in the tile */
-        return createFlatGpuVector(geometryTypeVector, triangleOffsets, indexBuffer, vertexBuffer);
+    if (vertexBuffer === undefined) {
+        throw new Error("Geometry column is missing its vertex buffer.");
     }
 
     if (indexBuffer) {
+        if (triangleOffsets === undefined) {
```
HarelM, line 223: Wasn't this already tested above?
This method is huge, I suggest splitting it, otherwise understanding which part can happen in which case is a nightmare...

ts/src/vector/geometry/geometryVectorConverter.ts
```diff
@@ -15,9 +15,9 @@ export function convertGeometryVector(geometryVector: GeometryVector): Coordinat
 
     const mortonSettings = geometryVector.mortonSettings;
     const topologyVector = geometryVector.topologyVector;
-    const geometryOffsets = topologyVector.geometryOffsets;
-    const partOffsets = topologyVector.partOffsets;
-    const ringOffsets = topologyVector.ringOffsets;
+    // These topology arrays are present only for the geometry types that need them (e.g. multi-geometries
```
HarelM: I suggest to move this comment into the `geometryVector.topologyVector` property as a tsdoc comment instead of adding it here.

HarelM:
Looking at the errors that were added, I'm not sure they are a lot better than "x is not a property of undefined", as they don't really help understanding where the issue actually is and how to solve it.
IDK...

ts/src/decoding/geometryDecoder.ts
CommanderStorm: Yes, but TS cannot proof that `!!indexBuffer` implies `!!triangleOffsets`.
So we need to abort here again.

So I need to abort in this scope again.

ts/src/vector/geometry/gpuVector.ts
CommanderStorm: the cpde above is roughly what we will need once we use the gpu vector for better perf.

I suggest we keep it, but if you want to churn here it is your call.

CommanderStorm:
The issues should not be producible on valid input by definition.
So if we design our encoders well, they are impossible by construction.

Problem is that I still will have undefined in the type, and that is not something I think I can fix..

CommanderStorm:
If we want ts6, strict mode seems to be one of the requirements, so therefore we need to actually handle our undefined..

Issue #1137, opened 2026-03-12
Clarify support for plain integer streams


DoFabien:
PR #1138   fixes a real TypeScript bug around integer `PhysicalLevelTechnique.NONE`, but while working on it I found that the repository is not aligned on whether plain integer streams are actually part of the supported design.

Current state:

- `docs/specification.md` lists integer physical techniques as `SIMD-FastPFOR` and `Varint`
- `docs/encodings.md` still defines `Plain` integer and long payloads as little-endian byte streams
- Rust supports and emits plain integer streams (`PhysicalEncoder::None`)
- Java integer decoding only supports `FAST_PFOR` and `VARINT`
- the generic C++ integer decoder rejects `NONE`, although geometry code has a partial tolerance path

So the question is: what is the intended project-wide direction here?

1. Keep plain integer streams as supported
2. Remove them from the supported design
3. Other / mixed approach

HarelM:
Thanks for opening this issue! I'll post it in the slack channel to get more traction.
Very good and relevant questions.

mactrem:
There are pros and cons. I would recommend keeping plain integer streams, mostly as they enable zero-copy use cases on fast networks where transcoding speed can become the bottleneck rather than transfer size.
This is also consistent with established columnar storage formats that maintain plain encodings such as [Apache Parquet](https://parquet.apache.org/docs/file-format/data-pages/encodings/), [BtrBlocks](https://www.cs.cit.tum.de/fileadmin/w00cfj/dis/papers/btrblocks.pdf), [Nimble](https://github.com/facebookincubator/nimble/blob/2a4ddcb67f193698816792f648945dedefc4d1ef/dwio/nimble/docs/Encodings.md).

CommanderStorm:
now supported everywhere

PR #1471, opened 2026-06-24
test(rust): Add Rust <-> C++ differential decoder fuzzer

CommanderStorm:
a diff fuzzer for rust and cpp.

Currently not wired into CI, because it finds issues 🎉 

- https://github.com/maplibre/maplibre-tile-spec/issues/1137


CommanderStorm:
+1 by @nyurik

PR #1472, opened 2026-06-24
test(synth): Handle upper coordinate space IDs in `RLE`/`Delta`.. logcial space

### LINKED ISSUE #1075: OMT fixture ID mismatch on large feature IDs

CommanderStorm:
Adds Rust-only synthetic fixtures for u32::MAX and u64::MAX feature ids under the delta, rle, and delta_rle logical encodings; the u32 delta-encoded ones are decoded incorrectly by the C++ decoder today (sign-extended to u64), which these fixtures pin down.

Both TS and CPP fail this.
TS does confuse `u32::MAX` being `-1`, so a signed/unsigned bug
CPP does assume that  `u32::MAX` and  `u64::MAX` fit the same space and that `-1` is the same for both. It is not, so incorrect.

Maybe resolves https://github.com/maplibre/maplibre-tile-spec/issues/1075

Hard to tell..


HarelM (approved):
TS part approved.

PR #1486, opened 2026-07-02
fix: FSST shared dictionary encoded as DictionaryType::Single

### LINKED ISSUE #1085: FSST shared dictionary encoding results in `DictionaryType::Single`
### Bug Description
Currently, when a FSST shared dictionary is encoded, this results in `DictionaryType::Single`. This is very likely a bug, since that does not make a lot of sense.

### Expected Behavior
Encoding a FSST shared dictionary should not produce `DictionaryType::Single`. It should be reviewed to ensure the correct dictionary type is used.

### Steps to Reproduce
- Encode a FSST shared dictionary with the current implementation.
- Observe that the result is `DictionaryType::Single`.

### Additional Context
This behavior appears inconsistent with the intended usage of shared dictionaries.

mvanhorn:
## Summary

Tags the FSST shared-dictionary corpus stream with `DictionaryType::Shared` instead of `DictionaryType::Single` when encoding string properties. The stream metadata now matches how the data is actually laid out, so a shared FSST dictionary round-trips correctly.

## Why this matters

Issue #1085 reports that FSST shared dictionaries are encoded with the wrong `DictionaryType`. The encoder emitted `Single` for the corpus stream even though the values reference a shared dictionary, so the stream header disagreed with the payload. A decoder trusting the header would mis-read a shared FSST dictionary.

The fix sets the corpus stream's dictionary type to `Shared` in the FSST shared-dictionary path. It is a one-line correction to the stream metadata; the encoded bytes for the dictionary itself are unchanged.

## Testing

`cargo test -p mlt-core --features __private --test mvt_round_trip` — 135 passed. Added a focused round-trip regression that encodes a shared FSST dictionary and asserts the corpus stream is tagged `Shared` and decodes back to the original strings.

Fixes #1085

### DIFF
--- a/rust/mlt-core/src/encoder/property/shared_dict.rs
+++ b/rust/mlt-core/src/encoder/property/shared_dict.rs
@@ -362,7 +362,7 @@ impl Codecs {
         // Write stream data: total count, corpus streams, then per-child streams.
         enc.write_varint(stream_len)?;
         if let Some(ref raw) = fsst_raw {
-            write_fsst_data(raw, DictionaryType::Single, &shared_dict.prefix, enc, self)?;
+            write_fsst_data(raw, DictionaryType::Shared, &shared_dict.prefix, enc, self)?;
         } else {
             let lengths = strings_to_lengths(&dict)?;
             let typ = StreamType::Length(LengthType::Dictionary);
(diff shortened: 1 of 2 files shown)


CommanderStorm:
We will fix this bug in v2, but for v1 we are stuck with this being `Single` I think.
I am sure we will revisit this once the code for this has been merged.

Sorry

PR #1492, opened 2026-07-07
chore(cpp): Replace fastpfor dependency with snapshot that doesn't require SIMD/SSE to build

TimSylvester:
The FastPFOR dependency removed SIMDE, which was the only way SIMD support would build on ARMv7.

We don't currently use the SIMD encoder (because the output is not compatible with the non-SIMD decoder), so we don't need it.

This moves the FastPFOR submodule to local files in vendor/fastpfor and removes the SIMD codecs.

As a result, FastPFOR is no longer a build option.


cpp/CMakeLists.txt
```diff
@@ -101,9 +99,15 @@ foreach(_target mlt-cpp mlt-cpp-encoder)
     target_include_directories(${_target}
         PUBLIC ${PROJECT_SOURCE_DIR}/include
         PRIVATE ${PROJECT_SOURCE_DIR}/src
+        PRIVATE ${PROJECT_SOURCE_DIR}/vendor/fastpfor
```
louwers: I think you can achieve the same with `target_link_libraries`.

cpp/CMakeLists.txt
```diff
@@ -142,6 +146,7 @@ target_sources(mlt-cpp PRIVATE
     ${PROJECT_SOURCE_DIR}/src/mlt/util/space_filling_curve.hpp
     ${PROJECT_SOURCE_DIR}/src/mlt/util/vectorized.hpp
     ${PROJECT_SOURCE_DIR}/src/mlt/util/zigzag.hpp
+    ${FASTPFOR_SOURCES}
```
louwers: Again, I think `target_link_libraries` is the way to go here.

cpp/BUILD.bazel
```diff
@@ -3,18 +3,35 @@ load("@rules_cc//cc:cc_library.bzl", "cc_library")
...
+    visibility = ["//visibility:private"],
+)
+
 cc_library(
     name = "mlt_cpp",
     srcs = glob(
-        ["src/mlt/**/*.cpp"],
+        [
+            "src/mlt/**/*.cpp",
+            "vendor/fastpfor/fastpfor/*.cpp",
```
louwers: `fastpfor` should be a separate `cc_library`.

louwers (approved):

PR #1497, opened 2026-07-10
fix(cpp): Minor fixes for MLN build issues on diverse platforms

TimSylvester:
MLN builds on more platforms and configurations than the MLT CI covers.


CommanderStorm (approved):
No clue why this code is nessary, but sure. You likely have a reason

TimSylvester:
These fix MLN build errors on MSVC and Android that we didn't catch within the MLT project because we're not building those.

PR #1498, opened 2026-07-12
feat(rust): Add PMTiles compression support to the Rust MVT-to-MLT converter

candux:
I'm looking into switching my project [wogibt.es](https://wogibt.es/) to MLT.

When I first ran `mlt convert`, I was surprised that the output PMTiles archive was much larger than the input. The MVT tiles were gzip-compressed, while the generated MLT tiles were not. This PR adds compression support and preserves compatible input compression by default.

## Changes

- Add `--tile-compression auto|gzip|none`.
- Use `auto` by default.
- Preserve uncompressed, gzip, Brotli, and Zstandard input compression in auto mode.
- Require an explicit `gzip` or `none` override for zlib input because PMTiles v3 cannot represent zlib compression.
- Write the selected compression to the PMTiles header and metadata.
- Reject explicit compression options for unsupported input/output combinations.
- Report raw payload and complete archive sizes separately.
- Add documentation and tests for compression handling and round trips.

Brotli and Zstandard are intentionally only preserved through auto-detection and are not exposed as explicit CLI options. Common web PMTiles clients require custom decompression support for these codecs, so exposing them directly
could make it easy to produce archives that clients cannot read.

If exposing every PMTiles-supported compression mode is preferred, I can add explicit `brotli` and `zstd` options as well.

Example output:

```text
input.mvt.pmtiles -> output.mlt.pmtiles (pmtiles):
converted 5 tiles (5 unique encoded, 0 cache hits) in 740.9ms
size raw/archive: MVT(gzip) 813.7kB/459.8kB -> MLT(gzip) 460.3kB/357.0kB
```

## AI Assistance

This pull request was developed with assistance from OpenAI Codex. The changes were reviewed and tested by me.


rust/mlt/src/convert/mod.rs
```diff
@@ -120,6 +121,60 @@ enum SortMode {
...
+    /// Auto may retain Brotli or Zstandard when the source already uses it,
+    /// but they are intentionally not explicit options because common web
+    /// `PMTiles` clients require custom decompression support for those codecs.
+    #[default]
+    Auto,
+    /// Store MLT tile payloads without outer compression
+    None,
+    /// Gzip-compress each MLT tile payload
+    Gzip,
+}
```
CommanderStorm: This right here is a tradeoff between speed and size.
Yes, you can get 10-20% from GZ, but at what cost..

```suggestion
pub(super) enum PmtilesTileCompression {
    /// Store MLT tile payloads without outer compression
    #[default]
    None,
    /// Gzip-compress each MLT tile payload for size, sacrificing decoding speed
    Gzip,
}
```

CommanderStorm (commented):
Here are my benchmarks (what our v1 rust encoder was build against).
In v2, we plan to remove the metadata part of small tiles, which will reduce the total tile size metrics.

 If your benchmarks are significantly diffferent (you said significant 🤔), could you maybe share your data with me? We possibly have a bug where one of our heuristics might be misfiring or similar.
=> could you share your data if this is possible, so I can look into where we are going wrong?

Here are my benchmarking metrics (they are sort of expeced results until we can start towards v2. unfortunately, small tiles are a bit larger for us currently):

Image

Image

Will look over the PR later in the train

candux:
> If your benchmarks are significantly diffferent (you said significant 🤔), could you maybe share your data with me? We possibly have a bug where one of our heuristics might be misfiring or similar. => could you share your data if this is possible, so I can look into where we are going wrong?

Sure, I can share the data:

Input MVT: https://wogibt.es/osm/old/osm-wogibtes.pmtiles
Output MLT: https://wogibt.es/osm/old/osm-wogibtes.mlt.gzip.pmtiles

Feel free to use them however you like. They contain only OSM data.

For 397,834 tiles:

| Format | Raw | Archive |
|---|---:|---:|
| MVT gzip | 3.1 GB | 1.1 GB |
| MLT none | 1.4 GB | 1.4 GB |
| MLT gzip | 1.4 GB | 870.3 MB |

The dataset is a bit unusual because it is dominated by an OSM hiking-relations layer, resulting in many features per tile. The tiles contain numerous route relations, repeated strings, and list-like relation properties. One example z4 tile contains 86,615 features.

My impression is that my tiles are far from optimal. This has also exposed a few performance problems in the MLT implementation here and in MapLibre GL JS. I have a few more PRs addressing those—or perhaps I should just optimize my tiles. :)

rust/mlt/src/convert/mod.rs
candux: I would have expected gzip decompression to take somewhere from sub-millisecond to a few milliseconds per tile, but I haven’t benchmarked it.

For my dataset, gzip reduced the archive size by 38%, so I assumed the size savings would be worth the decompression cost.

I can remove `auto` entirely and make `none` the default if that is your preference.

CommanderStorm: Yes, MLT uses lightweight encodings, so while hevyweight encodings CAN decrease the size further a bit, it is usually not desirable from an speed perspective..

CommanderStorm:
> Input MVT: https://wogibt.es/osm/old/osm-wogibtes.pmtiles

Interestingly, the reason why we are larger is that the following heristic is misfiring and adding shared dicts when it should not.
We don't want to remove it entirely, since it is -7% on OMT (mostly because cyrilic languages), but I think I am thinking about smarter alternatives

- #1283

rust/mlt/src/convert/common.rs
```diff
@@ -46,16 +46,18 @@ pub fn make_encode_cache() -> EncodeCache {
 /// Only small tiles (ocean, empty land, ...) actually repeat across a tileset;
 /// big city tiles are essentially unique, so tiles over
 /// [`MAX_TILE_CACHE_TRACK_SIZE_BYTES`] skip the cache (any rare repeat still
-/// dedups when the container is written). Returns the encoded bytes and whether
-/// they came from the cache. Mirrors `files.rs` and the mbtiles transcoder.
+/// dedups when the container is written). Returns the encoded bytes, the raw
+/// MVT size, and whether the result came from the cache. Mirrors `files.rs` and
+/// the mbtiles transcoder.
```
CommanderStorm:
```suggestion
/// MVT size, and whether the result came from the cache.
```

rust/mlt/src/convert/mod.rs
```diff
@@ -120,6 +121,54 @@ enum SortMode {
...
+                Encoding::Gzip => Ok(Compression::Gzip),
+                Encoding::Brotli => Ok(Compression::Brotli),
+                Encoding::Zstd => Ok(Compression::Zstd),
+                Encoding::Zlib => bail!(
+                    "PMTiles does not support zlib tile compression; use --tile-compression gzip or none"
+                ),
+            },
+        }
+    }
+}
```
CommanderStorm:
```suggestion
impl From<PmtilesTileCompression> for Compression {
    fn from(comp: PmtilesTileCompression) -> Compression {
        match comp {
            PmtilesTileCompression::None => Ok(Compression::None),
            PmtilesTileCompression::Gzip => Ok(Compression::Gzip),
        }
    }
}
```

rust/mlt/README.md
```diff
@@ -6,8 +6,33 @@ The `mlt` binary provides several commands for working with MLT files:
...
+cargo run --release -p mlt -- \
+  convert input.mvt.pmtiles output.mlt.pmtiles \
+  --tile-compression gzip
+```
+
+The default is `--tile-compression auto`, which preserves gzip, Brotli,
+Zstandard, and uncompressed input. PMTiles does not support zlib compression,
+so zlib input requires an explicit `gzip` or `none` override. Non-auto values
+are only accepted when the output path has a `.pmtiles` extension.
+
```
CommanderStorm:
```suggestion
```

CommanderStorm (approved):
shaved off ~150MB today in the other PR, but this is still more than the 870MB that this PR produces.
This is a good feature to have, lets merge

PR #1499, opened 2026-07-12
perf(ts): optimize full FSST dictionary decoding

candux:
## Summary

Optimizes full FSST dictionary decoding without changing decoder behavior.

The previous implementation accumulated decoded bytes in a JavaScript number array and then copied them into a `Uint8Array`. The new implementation:

1. Scans the compressed data to calculate the decoded size.
2. Allocates an exact-size `Uint8Array`.
3. Decodes directly into that array.

FSST decoding has a significant absolute cost only for unusually large dictionaries. For typical dictionaries the cost is already very small, but this change improves performance across every tested dictionary size.

This PR intentionally covers only full dictionary decoding, which benefits the current MapLibre GL JS access pattern.

## Benchmarks

The benchmark uses three complete MLT fixtures with typical, large, and xlarge FSST dictionaries. Results were measured with Node 24.11.1.

| Dictionary | Compressed | Decoded | Before | After | Speedup |
| --- | ---: | ---: | ---: | ---: | ---: |
| Typical | 4,670 B | 8,712 B | 0.0543 ms | 0.0273 ms | 1.99× |
| Large | 46,300 B | 78,929 B | 1.4007 ms | 0.4116 ms | 3.40× |
| xlarge | 4,545,692 B | 8,360,752 B | 144.85 ms | 32.42 ms | 4.47× |

Run the benchmark with:

```sh
cd ts
npm ci
npm run bench:fsst
```

The xlarge fixture is an OpenStreetMap-derived tile. Its provenance and license are documented alongside the fixture.

## AI assistance
This PR was developed with assistance from OpenAI Codex. The implementation was reviewed, benchmarked, and tested by me.

### DIFF
--- a/ts/package.json
+++ b/ts/package.json
@@ -15,6 +15,7 @@
   ],
   "scripts": {
     "build": "tsc",
+    "bench": "vitest bench --run src",
     "test": "vitest run --coverage --coverage.reportOnFailure",
     "lint": "eslint",
     "format": "biome format --write ./src",
--- a/ts/src/decoding/fsstDecoder.bench.ts
+++ b/ts/src/decoding/fsstDecoder.bench.ts
@@ -1,0 +1,74 @@
+import { readFileSync } from "node:fs";
+import { bench, describe } from "vitest";
+import decodeTile from "../mltDecoder";
+import { StringFsstDictionaryVector } from "../vector/fsst-dictionary/stringFsstDictionaryVector";
+import { decodeFsst } from "./fsstDecoder";
+
+type FsstDictionary = {
+    symbols: Uint8Array;
+    symbolLengths: Uint32Array;
+    compressed: Uint8Array;
+    decodedLength: number;
+};
+
+let checksum = 0;
+
+function loadDictionary(tilePath: string, tableName: string, columnName: string): FsstDictionary {
+    const tileData = readFileSync(new URL(tilePath, import.meta.url));
+    const featureTable = decodeTile(new Uint8Array(tileData.buffer, tileData.byteOffset, tileData.byteLength)).find(
+        (table) => table.name === tableName,
+    );
+    const dictionaryVector = featureTable?.propertyVectors.find((vector) => vector?.name === columnName);
+    if (!(dictionaryVector instanceof StringFsstDictionaryVector)) {
+        throw new Error(`FSST benchmark dictionary ${tableName}.${columnName} not found`);
+    }
+    // Access internal buffers so this benchmark isolates decodeFsst instead of measuring cached vector access.
+    const {
+        dataBuffer: compressed,
+        symbolOffsetBuffer: symbolOffsets,
+        symbolTableBuffer: symbols,
+        offsetBuffer,
+    } = dictionaryVector as unknown as {
+        dataBuffer: Uint8Array;
+        symbolOffsetBuffer: Uint32Array;
+        symbolTableBuffer: Uint8Array;
+        offsetBuffer: Uint32Array;
+    };
+    const symbolLengths = new Uint32Array(symbolOffsets.length - 1);
+    for (let i = 0; i < symbolLengths.length; i++) {
+        symbolLengths[i] = symbolOffsets[i + 1] - symbolOffsets[i];
+    }
+    return {
+        symbols,
+        symbolLengths,
+        compressed,
+        decodedLength: offsetBuffer[offsetBuffer.length - 1],
+    };
+}
+
+function consume(decoded: Uint8Array): void {
+    checksum = (checksum + decoded[0] + decoded[decoded.length >> 1] + decoded[decoded.length - 1]) | 0;
+}
+
+describe("decode whole FSST dictionary", () => {
+    const typical = loadDictionary("../../../test/expected/tag0x01/amazon_here/5_16_10.mlt", "pois", "name");
+    bench(
+        `typical: ${typical.compressed.length} compressed bytes to ${typical.decodedLength} decoded bytes`,
+        () => consume(decodeFsst(typical.symbols, typical.symbolLengths, typical.compressed)),
+        { warmupTime: 500, time: 5_000 },
+    );
+
+    const large = loadDictionary("../../../test/expected/tag0x01/omt/14_8299_10748.mlt", "poi", "name");
+    bench(
+        `large: ${large.compressed.length} compressed bytes to ${large.decodedLength} decoded bytes`,
+        () => consume(decodeFsst(large.symbols, large.symbolLengths, large.compressed)),
+        { warmupTime: 500, time: 5_000 },
+    );
+
+    const xlarge = loadDictionary("../../../test/fixtures/osm/xlarge.mlt", "hiking", "name");
+    bench(
+        `xlarge: ${xlarge.compressed.length} compressed bytes to ${xlarge.decodedLength} decoded bytes`,
+        () => consume(decodeFsst(xlarge.symbols, xlarge.symbolLengths, xlarge.compressed)),
+        { warmupTime: 500, time: 5_000 },
+    );
+});
--- a/ts/src/decoding/fsstDecoder.ts
+++ b/ts/src/decoding/fsstDecoder.ts
@@ -1,4 +1,26 @@
 /**
+ * Calculates the exact output size before decoding. This allows one final
+ * `Uint8Array` allocation and avoids growing a JavaScript number array and
+ * copying it into a typed array afterward. Traversing the compressed data
+ * twice is always worthwhile here because it avoids those larger temporary
+ * allocations.
+ */
+function getDecodedLength(symbolLengths: Uint32Array, compressedData: Uint8Array): number {
+    let decodedLength = 0;
+    for (let i = 0; i < compressedData.length; i++) {
+        const symbolIndex = compressedData[i];
+        if (symbolIndex === 255) {
+            decodedLength++;
+            i++; // Skip the literal byte following the escape marker.
+        } else {
+            decodedLength += symbolLengths[symbolIndex];
+        }
+    }
+
+    return decodedLength;
+}
+
+/**
  * Decode FSST compressed data
  *
  * @param symbols           Array of symbols, where each symbol can be between 1 and 8 bytes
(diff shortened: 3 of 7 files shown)


ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
```diff
@@ -21,19 +27,64 @@ export class StringFsstDictionaryVector extends VariableSizeVector<Uint8Array, s
...
+                    this.dataBuffer,
+                );
+                const symbolDecoders = decodersForSymbols ?? new WeakMap<Uint32Array, FsstDecoder>();
+                symbolDecoders.set(this.symbolOffsetBuffer, this.dictionaryDecoder);
+                if (!decodersForSymbols) {
+                    const dataDecoders =
+                        decodersForData ?? new WeakMap<Uint8Array, WeakMap<Uint32Array, FsstDecoder>>();
+                    dataDecoders.set(this.symbolTableBuffer, symbolDecoders);
+                    if (!decodersForData) dictionaryDecoders.set(this.dataBuffer, dataDecoders);
+                }
```
CommanderStorm: to have communicated this:
Why `WeakMap`?

ts/src/decoding/fsstDecoder.ts
```diff
@@ -6,26 +6,144 @@
...
+                decodedData[decodedOffset++] = symbols[symbolOffset++];
             }
         }
     }
-    return new Uint8Array(decodedData);
+
+    return decodedData;
+}
+
+const CHECKPOINT_INTERVAL = 1 << 14;
```
CommanderStorm: I am curious.

So why are we check pointing in the first place?
Is check pointing important? Does this improve perf?

CommanderStorm (commented):
The PR seems a bit weird because it puts together 4-5 different performance optimisations into the same PR.

Would you be open to pulling them into different PRs? This way I can gain the experience that you already have which of them is how effective and why.

So one PR for the bench+testing, …

candux:
> The PR seems a bit weird because it puts together 4-5 different performance optimisations into the same PR.
> 
> Would you be open to pulling them into different PRs? This way I can gain the experience that you already have which of them is how effective and why.
> 
> So one PR for the bench+testing, …

Sorry, I should have explained the complete picture earlier. Some optimizations only make sense together with the corresponding MapLibre GL JS changes.

This started when I tested MLT in MapLibre GL JS using my initial, very suboptimal tiles. Tile loading was slow, and queryRenderedFeatures() sometimes took more than a second. I have since optimized the tiles, but the investigation revealed several opportunities for a more efficient architecture:

1. Lazy GL JS materialization: GL JS currently materializes all MLT features, properties, and geometries eagerly. My GL JS branch creates feature wrappers lazily and decodes only the properties and geometries required by styles, filters, or queries.

2. On-demand FSST decoding: Lazy property access helps less if reading one string expands the entire FSST dictionary. The range decoder expands only the requested dictionary value and falls back to full decoding when access becomes dense.

3. FSST checkpoints: Without an index, every range lookup must scan from the beginning of the compressed stream. Checkpoints help with repeated out-of-order reads. Their usefulness depends on the actual access pattern.

4. Shared decoder and value caches: Columns using the same shared dictionary reuse one decoder, while frequently accessed dictionary strings are decoded only once. WeakMap prevents the module-level cache from retaining unloaded tiles.

5. Per-feature geometry: Lazy geometry in GL JS requires getGeometry(index) in the MLT decoder. Sequential access uses a cursor, sparse or random access uses topology checkpoints, and dense consumers retain the existing whole-vector path. This is #1501.

6. Faster full decoding: Dense access remains common, so the full FSST decoder writes directly into a pre-sized typed array instead of first building a JavaScript number array.

Thinking about it, I agree that these should be smaller, independently measurable changes. I’ll submit each optimization here alongside the corresponding MapLibre GL JS change that introduces its access pattern, and split the work accordingly.


Does this all sound good to you?

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
candux: Because the cache is module-scoped: a regular Map would retain every tile’s buffers and decoder indefinitely. WeakMap allows them to be garbage-collected when the tile is unloaded. The nested maps form an identity key from the compressed dictionary, symbol table, and symbol offsets, all of which must match to safely share a decoder.

ts/src/decoding/fsstDecoder.ts
candux: FSST has variable-length symbols, so locating a decoded byte range otherwise requires scanning from the beginning every time. Checkpoints let later out-of-order reads start near the requested range.

They help repeated sparse/random access, but not the first lookup or dense sequential access. Their benefit depends on the consumer’s access pattern, so I’ll separate them and only include them with benchmarks showing an improvement for the corresponding MapLibre GL JS change.

CommanderStorm:
> Does this all sound good to you?

Yes, any help is always appreciated!
Split LGTM.

candux:
@CommanderStorm I’ve narrowed this PR to the FSST benchmark and a generic full-decoding optimization that improves performance across all dictionary sizes.

ts/src/decoding/fsstDecoder.bench.ts
```diff
@@ -0,0 +1,73 @@
...
+type FsstDictionary = {
+    symbols: Uint8Array;
+    symbolLengths: Uint32Array;
+    compressed: Uint8Array;
+    decodedLength: number;
+};
+
+const typicalDictionary = loadDictionary("../../../test/expected/tag0x01/amazon_here/5_16_10.mlt", "pois", "name");
+const largeDictionary = loadDictionary("../../../test/expected/tag0x01/omt/14_8299_10748.mlt", "poi", "name");
+const xlargeDictionary = loadDictionary("../../../test/fixtures/osm/xlarge.mlt", "hiking", "name");
```
CommanderStorm: please inlne for readability below

ts/src/decoding/fsstDecoder.bench.ts
```diff
@@ -0,0 +1,73 @@
...
+        compressed,
+        decodedLength: offsetBuffer[offsetBuffer.length - 1],
+    };
+}
+
+function consume(decoded: Uint8Array): void {
+    checksum = (checksum + decoded[0] + decoded[decoded.length >> 1] + decoded[decoded.length - 1]) | 0;
+}
+
+const options = { warmupTime: 500, time: 5_000 } as const;
```
CommanderStorm: please inline

ts/src/decoding/fsstDecoder.bench.ts
```diff
@@ -0,0 +1,73 @@
...
+                consume(decodeFsst(symbols, symbolLengths, compressed));
+            },
+            options,
+        );
+    });
+}
+
+benchmarkDictionary("typical dictionary", typicalDictionary);
+benchmarkDictionary("large dictionary", largeDictionary);
+benchmarkDictionary("xlarge dictionary", xlargeDictionary);
```
CommanderStorm: instead of this, could you simplify it down to one describe with the unwrapping of `loadDictionary` and benchmarks?
Currently a bit hard to parse what is happening and "kind of a weird way to do things"

ts/package.json
```diff
@@ -15,6 +15,7 @@
   ],
   "scripts": {
     "build": "tsc",
+    "bench:fsst": "vitest bench --run src/decoding/fsstDecoder.bench.ts",
```
CommanderStorm: can we not hardcode this via something like this

```suggestion
    "bench:fsst": "vitest bench --run src/**/*.bench.ts",
```

or maybe even this?


```suggestion
    "bench:fsst": "vitest bench",
```

I assume we will need more benchs over time, so this would be a bit of an tedious pattern

ts/src/decoding/fsstDecoder.ts
```diff
@@ -6,26 +6,38 @@
...
     }
 
+    let decodedLength = 0;
     for (let i = 0; i < compressedData.length; i++) {
-        if (compressedData[i] === 255) {
-            decodedData.push(compressedData[++i]);
+        const symbolIndex = compressedData[i];
+        if (symbolIndex === 255) {
+            decodedLength++;
+            i++;
```
CommanderStorm: i++? Can you explain this? It does not look like before we did skip a byte here before..

ts/src/decoding/fsstDecoder.ts
```diff
@@ -6,26 +6,38 @@
...
+            decodedLength += symbolLengths[symbolIndex];
+        }
+    }
+
+    const decodedData = new Uint8Array(decodedLength);
+    let decodedOffset = 0;
+    for (let i = 0; i < compressedData.length; i++) {
+        const symbolIndex = compressedData[i];
+        if (symbolIndex === 255) {
+            decodedData[decodedOffset++] = compressedData[++i];
```
CommanderStorm: Same goes below.

Can we not do mixing of ++ and other operations?
It is very hard to know what is the eval order -> making this code unnessarily hard to read for future devs

CommanderStorm (commented):
code mostly LGTM, but I would appreciate a quick round of cleanup for the benchmark and I think I might have found an logic bug.

See code review below ^^

ts/src/decoding/fsstDecoder.ts
candux: The old code skipped it with compressedData[++i]. Byte 255 is an escape marker, so the following byte is a literal and must be skipped as a symbol index.

ts/package.json
candux: Changed to 
`bench": "vitest bench --run src"
`
since
`vitest bench --run` also discovered stale dist/**/*.bench.js files.

ts/src/decoding/fsstDecoder.bench.ts
candux: done

ts/src/decoding/fsstDecoder.bench.ts
candux: done

ts/src/decoding/fsstDecoder.bench.ts
candux: done

ts/src/decoding/fsstDecoder.ts
candux: Agreed. I changed it to:

```
 i++;
 decodedData[decodedOffset++] = compressedData[i];
```

CommanderStorm (approved):
LGTM from my side

ts/src/decoding/fsstDecoder.ts
```diff
@@ -6,26 +6,39 @@
...
-    //TODO: use typed array directly
-    const decodedData: number[] = [];
-    const symbolOffsets: number[] = new Array(symbolLengths.length).fill(0);
+    const symbolOffsets = new Uint32Array(symbolLengths.length);
 
     for (let i = 1; i < symbolLengths.length; i++) {
         symbolOffsets[i] = symbolOffsets[i - 1] + symbolLengths[i - 1];
     }
 
+    let decodedLength = 0;
```
HarelM: [Optional] consider extracting this part to a method - `getDecodeLength` or `allocateDecodedDataArray` and explain there using tsdocs why this is faster and better.

HarelM (approved):
Looks good, thanks! I've left an optional comment to consider before merging.

ts/src/decoding/fsstDecoder.ts
candux: Good point, especially since I have some more optimizations lined up. I introduced getDecodedLength with a tsdocs explanation

PR #1503, opened 2026-07-13
fix(rust): preserve PMTiles geographic header

candux:
## What does this change?

Preserve geographic PMTiles header fields when converting MVT tiles to MLT.

For PMTiles inputs, the converter now copies:

- Minimum and maximum zoom
- Bounds
- Center coordinates and zoom

For MBTiles inputs, the converter populates those fields from the corresponding TileJSON metadata when present. Missing MBTiles metadata continues to use the PMTiles writer defaults.

Output-specific and content-derived fields—including tile type, compression, clustering, and tile counts—remain controlled by the output writer. MLT PMTiles output therefore remains uncompressed.

## Why?

The converter previously initialized the output with default PMTiles header values. This could produce an incorrect zoom range, bounds, and center even when the source contained accurate values.

For example, an input with zoom range `0..12` was written with the default output range `0..22`.

### main
```
$ pmtiles show source.pmtiles | grep -e zoom -e center
min zoom: 0
max zoom: 12
center: (long: 2.329102, lat: 48.835789)
center zoom: 12
$ convert source.pmtiles target.pmtiles

$ pmtiles show target.pmtiles | grep -e zoom -e center
min zoom: 0
max zoom: 22
center: (long: 0.000000, lat: 0.000000)
center zoom: 0
```

### with this PR:
```
$ pmtiles show source.pmtiles | grep -e zoom -e center
min zoom: 0
max zoom: 12
center: (long: 2.329102, lat: 48.835789)
center zoom: 12
$ convert source.pmtiles target.pmtiles

$ pmtiles show target.pmtiles | grep -e zoom -e center
min zoom: 0
max zoom: 12
center: (long: 2.329102, lat: 48.835789)
center zoom: 12
```

## AI notice

This change was developed with assistance from OpenAI Codex. The implementation and generated tests were reviewed and validated me

### DIFF
--- a/rust/mlt/src/convert/from_mbtiles.rs
+++ b/rust/mlt/src/convert/from_mbtiles.rs
@@ -14,10 +14,20 @@ use size_format::SizeFormatterSI;
 use usize_cast::FromUsize as _;
 
 use super::common::{
-    ENCODE_CACHE_BYTES, EncodedTile, MAX_TILE_CACHE_TRACK_SIZE_BYTES, TileStats, encode_tile,
-    make_encode_cache, make_progress_bar,
+    ENCODE_CACHE_BYTES, EncodedTile, MAX_TILE_CACHE_TRACK_SIZE_BYTES, PmTilesGeography, TileStats,
+    encode_tile, make_encode_cache, make_progress_bar,
 };
 use super::{ContainerFormat, MbtFormat, encode_one};
+
+fn geography_from_metadata(metadata: &Metadata) -> PmTilesGeography {
+    let tilejson = &metadata.tilejson;
+    PmTilesGeography {
+        min_zoom: tilejson.minzoom,
+        max_zoom: tilejson.maxzoom,
+        bounds: tilejson.bounds.map(|v| (v.left, v.bottom, v.right, v.top)),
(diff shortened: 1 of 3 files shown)


CommanderStorm (approved):
LGTM, but CI needs to be green. Curently, it is very red ^^

rust/mlt/src/convert/from_mbtiles.rs
```diff
@@ -14,11 +14,21 @@ use size_format::SizeFormatterSI;
...
+    encode_tile, make_encode_cache, make_progress_bar,
 };
 use super::{ContainerFormat, MbtFormat, encode_one};
 
+fn geography_from_metadata(metadata: &Metadata) -> PmTilesGeography {
+    let tilejson = &metadata.tilejson;
+    PmTilesGeography {
+        min_zoom: tilejson.minzoom,
+        max_zoom: tilejson.maxzoom,
+        bounds: tilejson.bounds.map(|v| (v.left, v.bottom, v.right, v.top)),
```
CommanderStorm, line 27: Ideally, I would prefer we get them from our actual encoding to prevent bad metadata being passed along.
Does not have to be this PR though.

PR #1504, opened 2026-07-13
fix(rust): use Option::filter instead of manual and_then

CommanderStorm:
## Summary
- Replace manual `and_then(|v| if v > max_bytes { None } else { Some(v) })` with idiomatic `filter(|&v| v <= max_bytes)` to fix `clippy::manual_filter` lint

## Test plan
- [x] `cargo clippy -p mlt-core -- -D warnings` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CommanderStorm:
fixes new clippy lint, force merging

PR #1507, opened 2026-07-15
feat(rust): [PoC] Support 3D geometry (`GeometryZ`)

yutannihilation:
The current [specification](https://maplibre.org/maplibre-tile-spec/specification/#geometry-column) supports 3D geometry:

> The x, y, and optional z coordinates are stored interleaved in a VertexBuffer for efficient CPU processing and direct copying to GPU buffers. If the z coordinate is not needed for rendering, it can be stored separately as an M-value (see vertex-scoped properties).

However, as far as I can tell, none of the current implementations support 3D geometry in actual. I know there's [an ongoing discussion about 3D support](https://github.com/maplibre/maplibre-tile-spec/discussions/1182), but I believe the MLT v1 spec is ready for supporting simple `(x, y, z)` coordinate.

This pull request is the proof of concept; while this is naively done by AI, I believe the implementation would be like this anyway if I implement all by hand. The changes are a bit lengthy, but the points are simple:

- Use the geo-traits crate for 3D coordinates (as you know, the geo-types only supprts 2D)
  - As geo-traits doesn't provide an actual `struct` for 3D geometry, use the wkt crate for this.
- Add `GeometryZ` to the column type, and treat `Geometry` and `GeometryZ` differently.

I also have [the corresponding TypeScript implementation](https://github.com/yutannihilation/maplibre-tile-spec/compare/feat/rust-3d...yutannihilation:maplibre-tile-spec:feat/ts-3d), and it works fine. So, I don't see any blocking issue on supporting 3D coordinates in this way.

Are there any particular reasons why the current implementations don't support the optional z coordinate? I don't expect this pull request to be merged as-is, but I'm hoping it can serve as a starting point for discussion — I'd love to know whether this is heading in the right direction!


CommanderStorm:
> and optional z

Since adding the new type would likely be a bit breaking for existing consumers of mlt@v1, we need to remove this part and re-add it in v2 😉
Which means we can do it properly and get the 3D data support right imo.

Would this be a good alternative for you?

yutannihilation:
Thanks. I don't have strong opinion here, but let me clarify a bit for discussion.

> Since adding the new type would likely be a bit breaking for existing consumers of mlt@v1

In my understanding, adding *the ability* to decode/encode `GeometryZ` shouldn't hurt anything; it's not like adding a new field. As the X and Y coordinates are already stored interleaved in a `VertexBuffer`, adding Z doesn't need additional field.

If you mean that the TypeScript decoder fails when the data contains `GeometryZ`, I don't think this is a problem. Not all decoders keep up with all encoders. For example, not all decoders implement FastPFOR, but we find FastPFOR is useful.

That said,

>  we need to remove this part and re-add it in v2

this sounds good to me. One of the things I feel uneasy with is the gap between the specification and the implementation, so it sounds a reasonable idea to tweak the specification to match with the implementation. If v2 will come in the foreseeable future, I'm happy with your idea!

CommanderStorm:
I have a hunch that I might get the time to work towards V2 after my exams this way or that. ^^

yutannihilation:
Good luck on your exam!

nyurik:
@yutannihilation I didn't go too far into the implementation yet, but I do want to mention a few direction thougths.

* we should not make any "logical" changes to MLTv1 -- v1 should have a feature parity with MVT - i.e. it should be possible to roundtrip MVT->MLTv1->MVT or to convert any MLTv1 to MVT without data loss. There are a few minor exceptions to this due to columnar nature of the storage, but it should work for pretty much every regular tile. Moreover, the set of v1 compressions and specifications should never change - so if a lib declares MLTv1 support, every single v1 will be supported going further.
* MLT has a framing concept - so it is possible to mix in multiple different frames one after another - so if v2 supports 3D, all parsers will know what to do -- either parse it (if they support v2), or skip it (because they know the size)
* we are currently trying to figure out the v2 support in #1201 - it could include the Z coordinates as either additional geometry sub-stream or as M-value (or both). I'm all for figuring it out. We will just need to make sure our renderers would be able to handle it properly.  And of course, as you mentioned, there is a long desired but highly uncertain #1182 - we do want the proper 3D, we just don't have a finalized implementation. It will most likely have to go into a separate frame number, e.g. v3

yutannihilation:
Thanks for sharing the directions! But, let me make it clear about this point.

> we should not make any "logical" changes to MLTv1

I don't change anything. The current MLT specification supports 3D coordinates, and I just naively implemented it. I understand "a feature parity with MVT" is your current consensus on v1, but it's not what the specification says.

nyurik:
@yutannihilation I am not saying you did anything wrong :)  The specification was written long before we started v1 implementation, and it was initially trying many things at once. Later, we tried to clean it up to limit just the MVT-logical-compat for v1 - but we might have missed a thing or two - that should be clearly marked as the future development rather than part of the v1 specification.  I will update it to clearly state that the current v1 implementation is only 2D, whereas we can easily add the 3D support later, using either the method mentioned in the spec, or we could alter it significantly - none of the z support is set in stone yet.

nyurik:
See #1509

yutannihilation:
Ah, makes sense. Thanks for the context!

yutannihilation:
I forgot to explain one caveat in this implementation; in order to use only what the specification document says, this implementation stores Z coordinate as-is. But, it's not obvious what this value should be translated.

X and Y are in-tile coordinates. This means these have implicit scale and offset, which are uniquely determined by the combination of X, Y, and zoom level. But, it's not the case with Z coordinate. So, I wish v2 define some metadata like `z_scale` and `z_offset`.

CommanderStorm:
needs to be rebased onto v2, which is my currently open stack at least.
Plus the layout byte needs to be used for the tracking what kind of geom this is, there are two variants, maybe polygonZ and linez or geomeetryZ and an extension, not sure

yutannihilation:
Yeah, I agree this pull request is now superseded with your ongoing proposal on the layout bytes. I'm happy that I'm proven wrong on this!

PR #1508, opened 2026-07-15
fix(java): Fix handling of geometrycollection and empty geometries

TimSylvester:
Features with empty geometry still got an ID assigned, leading to a mismatch on decoding.  The geometry collection type wasn't handled at all.


CommanderStorm (approved):
Lgtm

PR #1505, opened 2026-07-13
fix(rust): gate trigram-based shared-dict merging on corpus size

CommanderStorm:
Fixes shared dictionary grouping inflation on datasets with many high-cardinality string columns (e.g. OSM hiking relations with `@id`, `relation_ids`, `relation_lengths_m`).

`max(exact, trigram) > 0.075` merges columns that share character trigrams (digits, URLs) despite sharing zero actual values. Through transitive union-find closure, nearly all string columns get merged into one giant dictionary.

**Fix**:
After forming groups with the existing aggressive heuristic, each group whose combined corpus exceeds 100 KB is checked for actual cross-column value deduplication. Groups with <5% dedup are dissolved.
Tjios means that existing OMT groupings are preserved, since they have genuine value overlap (multilingual name translations share most values across `name`, `name:latin`, `name_de`, `name_en`, `name_int`):

| Tile | Before | After | Change |
|---|---:|---:|---:|
| OMT z14 (23K features) | 378.1 KB | 378.1 KB | **+0.0%** |
| OMT z7 | 178.1 KB | 178.1 KB | **+0.0%** |
| Hiking z0 | 180.0 KB | 155.5 KB | **−13.6%** |
| Hiking z2 | 678.9 KB | 555.2 KB | **−18.2%** |
| Hiking z4 (86K features) | 5.2 MB | 4.4 MB | **−14.8%** |
| Hiking z6 | 3.4 MB | 2.9 MB | **−16.0%** |
| Hiking z8 | 543.0 KB | 445.7 KB | **−17.9%** |
| Hiking z10 | 173.6 KB | 138.6 KB | **−20.2%** |

PR #1512, opened 2026-07-16
perf(ts): cache decoded FSST shared dictionaries

candux:
## Summary

MLT `SharedDict` columns store strings for multiple child property columns in one dictionary corpus. When the corpus is FSST-compressed, the TypeScript decoder exposes each child as a separate `StringFsstDictionaryVector`.

Previously, the first access to each vector decoded the complete shared FSST corpus again. Consumers such as MapLibre GL JS may access several of these vectors while materializing properties for styles, filters, or `queryRenderedFeatures()`.

This PR:

- Adds a benchmark for ordinary and shared FSST dictionary access.
- Creates one lazy decoded-dictionary cache per `SharedDict`.
- Shares that cache only between the vectors belonging to that group.
- Keeps ordinary FSST vectors on their existing direct decoding path.
- Ties the cache lifetime to the vectors and tile without a module-level `WeakMap`.
- Tests that two vectors from one FSST `SharedDict` call `decodeFsst()` only once.

#1505 makes this optimization slightly less relevant by preventing some oversized shared-dictionary groups with little actual value overlap, but beneficial `SharedDict` groups remain.

## Benchmark

The benchmark uses the real FSST dictionary from `14_8299_10748.mlt`, which does not contain a `SharedDict`, to construct controlled cold-access scenarios.

`xlarge.mlt` contains a real shared FSST dictionary, but reusing its vectors would warm the cache after the first iteration. Re-decoding the entire 5.4 MB tile for every iteration would make full-tile decoding dominate the cache overhead being measured.

Node 24.11.1 results, averaging the mean latency from two runs:

| Cold access | Before | After | Result |
| --- | ---: | ---: | ---: |
| One ordinary FSST vector | 0.4117 ms | 0.4057 ms | No regression |
| Two ordinary FSST vectors | 0.8146 ms | 0.8192 ms | No regression |
| Two vectors from one `SharedDict` | 0.8048 ms | 0.4132 ms | 1.95× faster |
| Five vectors from one `SharedDict` | 1.9826 ms | 0.4123 ms | 4.81× faster |

## AI notice

This PR was developed with assistance from OpenAI Codex.

### DIFF
--- a/ts/src/decoding/stringDecoder.ts
+++ b/ts/src/decoding/stringDecoder.ts
@@ -11,7 +11,10 @@ import { decodeUnsignedInt32Stream, decodeLengthStreamToOffsetBuffer } from "./i
 import { type Column, ScalarType } from "../metadata/tileset/tilesetMetadata";
 import { decodeVarintInt32 } from "./integerDecodingUtils";
 import { decodeBooleanRle, skipColumn } from "./decodingUtils";
-import { StringFsstDictionaryVector } from "../vector/fsst-dictionary/stringFsstDictionaryVector";
+import {
+    type FsstDictionaryCache,
+    StringFsstDictionaryVector,
+} from "../vector/fsst-dictionary/stringFsstDictionaryVector";
 
 export function decodeString(
     name: string,
@@ -204,6 +207,8 @@ export function decodeSharedDictionary(
 
     const childFields = column.complexType.children;
     const stringDictionaryVectors = [];
+    /** Shared by every FSST child-column vector in this SharedDict and populated on first access. */
+    const sharedDictionaryCache: FsstDictionaryCache | undefined = symbolTableBuffer ? {} : undefined;
     let i = 0;
     for (const childField of childFields) {
         const numStreams = decodeVarintInt32(data, offset, 1)[0];
@@ -259,6 +264,7 @@ export function decodeSharedDictionary(
                   symbolOffsetBuffer,
                   symbolTableBuffer,
                   presentStreamBitVector,
+                  sharedDictionaryCache,
               )
             : new StringDictionaryVector(
                   columnName,
--- a/ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.bench.ts
+++ b/ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.bench.ts
@@ -1,0 +1,92 @@
+import { readFileSync } from "node:fs";
+import { bench, describe } from "vitest";
+import decodeTile from "../../mltDecoder";
+import type BitVector from "../flat/bitVector";
+import { StringFsstDictionaryVector } from "./stringFsstDictionaryVector";
+
+type FsstVectorData = {
+    indexBuffer: Uint32Array;
+    offsetBuffer: Uint32Array;
+    dataBuffer: Uint8Array;
+    symbolOffsetBuffer: Uint32Array;
+    symbolTableBuffer: Uint8Array;
+    nullabilityBuffer: BitVector;
+};
+
+describe("cold FSST dictionary access", () => {
+    const tileData = readFileSync(new URL("../../../../test/expected/tag0x01/omt/14_8299_10748.mlt", import.meta.url));
+    const featureTable = decodeTile(new Uint8Array(tileData.buffer, tileData.byteOffset, tileData.byteLength)).find(
+        (table) => table.name === "poi",
+    );
+    const dictionaryVector = featureTable?.propertyVectors.find((vector) => vector?.name === "name");
+    if (!(dictionaryVector instanceof StringFsstDictionaryVector)) {
+        throw new Error("FSST cache benchmark dictionary poi.name not found");
+    }
+
+    // Access internal buffers to construct cold vectors around the same production dictionary.
+    const vectorData = dictionaryVector as unknown as FsstVectorData;
+    const valueIndex = Array.from({ length: dictionaryVector.size }, (_, index) => index).find((index) =>
+        dictionaryVector.has(index),
+    );
+    if (valueIndex === undefined) throw new Error("FSST cache benchmark dictionary contains no values");
+
+    let decodedLengthChecksum = 0;
+    const createVector = (dataBuffer: Uint8Array) =>
+        new StringFsstDictionaryVector(
+            "name",
+            vectorData.indexBuffer,
+            vectorData.offsetBuffer,
+            dataBuffer,
+            vectorData.symbolOffsetBuffer,
+            vectorData.symbolTableBuffer,
+            vectorData.nullabilityBuffer,
+        );
+
+    bench(
+        "one ordinary FSST vector",
+        () => {
+            // A fresh buffer keeps every benchmark iteration cold.
+            const vector = createVector(vectorData.dataBuffer.slice());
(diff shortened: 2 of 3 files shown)


ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
```diff
@@ -16,18 +18,20 @@ export class StringFsstDictionaryVector extends VariableSizeVector<Uint8Array, s
         private readonly symbolOffsetBuffer: Uint32Array,
         private readonly symbolTableBuffer: Uint8Array,
         nullabilityBuffer: BitVector,
+        private readonly dictionaryCache?: FsstDictionaryCache,
```
CommanderStorm: so `dictionaryCache` and `decodedDictionary` is optional?
And why is this not `Uint8Array` directly?

CommanderStorm (commented):
Caching this is likely fine, I just don't understand why the type was done this way

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
candux: Yes. `dictionaryCache` is only provided for vectors belonging to a `SharedDict`. Its `decodedDictionary` is initially absent because decoding remains lazy; the first accessed vector fills it and the others reuse it.

It cannot be passed as a `Uint8Array` directly because that would require decoding eagerly. The small mutable cache object lets multiple vectors share a result that is created only on first access. I renamed it to `sharedDictionaryCache` to make that scope clearer.

CommanderStorm: I don't fully follow.

but why do we have `sharedDictionaryCache?.decodedDictionary?` that does not quite make sense to me.
Would `sharedDictionaryCache?: Uint8Array` not be the exact same in terms of code and better? are we planning to add something there that I am missing?

candux: The wrapper is needed as shared mutable state.

All StringFsstDictionaryVector instances for the child columns of one SharedDict are constructed before the dictionary is decoded, so they receive the same initially empty holder. On first access, one instance decodes the dictionary and stores the resulting Uint8Array in that holder; the other instances can then reuse it.

If the parameter were sharedDictionaryCache?: Uint8Array, assigning the decoded array would update only that instance’s property—the other instances would remain undefined. Passing the array directly would therefore require decoding the dictionary eagerly.
I added some explanatory comments.

CommanderStorm: Okay, I let claude explain it for me. (should have likely done that the first time)

- Outer `sharedDictionaryCache?` -> "is this vector a member of a shared-dict group?" (absent for ordinary FSST vectors)
- Inner `decodedDictionary?` -> "has the corpus been decoded yet?" (lazy; undefined until first access)

CommanderStorm (approved):

ts/src/decoding/stringDecoder.ts
```diff
@@ -204,6 +207,8 @@ export function decodeSharedDictionary(
 
     const childFields = column.complexType.children;
     const stringDictionaryVectors = [];
+    // Give every FSST child-column vector in this SharedDict the same holder without decoding the dictionary eagerly.
```
HarelM: Use tsdocs instead.

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.bench.ts
```diff
@@ -0,0 +1,98 @@
...
+    indexBuffer: Uint32Array;
+    offsetBuffer: Uint32Array;
+    dataBuffer: Uint8Array;
+    symbolOffsetBuffer: Uint32Array;
+    symbolTableBuffer: Uint8Array;
+    nullabilityBuffer: BitVector;
+};
+
+describe("cold FSST dictionary access", () => {
+    // xlarge.mlt contains a real shared FSST dictionary, but reusing its vectors would warm their cache after one
```
HarelM: Try and only keep the test name, this comment us too verbose...

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
```diff
@@ -36,6 +43,14 @@ export class StringFsstDictionaryVector extends VariableSizeVector<Uint8Array, s
         return decodeString(this.decodedDictionary, start, end);
     }
 
+    private decodeDictionary(): Uint8Array {
+        if (this.symbolLengthBuffer == null) {
+            // TODO: change FsstEncoder to take offsets instead of length to get rid of this conversion
```
HarelM: Can we avoid adding todos?

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
```diff
@@ -16,18 +18,23 @@ export class StringFsstDictionaryVector extends VariableSizeVector<Uint8Array, s
         private readonly symbolOffsetBuffer: Uint32Array,
         private readonly symbolTableBuffer: Uint8Array,
         nullabilityBuffer: BitVector,
+        // This mutable holder is shared by StringFsstDictionaryVector instances for the child columns of one
```
HarelM: Use tsdocs with shorter commets.

ts/src/decoding/stringDecoder.spec.ts
```diff
@@ -365,28 +366,40 @@ describe("decodeSharedDictionary", () => {
     });
 
     describe("FSST encoding", () => {
-        it("should decode FSST-compressed shared dictionary", () => {
+        afterEach(() => {
+            vi.restoreAllMocks();
+        });
+
+        it("should decode an FSST-compressed shared dictionary once", () => {
+            const decodeFsst = vi.spyOn(fsstDecoder, "decodeFsst");
```
HarelM: This is too specific to the class implemetation, since you added a benchmark test I would recommend removing this change in test.

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
candux: That's much easier to understand. I updated the comments

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.bench.ts
candux: done

ts/src/decoding/stringDecoder.ts
candux: done

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
candux: done

ts/src/vector/fsst-dictionary/stringFsstDictionaryVector.ts
candux: it was there before already. I removed it now

ts/src/decoding/stringDecoder.spec.ts
candux: Agreed. I removed it

PR #1513, opened 2026-07-16
refactor(rust): carve v1 wire-format seams to prep for v2

nyurik:
Prepare for v2 with no wire-format or behavior change

- Convert RawPresence from a tuple struct over Option<RawStream> into an enum (AllPresent / Stream) and centralize all presence decoding behind RawPresence::decode_bits / decode_bools. v2 can then add raw-bitfield and shared-presence-group variants without touching downstream code.
- Move the v1 stream-header wire codec (stream_type byte, encoding byte, counts, RLE/Morton extras) into decoder/stream/header01.rs, absorbing the stream_type byte codec from physical.rs. StreamMeta / RawStream / IntEncoding constructors move to model.rs as format-independent in-memory types, so a future header02 codec can sit alongside.
- Rename parse_struct_children to parse_shared_dict_children to match shared-dictionary terminology used elsewhere (per #1201).


rust/mlt-core/src/decoder/stream/header01.rs
```diff
@@ -1,56 +1,84 @@
...
+//! ```text
+//! [u8     stream_type]      category / subtype nibbles
+//! [u8     encoding]         logical1 (bits 7-5), logical2 (bits 4-2), physical (bits 1-0)
+//! [varint num_values]
+//! [varint byte_length]
+//! [varint runs]             RLE streams only (non-bool)
+//! [varint num_rle_values]   RLE streams only (non-bool)
+//! [varint bits]             Morton streams only
+//! [varint shift]            Morton streams only
+//! ```
```
CommanderStorm, line 14: this being in the md docs instead would be nice

CommanderStorm (approved):

PR #1516, opened 2026-07-16
chore(rust): dedup some code with generics

nyurik:
## Summary by CodeRabbit

* **Improvements**
  * Stream decoding now uses a unified approach across 32-bit and 64-bit values.
  * Supports decoding raw little-endian data, FastPFOR-compressed data, and variable-length encoded values through a consistent interface.
  * Invalid or unsupported physical encodings are handled more reliably.
* **Compatibility**
  * Replaced specialized 32-bit and 64-bit decoding methods with a generic decoding interface.


CommanderStorm (approved):
don't have a strong oppinon, thought previous code was also fine

PR #1517, opened 2026-07-17
chore(ci): add github token to avoid CI download failures

nyurik:
## Summary by CodeRabbit

* **Chores**
  * Updated the Python release workflow to improve authentication for Linux, macOS, and source distribution builds.
  * Updated the CI paths-check workflow to explicitly set permissions for repository access and provide an authentication token to the paths filter step.
  * No changes to application functionality or release triggers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,9 @@ on:
 jobs:
   changes:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):
we can try

nyurik:
sigh, turns out it already uses github token it seems :'(

nyurik:
no, seems like this is a noop - both actions use github_token as the default

PR #1522, opened 2026-07-17
fix(docs): Fix a typo about geometry column layout

yutannihilation:
The spec says

> The geometry column uses a Structure of Arrays (SoA) layout 

but I believe this is a typo of "Array of Structure".


CommanderStorm:
No, the wording is accurate I think.
Different streams, one after another -> Structure of arrays..

yutannihilation:
Ah, got it... Sorry for the noise!

CommanderStorm:
No worrries and thanks for asking the question 😉

PR #1541, opened 2026-07-28
test(synthetics): add integer physical NONE fixture

DoFabien:
This is a small follow-up to #1138.
It adds a minimal Rust-generated synthetic tile with a single `u32` property encoded using `physical: NONE`.

### Java and C++

Integer `NONE` decoding is not implemented in Java or C++ yet, so the fixture is currently marked as unsupported in both implementations.

For Java, I followed the same idea as in C++: the fixture is still executed and the test checks that decoding fails instead of silently skipping it. I am not completely sure this is the preferred way to handle an unsupported synthetic in the Java test suite, so feedback on this part would be appreciated.


CommanderStorm (approved):

michaelkirk:
If I'm understanding, [this](https://github.com/maplibre/maplibre-tile-spec/pull/1138) allows TS to decode the MLT rust started creating as of #1466  👍

CommanderStorm:
Yes, just cpp nessary (or merging the rust decoder PR)

PR #1542, opened 2026-07-28
fix(ts): preserve nullable uint64 ID alignment in number mode

DoFabien:
Nullable 64bit IDs were shifted in the TypeScript decoder when `decodeTile` used number mode. The compact values were decoded correctly, but null positions were not restored before creating the ID vector, so every ID after a null could be assigned to the wrong feature.

This change passes the nullability buffer through the number decoding path and uses the existing `unpackNullable` helper to restore the original feature indexes.

## Tests

I added a focused unit test for nullable `uint64` values and synthetic regression tests using the existing `ids64_opt.mlt` and `ids64_opt_delta.mlt` fixtures. The synthetic tests cover the public `decodeTile` path in number mode.


CommanderStorm (approved):
LGTM, maybe we can make the ts/src/synthetic.spec.ts smarter to automatically test for this, not sure.
Not this PR

ts/src/synthetic.spec.ts
```diff
@@ -43,11 +43,21 @@ describe("MLT Decoder - Synthetic tests", () => {
     for (const { name, content, fileName } of testCases.skipped) {
         it(`${name} (unsupported)`, () => expectUnsupported(() => decodeMLT(fileName), content));
     }
+
+    for (const name of ["0x01/ids64_opt", "0x01/ids64_opt_delta"]) {
```
HarelM: Not sure I fully understand this, can you clarify what this is?

DoFabien: Sorry, this is not very clear. For these uint64 IDs, the tests above only cover the BigInt path. This loop reruns the two existing nullable fixtures in number mode, one with NONE and one with DELTA, to check through decodeTile that null positions stay aligned

HarelM:
Reading this again I'm still confused, sorry. 
How would the person using the API know if they should pass true or false? 
If the unit test and current synthetic test cover this code flow then I don't think there's a need to add "special" synthetic tests, the point of synthetic tests is "end to end", so from a user point of view I should send a tile and get back data, nothing else, in theory.

DoFabien:
Yes, I was overcomplicating the synthetic tests. I've removed the special cases and kept the focused unit test. The boolean predates this fix, but I agree that its contract is confusing. I have a follow-up PR prepared where the default decoding automatically uses number when safe and falls back to BigInt otherwise.

HarelM (approved):

HarelM:
THANKS!

HarelM:
@DoFabien any chance you can take a look at the two PRs I opened (ts encoder and nested property support) and let me know what you think?

PR #1543, opened 2026-07-29
Fix(ts) Morton dictionary decoding MULTIPOINT

DoFabien:
The TypeScript decoder treated `MULTIPOINT` entries from a Morton dictionary as regular interleaved coordinates, producing invalid points.

This change makes `MULTIPOINT` use the same dictionary aware decoding path as the other geometry types.

Added a focused unit test and a synthetic fixture covering a Morton dictionary with MULTIPOINT geometry.


ts/src/vector/geometry/geometryVectorConverter.ts
```diff
@@ -258,32 +263,26 @@ export function convertGeometryVector(geometryVector: GeometryVector): Coordinat
...
 
-function decodeDictionaryEncodedLineStringOrRing(
+function decodeDictionaryEncodedVertices(
     vertexBufferType: VertexBufferType,
     vertexBuffer: Int32Array | Uint32Array,
     vertexOffsets: Uint32Array,
     vertexOffset: number,
     numVertices: number,
-    closeLineString: boolean,
+    closeRing: boolean,
```
michaelkirk: Is there such a thing as an unclosed ring? Previous name was better IMO.

michaelkirk (commented):
Seems like a good catch. I only have a nitpick, but I'm not very familiar with the TS code.

ts/src/vector/geometry/geometryVectorConverter.ts
michaelkirk: I guess the point is that this might not be a line string at all, so maybe isRing?

It's probably fine how you had it tbh

DoFabien: Yes, exactly. The flag is only true when decoding polygon rings, which is why closeLineString felt a bit ambiguous to me. I initially chose closeRing to describe the action, but I agree that isRing expresses the intent more clearly here.

CommanderStorm (approved):
thanks

HarelM (approved):

PR #1553, opened 2026-08-06
feature(ts): support nested properties in ts decoder

HarelM:
Adds the relevant decoding code for nested properties 
Adds encoding code for unit tests and removed the skipped tests.
To be honest, I'm not sure what gl-js will do with these values, but that doesn't mean the decoder can't support them...


ts/src/vector/flat/objectFlatVector.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Unlike the other vectors there is no packed buffer to index into: nested property (MAP) columns
+ * decode to plain JavaScript maps, arrays and scalars, so the values are kept as-is. A `null` entry
+ * means the property is absent for that feature.
+ */
+export class ObjectFlatVector extends Vector<Uint8Array, unknown> {
+    constructor(
+        name: string,
+        private readonly values: unknown[],
+    ) {
+        super(name, new Uint8Array(0), values.length);
```
DoFabien: I think this loses the nullability information. A null value means that the property is absent, but because only values.length is passed here, `Vector.has(index)` always returns true.
Could we preserve the nullability here and add a focused test checking both `getValue()` and `has()` for an absent nested property?

ts/src/decoding/mapPropertyDecoder.ts
```diff
@@ -0,0 +1,372 @@
...
+    return decodeMapEntries(flattenedValues, startIndex, endIndex, dictionary);
+}
+
+function decodeMapEntries(
+    flattenedValues: Uint32Array,
+    startIndex: number,
+    endIndex: number,
+    dictionary: MapValue[],
+): Decoded<Record<string, MapValue>> {
+    const value: Record<string, MapValue> = {};
```
DoFabien: I think there is a small edge case here. Since key comes from the tile data, it can be `__proto__`. On a regular object, `value[key] = ...` would then change the object's prototype instead of storing the decoded value. Could we use `Object.create(null) ` here ?

DoFabien:
Thanks for this PR! I am really interested in this support because I have a real use case for it in my main application.

I display data where many records can share exactly the same geometry. In dense areas, one geometry can have several hundred records linked to it, and a tile can contain several hundred thousand records in total.

Today, I have to represent them as many features with the same geometry. This increases the size of the data, but also the amount of work MapLibre GL JS has to do, especially for filters and label placement.

I would prefer to group this data into a single feature, for example:

```json
{
  "item_count": 200,
  "latest_value": 250000,
  "items": [
    {"date": "...", "value": 250000},
    {"date": "...", "value": 210000}
  ]
}
```

With `get`, `at` and `length`, we can already use a good part of this structure in the style: count the items, read the first or last one, access a value, filter data, or build a label.

So overall, I really like the direction of this PR.

My main concern is memory usage and decoder performance. If I understand the code correctly, the streams are first decoded into a fairly compact form. But then all nested values for all features are immediately turned into JavaScript objects and arrays, and stored in `ObjectFlatVector`.

For small nested properties, this is probably fine. But in my case, it could create several hundred thousand objects for a single tile, with many allocations and more work for the garbage collector.

For a first version, this approach looks reasonable to me. I see `ObjectFlatVector` as a good first implementation, but maybe not as the final solution.

In the future, it would be good to keep the data in a compact form and only create the object when a feature is read. Ideally, we might even be able to read something like `items[0].value` without building the whole array.

This is quite close to the work I started in #757 around lazy decoding.

Later, I think it would also be useful to look at the MapLibre GL JS integration, so that it does not always materialize every feature. Lazy decoding is really useful only if the data stays lazy until a value is actually requested.

MLT decoding is more complex than MVT decoding. I think we will get the most value from MLT if we make full use of the columnar format: keep the data compact, reduce allocations, and decode only what is actually used. If GL JS calls `getFeatures()` and rebuilds everything right away, we lose a large part of those benefits. My main point here is to keep this option open for the future.

So for me, this is a very good starting point and really useful support.

HarelM:
Thanks for looking into this! Since the style spec doesn't support getting objects out of properties I don't think it will be very useful besides allowing to store more complicated data, but as a format it's good to be able to support that. I agree about the possibility to make future optimization, and keeping it in mind for the future.

ts/src/metadata/tile/mapControlValue.ts
```diff
@@ -0,0 +1,15 @@
...
+ * structure; anything else is an index into the combined dictionary, offset by `COUNT`. Booleans are
+ * encoded directly as tokens rather than being added to a dictionary.
+ */
+export enum MapControlValue {
+    FALSE = 0,
+    TRUE = 1,
+    /** A nested map follows: this token, the payload length including these two tokens, the payload. */
+    START_MAP = 2,
+    /** A list follows, laid out the same way as START_MAP. */
+    START_LIST = 3,
```
CommanderStorm, line 12: this reminds me of a bunch of very hard to work with formats, especially since true/false/start/count is in the same bucket..
Seems a bit weird.. not sure if this is 100% how we should do things, but that is  a different thing since you did not invent this.

CommanderStorm (approved):
LGTM

ts/src/vector/flat/objectFlatVector.ts
HarelM: Done in f011352

ts/src/decoding/mapPropertyDecoder.ts
HarelM: Done in f011352

PR #1554, opened 2026-08-06
feature(ts): add encoder

HarelM:
This is a very simple, not optimized encoder written in typescript.
It passes all the synthetic tests (besides v2 obviously).
Most of the code was already there this was my initial vision around unit tests using all the encoding code.
The only thing left was to write them all down in the right order.
The following verification was made: decode mlt -> encode mlt (using typescript) -> decode mlt -> compare.
This allowed to make sure the encoding is working well enough (it might not produce all the possible encoding permutations, which is fine).
Overall, not a lot of code.


CommanderStorm:
On an conceptual level, I am not too happy about adding another encoder, similarly to the cpp encoder.

Adding another will mean that now bugs need to be fixed in 
- java
- rust
- cpp (if someone maintains it)
- js

Bugs occur pretty regularly as can be seen in the wonky plain encoding support.
Also this increses the burden for the future format by 1/3 (3 -> 4 encoders), so makes v2 harder.

What is the usecase for the pure-ts encoder?
Except that it is not a lot of code that is?
Where is it better compared to just wrapping the mlt-wasm lib by adding encoding wiring?

HarelM:
It's a lot less code to load which won't bloat the bundle size.
A use case I would like to utilize it, once nested properties land is geojson to mlt in order to send data from worker thread to main thread in maplibre.
Currently there's a nasty hack that I'd like to remove, and also start seeing how this capability can be used later on for M values and elevation in geojson. So I want to start seeing if this is a valid use case.

There's no increase in burden, if someone needs the relevant implemtation to support the relevant feature they can contribute it, or the encoder simply won't support it. No actual requirement, no feature. There's no bar we "have" to meet.

Fixing bugs in both encoders and decoders is not really any different, you need to do it either way. Forcing wasm/rust down people's throat is not a strategy that works, not for the cpp and not for typescript, sorry to disappoint you...

CommanderStorm (approved):
LGTM

PR #1571, opened 2026-08-21
feat(rust): implement basic header and metadata format

CommanderStorm:
squashed and rebased due to gh stack limitations


rust/mlt-core/src/decoder/stream/header02.rs
```diff
@@ -0,0 +1,363 @@
...
+use crate::utils::{BinarySerializer as _, parse_u8, take};
+use crate::{MltError, MltRefResult, MltResult, Parser};
+
+/// Bit 7 of the encoding byte: an explicit count varint follows.
+pub(crate) const HAS_EXPLICIT_COUNT: u8 = 0x80;
+
+/// Logical field (bits 6-4 of the encoding byte). Bit pattern 7 is reserved.
+#[derive(Debug, Clone, Copy, PartialEq, TryFromPrimitive)]
+#[repr(u8)]
+pub(crate) enum LogicalField {
```
nyurik, line 60: use binary rep maybe?  e.g. `0b0000_0010` and also if logical and physical will exist in the same byte, use the actual values that will be in the data - without shifting the value first.

rust/mlt-core/src/decoder/stream/logical.rs
```diff
@@ -35,6 +54,32 @@ impl RleMeta {
         Ok(result)
     }
 
+    /// Tag `0x02` layout: `(run_len, value)` pairs. The run count is derived from
+    /// the data length; `num_rle_values` comes from the stream's count context.
+    #[cfg(feature = "unstable-v2")]
```
nyurik, line 74: document somewhere in this file what RLE actually means - a vibed doc for both split and interleaved variants.  I tend to forget the exact format.

rust/mlt-core/src/decoder/model02.rs
```diff
@@ -0,0 +1,182 @@
...
+use num_enum::TryFromPrimitive;
+
+/// Column data type of a v2 property column, as stored in the tile.
+///
+/// Unlike v1's [`super::ColumnType`], geometry is not a column (the layer's
+/// geometry section precedes the counted columns), and each string encoding
+/// variant will get its own flat code instead of a runtime `stream_count`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromPrimitive)]
+#[repr(u8)]
+pub(crate) enum ColumnType02 {
```
nyurik: lets use `0x..` because it is all in a u8 bit space, and hex is easier for that, unless we share these with v1.  Also note that column type byte may include shared presence reference - e.g. half of   the byte is data type (id/int/float/... - but without the "Opt" variants), and half of it is a reference into shared presence columns or inline presence ("Opt").   One other thing - for shared dictionary types (sub-columns), we could have one a  presence column one per sub-column, so it wouldn't be even stored here i think, so you can reuse these bits for something else.

rust/mlt-core/src/decoder/root02.rs
```diff
@@ -0,0 +1,223 @@
...
+//! ```text
+//! [varint name_len] [name bytes]
+//! [varint extent]
+//! [varint feature_count]
+//! ── geometry section ─────────────────────────────────
+//! [u8 geometry_layout]              see GeoLayout
+//! [types stream]                    count = feature_count
+//! [length streams per layout]       explicit counts
+//! [vertex stream]                   explicit count
+//! ── counted columns ──────────────────────────────────
```
nyurik, line 21: i think there should be a shared presence section here - as it is separate from columns i think

rust/mlt-core/src/decoder/root02.rs
```diff
@@ -0,0 +1,223 @@
...
+use crate::decoder::stream::header02;
+use crate::decoder::{
+    ColumnType02, DictionaryType, Extent, GeoLayout, Id, Layer01, LengthType, RawGeometry, RawId,
+    RawIdValue, RawPresence, RawScalar, StreamType,
+};
+use crate::utils::{SetOptionOnce as _, parse_string, parse_u8, take};
+use crate::{Lazy, MltError, MltRefResult, MltResult, Parser};
+
+/// Parse a v2 layer body (the bytes after the `tag = 2` byte).
+pub(crate) fn parse_layer02<'a>(
```
nyurik, line 51: should this be part of the 02-specific file?

rust/mlt-core/src/decoder/root02.rs
```diff
@@ -0,0 +1,223 @@
...
+/// Parse the geometry section: layout byte + streams in the layout's fixed order.
+///
+/// Stream roles are assigned by position, mirroring the `stream_type` bytes the
+/// v1 encoder would have written, so [`RawGeometry`] decoding is shared.
+fn parse_geometry<'a>(
+    input: &'a [u8],
+    feature_count: u32,
+    parser: &mut Parser,
+) -> MltRefResult<'a, RawGeometry<'a>> {
+    let (input, layout_byte) = parse_u8(input)?;
```
nyurik: move layout_byte to the root instead of geometry only -- this way we can use some of its bits for other things like indicating if shared presence columns exist.  Moreover, if there are only 14 geo layouts (4 bits), we can actually store the number of presence columns here as well as 4 other bits.  Moreover, we may even restrict the number of shared dictionaries to 6 instead of 15 and leave one bit just in case we need to have some other flag.  I honestly don't think we will have this many SHARED presence columns.

rust/mlt-core/src/decoder/root02.rs
CommanderStorm: I think so, or? I refactored the modules so that there also now is root01.rs

nyurik (approved):
lgtm

PR #1580, opened 2026-08-25
chore(rust): split the geometry encoder into encode01 and streams

CommanderStorm:
Splits `encoder::geometry::encode` into `encode01` (v1 geometry column assembly) and `streams` (the shared stream/vertex-buffer helpers); pure move, no behaviour change.


nyurik (approved):
my only minor concern are all these `pub(super)` - do we really want to restrict them to one dir level, or would it be simpler to just use `pub(crate)` everywhere?

CommanderStorm:
if they get used outside something is very brokem. they are geom specific I think.
they are only needed due to v2 in a different geometry module.

Issue #12, opened 2023-08-15
Add support for nested properties


mactrem:
based on the Dremel encoding

Issue #753, opened 2025-12-16
RFC: How should MLT support nested `List` / `Map` types (format decision)?


DoFabien:
## Context / Motivation

I’d like to contribute support for nested property types—starting with `List<T>` and later `Map<K,V>`—in the official TypeScript decoder (this repo, `ts/`).

In parallel, I’m developing a Rust encoder that converts **GeoBuf**, **FlatGeoBuf**, and **GeoJSON** into MLT. I need `List` support (and likely `Map` later) to represent real-world attributes—for example:
- A parcel with multiple transaction years: `[2015, 2018, 2021]`
- A feature with repeated price history: `[{price: 100000, year: 2015}, {price: 150000, year: 2021}]`

That’s why I’m motivated to contribute this work upstream.

Before implementing anything, I need guidance on the **wire-format choice**, because the current Tag `0x01` format appears intentionally “MVT-equivalent”.

### Tag `0x01` type codes (current state)

Across the existing implementations (C++ / Java / TS / Rust parser), Tag `0x01` supports:

| Code range | Meaning |
|---|---|
| `0–3` | `ID` (32/64 + nullable variants) |
| `4` | `GEOMETRY` |
| `10–29` | scalar types (nullable via bit0) |
| `30` | `STRUCT` (only complex type with children) |

There is **no distinct `LIST` or `MAP` type code today**.

Relevant references:
- C++ type map: `cpp/include/mlt/metadata/type_map.hpp`
- TS type map: `ts/src/metadata/tileset/typeMap.ts`
- Java type map: `java/mlt-core/src/main/java/org/maplibre/mlt/converter/encodings/MltTypeMap.java`
- Rust (v01 parser): `rust/mlt-nom/src/v01/column.rs`
- Proto schema shows `ComplexType = { GEOMETRY, STRUCT }`: `spec/schema/mlt_tileset_metadata.proto`

### Implication

To support `List` / `Map` in the TS decoder, we likely need to **extend the format** (or introduce a new tag), otherwise there is nothing encoded to decode.

## Decision needed

What is the preferred approach for adding nested types (`List`/`Map`) to MLT while keeping compatibility expectations clear?

### Option A — Extend Tag `0x01`

Add new type codes for `LIST` and `MAP` (including nullable variants) and define their stream layouts.

Pros:
- Minimal framing change (still `tag == 1`)
- Easiest to iterate on in all decoders/encoders

Cons:
[...]

PR #1592, opened 2026-08-27
feat(cpp): decode nested properties (0x02 MAP columns)

johncarmack1984:
Part of #12.

The C++ decoder silently dropped any layer with tag 0x02 (decoder.cpp only accepted tag 1), so tiles with nested properties came back as empty layers. This adds MAP column support to the C++ decoder, following the Java (#1379) and TypeScript (#1553) implementations as the wire-format reference. The nine 0x02-java/prop_nested_* synthetics now decode and match, and the C++ skip list is empty.

Nested values are a small self-contained NestedValue tree in the public headers rather than nlohmann::json, because the Bazel build (what maplibre-native consumes) has MLT_WITH_JSON=0. Layer gains a mapProperties member, so anything linking a prebuilt mlt needs a rebuild. Encoder support is not included.

AI disclosure: AI tools were used to scan for repo conventions and iterate using idiomatic solutions. Comments explicitly de-vibed.


CommanderStorm:
v2 is not yet ready and we have not finalised the exact format for nested properties, so I would rather not like to tie us down in the decoders.
At least it needs to be feature flagged since we don't want to accidentally support something unstable and unready.

V2 is more than just nested properties as you can see in the open PR stack.

CommanderStorm (changes requested):
I would prefer we do go for nested properties after a good review of how this is currently done and if this a good idea. Code looks a bit hacky and I don't know if this is because of weird design or because of the code.

johncarmack1984:
Understood, parking makes sense to me.

On the code, I can guess where it looks hacky. The tag dispatch is a ternary between `Tag0x01` and `Tag0x02` at every call site instead of one seam, and `Tag0x02` is a struct that shadows its parent's static methods. That's me porting the Java shape instead of thinking about what a C++ decoder for this should look like, and it's where I'd start a rewrite.

The other bit is the format rather than the port: root-level maps have no `START_MAP` header, so the decoder decides scalar vs list vs map by token count and first token. That's probably worth a look in the design review. Related, tag 2 is currently doing two jobs, Java and TS ship it on main as the nested-property layout and the Rust v2 stack uses `Tag02` for the new format behind `unstable-v2`. Happy to help with either, including a writeup of the current Java layout on #753 if that makes the review easier.

CommanderStorm:
We need to first figure out what to do with nested properties. I am leaning towards shredding and reassebling.

PR #1603, opened 2026-09-01
feat(rust): encode and decode v2 float columns as ALP

CommanderStorm:
ALP stores a float as a scaled integer and recovers it exactly, so decimal data compresses where the raw bit patterns did not. Exception-free only: a column takes ALP when every value returns bit-for-bit, and it then competes with the dictionary on stored bytes.

### DIFF
--- a/rust/mlt-core/src/decoder/property/decode.rs
+++ b/rust/mlt-core/src/decoder/property/decode.rs
@@ -35,11 +35,17 @@ impl<'a> RawFloats<'a> {
     /// Decode the column, reading whichever stream set its encoding uses.
     fn decode<T>(self, dec: &mut Decoder) -> MltResult<ParsedScalar<'a, T>>
     where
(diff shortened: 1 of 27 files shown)


rust/mlt-core/src/decoder/property/decode.rs
```diff
@@ -35,12 +35,18 @@ impl<'a> RawFloats<'a> {
     /// Decode the column, reading whichever stream set its encoding uses.
     fn decode<T>(self, dec: &mut Decoder) -> MltResult<ParsedScalar<'a, T>>
     where
-        T: Copy + PartialEq + num_traits::FromBytes,
+        T: Copy + PartialEq + num_traits::FromBytes + crate::codecs::float::FloatValue,
```
nyurik, line 38: rm prefix! :)

nyurik (approved):

PR #1605, opened 2026-09-01
refactor: remove the never-implemented Pseudodecimal encoding

CommanderStorm:
No encoder in this repo ever emitted a Pseudodecimal (`PDE` / `PSEUDODECIMAL`) stream, and no decoder ever read one.
In v2, I am adding a better version based on ALP with a nice exception mechanism instead

-> 🗑️


HarelM (approved):
Approving the ts code changes.

PR #1609, opened 2026-09-02
feat(rust): encode and decode v2 string columns

CommanderStorm:
Adds the four v2 string column layouts, with the layout in the leading stream's extension bits so dictionary codes keep their integer codec.


birkskyum:
@CommanderStorm , just want to note that this stack is able to consume an immensse amount of CI runners. it's presently occupying 59/60 runners (1/60 on martin) and thus stalling all the other maplibre repos. I noticed it both late yesterday and today where tehre was a very long queue. I don't know if it's in the cases wher ehte btotom of the stack updates, and all above prs rerun, or if it's only in cases of force push. Im' not yet experienced with PR stacks and how to best handle them efficiently.  cc @louwers , you might be affected too , although native does rely more on dedicated runners.

CommanderStorm:
I think I can change the relese checks for cpp, rust to not run on stacks at all.
We have an realse PR anyhow for this.

Unfortnunately with yuri not being able to review this week things stack up..
I wish I could keep the stack less tall.

CommanderStorm:
had to rebase a few changes from the top of the stack to the bottom (they would have been fixup PRs ordinarily, but since I won't get review this week on the wire format changes..)

rust/mlt-core/src/decoder/stream/header02.rs
```diff
@@ -257,29 +326,54 @@ fn no_extension(enc_byte: u8) -> MltResult<()> {
     }
 }
 
+/// Read an integer stream's logical field, in the terms the integer and string families share.
+fn logical_int(family: Family, enc_byte: u8, logical: Logical) -> MltResult<LogicalInt> {
+    Ok(match logical {
```
nyurik, line 331: can you `use Logical::*` here? Makes it easier to read.

nyurik (approved):

PR #1624, opened 2026-09-03
feat(rust): share v2 presence bitfields with dictionary children

CommanderStorm:
Extends v2's per-layer shared presence-bitfield dedup to shared-dictionary children, which shrinks 1655 Berlin tiles from +2.03% to +0.075% versus v1.


CommanderStorm:
Reopening from an upstream branch so this can join the PR stack.

PR #1626, opened 2026-09-04
feat(rust): scrollbars on every mlt ui panel and a sideways tree scroll

### LINKED ISSUE #971: Improve `mlt` RataTUI tool
This is more of a wishlist of all improvements that should be done to the Rust TUI `mlt` tool

### Performance
* [ ] Recursive file discovery should be async, adding more and more files to the tile list. The UI should remain usable.
* [ ] In parallel to directory loading, as more files are added to the list, they should be async-parsed and their data shown in UI
* [ ] Parsed tiles should use LRE cache with some reasonable max memory, possibly overridable from CLI arg

### Stability
* [ ] Establish UI testing infrastructure

### Functionality
* [ ] Be able to visualize tessellation data when available - e.g. draw polygon triangles as gray lines. This should only work if the user selects a single polygon feature, or a sub-part of the multi-polygon feature, or on a polygon mouse-over(?)
* [ ] Add proper scrollbars to all panels where it makes sense. Relevant scrollbar should only appear if the content does not fit vertically or horizontally. Scrollbar span should match the content.
* [ ] When selection is on all layers, properties panel should say how many layers there are, and geometries should list all types of geometries in all layers
* [ ] When selection is on layer name, show feature count and list of properties and their types in the properties panel, and list of geometry types and their count in the geometry panel
* [ ] Mouseover a geometry in the UI should depend on which level is being viewed at the moment:
  * When viewing all layers together, mouseover should highlight all geometries in the **layer**, and the info panels should act as if layer root was selected. Clicking should select layer root.
  * When viewing layer root, mouse over should highlight individual features (but not subparts). Info panels should be as if feature was selected.  All other layers should be shown in gray, and should not be clickable or react to mouseover.
  * When viewing feature, mouseover should highlight individual subparts when available, or the whole feature if not. Other features should be shown as lighter gray (not clickable/selectable). Other layers should be shown as darker gray.

johncarmack1984:
Closes #971.

The rest of the split landed as #1638, #1642, #1643, #1646, #1647, and #1649. Every panel draws a vertical scrollbar when its content does not fit, with the thumb sized to the visible share, and the geometry panel scrolls like the others. The feature tree also scrolls sideways with Shift+Left/Right or a sideways mouse scroll, four columns per step so a long name takes a few presses, with a horizontal scrollbar when a row is wider than the panel.

### DIFF
--- a/rust/mlt/README.md
+++ b/rust/mlt/README.md
@@ -64,6 +64,8 @@ Features:
   - "All" - every layer, with layer and feature counts and the tile's geometry types
   - Individual layers - the layer's features, with each property's value types and the layer's geometry types
   - Individual features - the feature's properties and geometry statistics
+  - Every panel scrolls and shows a scrollbar when its content does not fit
+  - The tree also scrolls sideways with `Shift+Left`/`Shift+Right` or a sideways mouse scroll
   - Hovered rows are highlighted with underlined green text
 - **Map Panel (right)**: Visual representation of the geometries
   - Shows the extent boundary as a thin gray rectangle
--- a/rust/mlt/src/ui/mod.rs
+++ b/rust/mlt/src/ui/mod.rs
@@ -78,6 +78,8 @@ pub const STYLE_BOLD: Style = Style::new().add_modifier(Modifier::BOLD);
 
 /// Throttle hover-driven redraws so mouse move over map doesn't flood the loop
 const HOVER_REDRAW_THROTTLE: Duration = Duration::from_millis(32);
+/// Columns the feature tree moves per sideways scroll step.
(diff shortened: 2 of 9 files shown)


CommanderStorm:
> Happy to break it out into constituent parts for easier review.

yea.. could you do that? there is a lot in this change...

CommanderStorm:
I don't know how to scroll to the right, but the state is much better 👍🏻

johncarmack1984:
Fair, it was too hidden. Shift+Left/Right or a sideways trackpad scroll does it, and `h` opens the help. #1657 adds `<`/`>` and stops the title clipping `h:help`.

PR #1627, opened 2026-09-04
fix(rust): floor the mbtiles zoom level with a tolerance

johncarmack1984:
Found while implementing #1626.

In `mlt ui` on an `.mbtiles` file, zooming to an integer level through the scroll wheel lands the viewport width a rounding error wide of the exact power of two, so `zoom_level` floored one level too low: the title read `1/0/0` at zoom 2.0 and the coarser tiles were requested. The floor now allows a small tolerance. The test reaches zoom 2 through two half steps, which floored to 1 before


CommanderStorm (approved):
tnx

PR #1630, opened 2026-09-04
feat(rust): bit-pack v2 dictionary codes when that beats a varint each

CommanderStorm:
One lightweight encoding which is frequently used is bitpacking.. we are for some reason not using it.

---


CommanderStorm:
TODO: misses synthetics

rust/mlt-core/src/codecs/bitpack.rs
CommanderStorm, line 1: todo: check that we have 100% MCDC test coverage on all encoders

nyurik (approved):
check codecov for bitpacking

PR #1632, opened 2026-09-04
chore(CI): only run heavy jobs on top of the stack

CommanderStorm:
Gates the expensive CI jobs — `cpp-build`, `fuzz`, `build-release-binaries`, `build-wasm-npm` and the Python wheel builds — so they only run on the top pull request of a stack.

> [!CAUTION]
> Mid-stack pull requests lose this coverage entirely until they reach the top.
> - A mid-stack change to `cpp/` gets **no** C++ CI at all. No build, no tests, no Bazel, no coverage. 
> - Likewise, no fuzzing runs mid-stack, and
> - a change to `rust/mlt-core` that breaks the PyO3 bindings will not be caught until the top, since the wheel guard only path-matches `rust/mlt-py/**`.
> 
> Most of these are fine since on release we do run full CI. There might need to be some rework though if we merge things without looking at the top of the stack

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ jobs:
       java: ${{ steps.filter.outputs.java }}
       ts: ${{ steps.filter.outputs.ts }}
       cpp: ${{ steps.filter.outputs.cpp }}
+      # Only the top pull request in a stack runs the expensive artifact and fuzz jobs.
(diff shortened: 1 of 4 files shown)


CommanderStorm:
Reason for the change: We are currently consuming 59/60 of our runners 🤦🏻

birkskyum (approved):

PR #1638, opened 2026-09-04
test(rust): drive the mlt ui from tests through TestBackend

johncarmack1984:
Part of #971.

First piece of #1626, split out as requested; the rest stack on this one. The event loop's frame drawing, key handling, mouse handling, and between-frame pump move into functions that tests can drive, so the screens get insta snapshots through `TestBackend` and the handlers get unit tests. Three small fixes the tests turned up ride along in their own commit: the file table selected the row below the one clicked, the rows next to a horizontal divider grabbed the divider instead of their content, and opening another file kept the previous file's hover, which could index past the new file's layers and panic.

### DIFF
--- a/rust/mlt/src/ui/tests.rs
+++ b/rust/mlt/src/ui/tests.rs
@@ -1,0 +1,1786 @@
+use std::collections::BTreeMap;
+use std::path::PathBuf;
+use std::sync::mpsc;
+use std::time::Duration;
+
+use crossterm::event::{KeyCode, KeyEvent, KeyModifiers, MouseButton, MouseEvent, MouseEventKind};
+use mlt_core::geo_types::{
+    Coord, Geometry, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon,
+};
+use mlt_core::geojson::{Feature, FeatureCollection};
+use ratatui::Terminal;
+use ratatui::backend::TestBackend;
+use ratatui::layout::Rect;
+use serde_json::{Value, json};
+
+use super::{
+    MouseState, PanelAreas, SCAN_FLAGS, UiArgs, build_app, collect_extensions,
+    collect_file_algorithms, collect_file_geometries, extent_from_fc, find_tile_files,
+    handle_filter_click, handle_key, handle_mouse, load_fc, parse_center_tile_xyz, render_frame,
+    tick,
+};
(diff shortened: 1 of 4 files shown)


rust/mlt/src/ui/tests.rs
```diff
@@ -0,0 +1,1919 @@
...
+fn layer_overview_starts_on_all_layers() {
+    let mut app = sample_app();
+    insta::assert_snapshot!(render(&mut app), @r#"
+    "┌sample.mlt - Enter/+/-:expan┐┌Map View────────────────────────────────────────────────────────────┐"
+    "│>> All                      ││                                                                    │"
+    "│     Layer: water (2 feature││                                                                    │"
+    "│     Layer: roads (2 feature││     ⢰⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⢲⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⠒⢒⡲⡆     │"
+    "│     Layer: poi (2 features,││     ⢸                           ⢸                 ⡤⠤⠤⠤⠤⠤⠤⢤⡔⠁ ⡇     │"
+    "│                            ││     ⢸                           ⢸                 ⡇    ⡠⠊⠁⡇  ⡇     │"
+    "│                            ││     ⢸      ⢀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣸⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣀⣇⣀⣀⠔⠊   ⡇  ⡇     │"
```
CommanderStorm, line 166: looks kind of broken, but I think that is GH

CommanderStorm (approved):
Thanks, PR will merge, but we are currently in an bit of an CI-shortage due to this repos bad CI config (fix is in main or on PRs, but runners have not caught up yet)

rust/mlt/src/ui/tests.rs
johncarmack1984: Yep, GitHub's font. The braille cells come out a different width there; it lines up in a terminal.

PR #1640, opened 2026-09-04
chore(ci): consolidate and properly merge Coverage

CommanderStorm:
Currently, we run CI twice. Once under codecov, once without.
This PR adds an merging of coverage step and reports this to github instead...

Means no good line coverage instrumentation, but since we only intermittently build ("only what changed"), the line coverage is not reliable anyhow. 
Switching to GH means we can add a ruleset for merging on this 🎉

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,5 @@
+paths:
+  .github/workflows/**.yml:
+    ignore:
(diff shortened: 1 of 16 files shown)


HarelM (approved):
Not sure I fully follow, and this updates all CI runs, but I guess it's ok...

CommanderStorm:
reduces CI which seems to be very very short supply nowerdays.

PR #1642, opened 2026-09-04
feat(rust): scan the mlt ui directory in the background

johncarmack1984:
Part 2 of #971 / #1626.

The file browser used to walk the whole directory and analyze every file before drawing anything, so a large tree meant staring at an empty terminal. It now starts empty and fills in as a background walk finds tile files, each file is analyzed on the rayon pool and its row updates when the analysis lands, the title shows the progress, and the header sort waits for the scan to finish. Entries the walk cannot read are skipped instead of failing the whole browser

### DIFF
--- a/rust/mlt/README.md
+++ b/rust/mlt/README.md
@@ -52,8 +52,9 @@ cargo run -- ui path/to/directory
 ```
 
 **Directory Mode**:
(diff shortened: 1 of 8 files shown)


CommanderStorm (approved):
Works great, thanks

PR #1643, opened 2026-09-04
feat(rust): cache decoded tiles in the mlt ui

johncarmack1984:
Part of #971.

Third piece of #1626. Moving between files decoded the same tile again on every visit. Decoded tiles now go through a memory-bounded LRU cache shared by the preview, the layer view, and Enter; the budget is 256 MB, and the mbtiles map prunes its decoded tiles by the same budget instead of a fixed tile count, keeping the visible tiles and their ancestors. A preview that fails to decode now says so instead of retrying every frame

### DIFF
--- a/rust/mlt/README.md
+++ b/rust/mlt/README.md
@@ -58,7 +58,6 @@ cargo run -- ui path/to/directory
 - Press `Enter` to open and visualize a file
 - Press `Esc` to go back to file list
 - Press `q` to quit
-- Decoded tiles stay in an LRU cache (256 MB by default, `--cache-mb` changes it), so moving between files does not decode them again; the `.mbtiles` map keeps its decoded tiles within the same budget, dropping the ones farthest from the viewport first
 
 Features:
 - **Tree View Panel (left)**: Browse layers and features in a hierarchical tree
--- a/rust/mlt/README.md
+++ b/rust/mlt/README.md
@@ -58,6 +58,7 @@ cargo run -- ui path/to/directory
 - Press `Enter` to open and visualize a file
 - Press `Esc` to go back to file list
 - Press `q` to quit
+- Decoded tiles stay in an LRU cache (256 MB by default, `--cache-mb` changes it), so moving between files does not decode them again; the `.mbtiles` map keeps its decoded tiles within the same budget, dropping the ones farthest from the viewport first
 
 Features:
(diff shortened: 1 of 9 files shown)


rust/mlt/README.md
```diff
@@ -58,6 +58,7 @@ cargo run -- ui path/to/directory
 - Press `Enter` to open and visualize a file
 - Press `Esc` to go back to file list
 - Press `q` to quit
+- Decoded tiles stay in an LRU cache (256 MB by default, `--cache-mb` changes it), so moving between files does not decode them again; the `.mbtiles` map keeps its decoded tiles within the same budget, dropping the ones farthest from the viewport first
```
CommanderStorm:
```suggestion
```

rust/mlt/src/ui/mod.rs
```diff
@@ -81,6 +81,9 @@ pub struct UiArgs {
     /// Start `MBTiles` map centered on this XYZ tile (`z/x/y`, e.g. `6/32/21`). `MBTiles` only.
     #[arg(long = "center-tile", value_name = "Z/X/Y")]
     center_tile: Option<String>,
+    /// Memory budget for decoded tiles kept in the LRU cache, in megabytes.
+    #[arg(long = "cache-mb", value_name = "MB", default_value_t = DEFAULT_CACHE_MB)]
+    cache_mb: u64,
```
CommanderStorm: lets not add configurability for this.
I doubt people will use it.

Lets just set `DEFAULT_CACHE_MB` to something reasonable (I think it is currently) and be done 😉

CommanderStorm (changes requested):
Please inline the cache size.. it is kind of unnessary to have that kind of an CLI argument, since the intended usecase for `mlt ui` is quick exploration.

Otherwise, LGTM assuming you do the appropriate inlining that this change in direction means 😉

rust/mlt/src/ui/mod.rs
johncarmack1984: Dropped, 256 MB is a constant now. 55fbf6ee

rust/mlt/README.md
johncarmack1984: Removed in the same commit.

CommanderStorm (approved):
Nice work

PR #1645, opened 2026-09-04
chore(ci): run the heavy jobs on pull requests outside a stack

johncarmack1984:
`github.event.pull_request.stack` is only present on stacked pull requests, so the gate from #1632 evaluated to false for every standalone pull request and skipped cpp-build, fuzz, the release binaries and the wasm package on all of them, not only mid-stack. Every standalone pull request since then ran without them, #1634 to #1639 included, and Dependabot bumps are standalone too. A pull request outside a stack has nothing above it, so it gets the heavy jobs; inside a stack only the top does. The output is renamed to match the `run-heavy` input it feeds.

Run 33895274920 (#1638, created 16:27 UTC) shows cpp, Fuzz, Build release binaries and the wasm package all skipped.


CommanderStorm (approved):
thanks

PR #1646, opened 2026-09-04
feat(rust): summarize layers and the whole tile in the mlt ui info panels

johncarmack1984:
Part of #971.

Fourth piece of #1626. With all layers or a layer selected the info panels only said to select a feature. They now summarize the selection. For all layers that is the layer and feature counts and the tile's geometry types, for a layer each property's value types and the geometry-type counts. The panels describe the hovered item when there is one


CommanderStorm (approved):
🎉

PR #1649, opened 2026-09-04
feat(rust): draw tessellation triangles in the mlt ui

johncarmack1984:
Part of #971.

Sixth piece of #1626. Tiles that carry the `Triangles` and `IndexBuffer` streams get their triangles decoded next to the feature collection, one list per polygon part, so a tile's tessellation can be checked by eye instead of by reading index buffers. The map draws them in gray behind the selected polygon feature or part and behind a hovered polygon inside a layer, and the geometry panel shows the triangle count.

### DIFF
--- a/rust/mlt/src/ui/rendering/map.rs
+++ b/rust/mlt/src/ui/rendering/map.rs
@@ -10,10 +10,11 @@ use ratatui::widgets::canvas::{Canvas, Context, Line as CanvasLine, Rectangle};
 
 use crate::ui::mbt::{MbtHoveredInfo, MbtTileData, TileTransform};
 use crate::ui::state::{App, LayerGroup, TreeItem};
+use crate::ui::tile::{Tessellation, Triangle};
 use crate::ui::{
     CLR_CONTEXT_FEATURE, CLR_CONTEXT_LAYER, CLR_DIMMED, CLR_EXTENT, CLR_HOVERED, CLR_INNER_RING,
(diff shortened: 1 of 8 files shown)


CommanderStorm (commented):
Rendering seems to be a bit buggy regarding the wird blochy lines. not sure if this is this PR though. needs a bit of investigation

Image

rust/mlt/src/ui/rendering/map.rs
```diff
@@ -141,6 +144,74 @@ fn draw_other_layers(ctx: &mut Context<'_>, app: &App, except: usize) {
     }
 }
 
+/// Tessellation edges of a feature that are not already ring edges.
+/// `part` restricts a multipolygon to one polygon.
+fn draw_tessellation(
```
johncarmack1984: Yep, this PR. The triangle pass redrew the ring edges in gray over the colored rings, and a braille cell keeps the last color painted. Now only the interior edges are drawn, in dark gray. 6e31ffc1

johncarmack1984: Before and after on the same screen. The only gray left is the interior edges.

1649-tessellation-before-after

CommanderStorm: not really, I was refering to the random pink brotches on the gray lines.

CommanderStorm: maybe better in another PR. I see no reason for this to come from this PR

https://docs.rs/ratatui/0.30.2/ratatui/prelude/symbols/enum.Marker.html#variant.Octant

PR #1653, opened 2026-09-06
fix(ts): keep byte RLE within the lengths its headers can express

MaxFreedomPollard:
`encodeBooleanRle` in `ts/src/encoding/encodingUtils.ts` wrote the packed bits as one literal run with the header `256 - numBytes`. A literal header only spans 1 to 128 bytes, so from 1025 booleans upward it wrapped into the run range: 1032 booleans pack into 129 bytes and gave header 127, which `decodeByteRle` reads back as a run of 130 copies of the first packed byte. Every boolean column and every PRESENT stream longer than 1024 values decoded to garbage, and `propertyEncoder`, `stringEncoder` and `mapPropertyEncoder` all build their nullability streams through it. It now packs the bits and hands them to `encodeByteRle`, the same shape `encode_byte_rle` in `rust/mlt-core/src/codecs/rle.rs` uses.

Right next to it, `encodeByteRle` let a run grow to 131 bytes, clamped the header with `Math.min(header, 0x7f)`, and still advanced 131 positions. A run of 131 identical bytes came out as a run of 130 and lost the last byte. The scan now stops at 130, the longest run a header can name, and the clamp that hid the overflow is gone. The Rust encoder already stops at 130.

Two round trips in `ts/src/decoding/decodingUtils.spec.ts` cover both: 1032 booleans, and 131 identical bytes. Both fail on `main` (the first mismatches from index 8, the second decodes 0 for the last byte) and pass with the change.

Verified from `ts/` with `npx vitest run src` (32 files, 2486 tests passed), `npm run lint`, `npm run format:check` and `npm run build`, all clean.

### DIFF
--- a/ts/src/decoding/decodingUtils.spec.ts
+++ b/ts/src/decoding/decodingUtils.spec.ts
@@ -140,6 +140,16 @@ describe("decodingUtils", () => {
             // All 8 bits should be set in the first byte
             expect(result[0]).toBe(0xff);
         });
+
+        it("should round trip more booleans than one literal run can hold", () => {
+            // 1032 booleans pack into 129 bytes, one more than a literal run can carry
+            const data = Array.from({ length: 1032 }, (_, i) => i % 3 === 0);
+            const encoded = encodeBooleanRle(data);
+            const offset = new IntWrapper(0);
+            const bits = new BitVector(decodeBooleanRle(encoded, data.length, encoded.length, offset), data.length);
+
+            expect(data.map((_, i) => bits.get(i))).toEqual(data);
+        });
     });
 
     describe("decodeByteRle", () => {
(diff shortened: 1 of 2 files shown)


HarelM (approved):

HarelM:
Thanks!
@CommanderStorm is it worth a synthetic test somehow?

HarelM:
A new version will be released shortly.

CommanderStorm:
Likely, yes. But since TS does not produce the synthetics, it is not checking...

CommanderStorm:
Maybe through fuzzing we could find something like this, on the rust side that was very effective.. but time and stuff

HarelM:
Oh, right, it's only in the encoder...

MaxFreedomPollard:
🫡

PR #1657, opened 2026-09-07
feat(rust): scroll the mlt ui tree sideways with < and >

johncarmack1984:
Follow-up to #1626.

The sideways scroll was too hard to find. Shift+Left/Right was the only key for it, some terminals and multiplexers eat Shift+arrow, and at the default split the tree title clips before its `h:help` hint. `<` and `>` now scroll the tree sideways with no modifier, the title leads with `h:help, q:quit`, and the help overlay and README list the keys.


CommanderStorm (approved):
❤️

PR #1676, opened 2026-09-10
test(ts): round trip the byte and boolean RLE encoders against their header limits

MaxFreedomPollard:
Adds encoder side round trip tests for the byte and boolean RLE streams. On #1653 the question was whether a synthetic test could have caught the header length bug, so this is that test for the corner of the encoder that fix touched. `encodingUtils.ts` had no spec file of its own until now.

The tests are table driven and sit at the two points where a header runs out of room. A byte RLE header carries at most 130 repeats or 128 literals, so `encodeByteRle` is pinned to its exact output at 130 and 131 repeats and at 128 and 129 literals, and round tripped through `decodeByteRle` at 129, 130 and 131 repeats and at 127, 128 and 129 literals, checking that the decoder also consumes the whole stream. `encodeBooleanRle` packs eight booleans to a byte first, so the same limits fall at 1040 and 1024 booleans: its output is pinned at 1040 and 1048 set booleans, and it round trips through `decodeBooleanRle` at 1032, 1040 and 1048 set booleans and at 1016, 1024 and 1032 mixed ones.

Following the review, the first version's seeded random inputs and hand written replay decoder are gone; every case is a literal input and the repo's own decoder. The present stream, fixed width and string round trips went too, since `decodingUtils.spec.ts` and `unpackNullableUtils.spec.ts` already cover them. The file is 117 lines, down from 384.

Nothing new broke, so this is tests only. Reverting `encodingUtils.ts` to its state before #1653 fails 5 of the 18 cases: both checks at 131 repeats, both boolean output pins, and the 1032 mixed boolean round trip.

Verified in `ts/` with `npm run lint`, `npm run format:check`, `npm run test` and `npm run build`, all clean.

### DIFF
--- a/ts/src/encoding/encodingUtils.spec.ts
+++ b/ts/src/encoding/encodingUtils.spec.ts
@@ -1,0 +1,384 @@
+import { describe, expect, it } from "vitest";
+import {
+    createStringLengths,
+    encodeBooleanRle,
+    encodeByteRle,
+    encodeDoubleLE,
+    encodeFloatsLE,
+    encodeStrings,
+    encodeUint32sLE,
+    encodeUint64sLE,
+} from "./encodingUtils";
+import { packNullable } from "./packNullableUtils";
+import {
+    decodeBooleanRle,
+    decodeByteRle,
+    decodeDoublesLE,
+    decodeFloatsLE,
+    decodeString,
+    decodeUint32sLE,
+    decodeUint64sLE,
+} from "../decoding/decodingUtils";
+import { unpackNullable } from "../decoding/unpackNullableUtils";
+import BitVector from "../vector/flat/bitVector";
+import IntWrapper from "../decoding/intWrapper";
+
+/**
+ * A byte RLE header either opens a run of `header + 3` copies of the byte that follows it, or
+ * introduces `256 - header` literal bytes, so one header carries at most 130 repeats or 128
+ * literals. `docs/encodings.md` defers to the ORC specification for this, and
+ * `rust/mlt-core/src/codecs/rle.rs` encodes to the same two limits.
+ */
+const MAX_RUN = 130;
+const MAX_LITERALS = 128;
+
+/** Fixed seeds keep the generated cases identical on every machine and every run. */
+function seededRandom(seed: number): () => number {
+    let state = seed >>> 0;
+    return () => {
+        state = (state + 0x6d2b79f5) >>> 0;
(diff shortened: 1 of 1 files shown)


ts/src/encoding/encodingUtils.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+ * A byte RLE header either opens a run of `header + 3` copies of the byte that follows it, or
+ * introduces `256 - header` literal bytes, so one header carries at most 130 repeats or 128
+ * literals. `docs/encodings.md` defers to the ORC specification for this, and
+ * `rust/mlt-core/src/codecs/rle.rs` encodes to the same two limits.
+ */
+const MAX_RUN = 130;
+const MAX_LITERALS = 128;
+
+/** Fixed seeds keep the generated cases identical on every machine and every run. */
+function seededRandom(seed: number): () => number {
```
HarelM, line 36: Can this use a library instead?

HarelM:
This is a bit heavy on the logic when it comes to tests, can this be simplified somehow?

MaxFreedomPollard:
Simplified in `0e5c93b`: the seeded PRNG and the hand-written replay decoder are gone. The tests now round trip through `decodeByteRle`/`decodeBooleanRle` and are table-driven at the header limits. I dropped the present-stream, fixed-width and string sections since `decodingUtils.spec.ts` and `unpackNullableUtils.spec.ts` already cover them.

HarelM:
According to what you just wrote only 5 of the tests added here to cover the bug that was solved are actually needed since the rest passed on main. Moreover, the tests are using two inputs: the pre decoded/encoded and the post encoded/decoded, so you can't really understand what they actually test. 
Instead of a file of 100+ lines, a single test or two tests covering the bug that was solved would be enough, hopefully with a round trip test so input=output, which is the easiest to understand and maintain. 
Thanks for looking into this, and sorry if the comment is too blunt.

MaxFreedomPollard:
Not too blunt, that's a fair read. The two round trips you describe are already on main from #1653: `should handle 130 run max` (131 repeats through `decodeByteRle`) and `should round trip more booleans than one literal run can hold` (1032 mixed booleans through `decodeBooleanRle`), both in `decodingUtils.spec.ts` and both input = output. They are the same two round trip cases that fail here without the fix, so cutting this PR down to what you describe would leave nothing new. Closing this one. Thanks for the reviews.

PR #1684, opened 2026-09-11
feat(rust): code a nested node's structure per row, not per entry

CommanderStorm:
A nested struct spends one presence bit which keys a row has
- per field and row, and
- a map one key per entry

This codes that structure as one shape id per row into a table of key-set bitmaps, off by default behind `EncoderConfig::with_row_shapes`.
They are kept only where it comes out smaller than the per-entry candidate.

On OMT this seems about -13.6% on a homogeneous `name*` group and -9.8% on an all-properties bag.

This is an experiment.
A node below the root carries the shapes bit in its type byte, but the prefix byte a root writes instead is a placeholder and still needs a real codepoint.


nyurik (approved):

CommanderStorm:
TODO: cleanup docs pass

PR #1687, opened 2026-09-12
feat(ts): export `encodeTile()`

### LINKED ISSUE #1685: js: Can `encodeTile()` be exported?
I was wondering if I can implement a MLT version of [maplibre-contour](https://github.com/onthegomap/maplibre-contour) for practice, but I found `@maplibre/mlt` doesn't export the encoder (`encodeTile()`). I know this encoder is limited compared to other implementations, but probably we'll find it handy if we can use the encoder on web browsers.

Are there any blocking issue related to this? If not, may I attempt creating a pull request for this?

yutannihilation:
Close #1685

#1554 added `encodeTile()`, but didn't export. While the implementation might not be very mature, I think it's usable and can be exported. So, this pull request simply exports it along with the related types.

```ts
import { encodeTile, type Layer } from '@maplibre/mlt';

const layers: Layer[] = [
    {
        name: 'places',
        features: [
            {
                id: 1,
                geometry: { type: 'Point', coordinates: [10, 20] },
                properties: { name: 'Example' },
            },
        ],
    },
];

const data = encodeTile(layers);
```

### DIFF
--- a/docs/implementation-status.md
+++ b/docs/implementation-status.md
@@ -17,7 +17,7 @@
 | Implementation | Language | Type | Notes |
 |---|---|---|---|
 | [maplibre-tile-spec/cpp](https://github.com/maplibre/maplibre-tile-spec/tree/main/cpp) | C++ | Decoder | Used by MapLibre Native |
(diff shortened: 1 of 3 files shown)


CommanderStorm (approved):

CommanderStorm:
(also did a version bump pr just now, should be online soonish)

yutannihilation:
Thanks!

PR #1691, opened 2026-09-13
stop the C++ decoder stripping a leading BOM from string values

CommanderStorm:
Pretty weird code, I don't think it is nessary or fixes an actual bug


CommanderStorm:
CI is red on the cpp encoder, due to unmaintained. Not because of actual bug in this PR.

TimSylvester:
BOMs on UTF-8 are a pointless waste of space, and they do break various things.  It would be much better to remove them elsewhere to make the behavior match.  Anything that emits or retains them is a bug, IMO.

[https://www.w3.org/International/questions/qa-utf8-bom.en.html](https://www.w3.org/International/questions/qa-utf8-bom.en.html)

CommanderStorm:
Does any of the tooling actually produce BOM?
I think this should be on the tooling to produce, not the encoder or recoder.

TimSylvester:
I encountered them while testing the original decoder, but I don't know the original source, but it's probably one of the test tiles in our repo.  Ideally the encoder would fix that problem, it doesn't make sense to either pass that through or reject it.

nyurik:
I don't think encoder should modify input values, esp to have BOM recognition

nyurik (approved):

PR #1694, opened 2026-09-13
chore: drop the C++ encoder from CI

CommanderStorm:
Is eesentially unmaintained, has not gotten the last 7 additions or last removal.
Is currently failng CI, so anoying, so getting the axe.

We have an rust FFI wrapper, I suggest people use that instead.


CommanderStorm:
if someone wants to maintian fine.
We can revert the PR, but since CI is red, this is making CI green again.

PR #1708, opened 2026-09-16
fix(ts): decode FastPFOR delta-RLE streams correctly

DoFabien:
I hit a `RangeError` when decoding a FastPFOR + DELTA + RLE property. 
With two runs, `getVectorType()` was reading the compressed bytes directly as an `Int32Array`. The fix reuses the existing physical decoder and keeps the offset unchanged.

I added a small unit test and a fixture generated with Rust. Both fail before the fix and pass after it. 
The fixture also passes with the Java, C++ and Rust decoders.


CommanderStorm (approved):
LGTM, makes sense. Thanks

PR #1712, opened 2026-09-16
perf(java): Cache FastPFOR codec instances

TimSylvester:
Avoid creating a new instance of FastPFOR codec objects in each stream, as each instance allocates roughly 1MB for working buffers.  These instances are not thread-safe, however, so a separate instance must be used per thread.

Converting ~200MB from a local pmtiles file:

```bash
% time java -jar mlt-cli/build/libs/encode.jar --pmtiles planet-20260112.pmtiles --outlines ALL --enable-fastpfor --enable-fsst --verbose --parallel --maxzoom 7
```

Before:

`650.22s user 50.79s system 1314% cpu 53.315 total`

After:

`244.67s user 22.60s system 1204% cpu 22.189 total`

Cursor Grok 4.6 was used to add overloads and pass the new parameters through the API layers, and to create tests.


CommanderStorm (approved):
OK

> Cursor Grok 4.6 

lets not use the nazi family of models for the future, ok?

<details><summary>Details</summary>
<p>

Image

</p>
</details>

PR #1724, opened 2026-09-20
feat(rust): allow annotating partial trees

CommanderStorm:


rust/mlt-core/src/decoder/stream/model.rs
```diff
@@ -443,3 +445,115 @@ impl<'a> RawStream<'a> {
         Self { meta, data }
     }
 }
+
+// Display impls the annotated dump's JSON carries instead of serde derives on this
+// vocabulary, so the wire enums do not become frozen public JSON API.
+
+impl Display for StreamType {
+    /// `present`, `data`, `data[fsst]`, `offset[string]`, `length[rings]`.
```
CommanderStorm:
```suggestion
impl Display for StreamType {
```

PR #1735, opened 2026-09-21
feat(rust): implement an inspector

CommanderStorm:
image
image


nyurik (approved):
vibing UI is fine.  Let's only focus on the core / wasm parts that are usable by others

